PR #32, opened 2026-03-02
fix: enforce scope isolation using runtime agentId in tools

1034378361:
## Bug
`memory-lancedb-pro` tools were using the registration-time `context.agentId` (often `undefined`).

When `agentId` is missing, scope filtering becomes overly permissive and can bypass `scopes.agentAccess` isolation, allowing access to all valid scopes.

## Symptoms / Repro
With a config like:

```json
"agentAccess": {
  "main": ["global", "agent:main"],
  "quant": ["global", "agent:quant"],
  "girlfriend": ["agent:girlfriend"]
}
```

From a **main** agent session, the following should be denied but can succeed:
- `memory_stats(scope="agent:quant")`
- `memory_list(scope="agent:quant")`
- `memory_recall(scope="agent:quant")`

Expected: `Access denied to scope: agent:quant`.

## Root Cause
Tools rely on `context.agentId` provided at registration time (`index.ts` uses `agentId: undefined`), while the real agent id is only known at runtime per tool call.

## Fix
Read the runtime agent id from the tool call context (`ctx.agentId`) and use it for:
- `getAccessibleScopes(agentId)`
- `isAccessible(scope, agentId)`
- `getDefaultScope(agentId)`

Applied to these tools:
- `memory_recall`
- `memory_store`
- `memory_forget`
- `memory_update`
- `memory_stats`
- `memory_list`

## Testing
- `npm test` (CLI smoke test)

## Notes
This PR only addresses permission filtering/scope isolation; no other behavior changes.

### DIFF
--- a/src/tools.ts
+++ b/src/tools.ts
@@ -23,7 +23,14 @@ interface ToolContext {
   store: MemoryStore;
   scopeManager: MemoryScopeManager;
   embedder: Embedder;
+  // NOTE: This value is set during plugin registration, but the real agent ID
+  // is only known at runtime per tool call.
   agentId?: string;
+}
+
+function resolveAgentId(runtimeAgentId: unknown, fallback?: string): string | undefined {
+  if (typeof runtimeAgentId === "string" && runtimeAgentId.trim().length > 0) return runtimeAgentId;
+  return fallback;
 }
 
 // ============================================================================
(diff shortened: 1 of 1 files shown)


win4r:
Thank you @1034378361 for identifying this scope isolation vulnerability! Great analysis of the root cause.

I've applied the fix using a slightly different approach — **OpenClaw's tool factory pattern** (`registerTool(factoryFn, opts)`) instead of adding a 3rd `ctx` argument to `execute()`. The reason is that `execute()` is defined by `@mariozechner/pi-agent-core`'s `AgentTool` interface as `(toolCallId, params, signal?, onUpdate?)`, so adding a context parameter there would not receive the plugin tool context at runtime.

The factory function receives `OpenClawPluginToolContext` (which includes `agentId`) each time the tool is resolved for an agent session, giving us the correct runtime `agentId`.

Released as **v1.0.23** with your co-author credit. 🎉

See: https://github.com/win4r/memory-lancedb-pro/releases/tag/v1.0.23

1034378361:
@win4r 你好，

我注意到项目从 win4r/memory-lancedb-pro 转移到 CortexReach/memory-lancedb-pro 后，我不在新项目的 Contributors 列表中了。

在之前的项目中，我是贡献者之一。转移后这个贡献记录丢失了。

能否帮忙修复这个问题，让我重新出现在 Contributors 列表中？

谢谢！

Issue #55, opened 2026-03-05
Reduce false positives for tag-style queries like proj:AIF (BM25-only + mustContain)


1034378361:
## Problem
Hybrid retrieval works well for natural language, but short tag-style tokens (e.g. `proj:AIF`) can produce semantic false positives because vector search dominates the candidate pool (and the internal vectorSearch minScore is currently permissive).

In practice, searching for a project tag like `proj:AIF` can recall unrelated memories that do not literally contain that tag.

## Expected
When the query contains tag-like tokens (e.g. `proj:XXX`), results should behave like an exact filter: only return entries that literally contain the tag token(s).

## Proposed change
Route queries that contain `proj:` tokens to an FTS-first path:
1) Run BM25 search (FTS-only).
2) Apply a hard filter: result text must contain *all* extracted `proj:` tokens (mustContain).
3) If there are no literal matches, fall back to the existing hybrid retrieval (so users still can find related wording).

This keeps the default hybrid behavior unchanged for normal natural-language queries, while making tag queries reliable and noise-free.

## PR
I opened a PR from my fork implementing this in `src/retriever.ts` (v1.0.30 base):
- Branch: `fix/tag-query-fts-only`
- Commit: 814e2ca

AliceLJY:
这个思路是对的：`proj:AIF` 这类 tag-style query 和自然语言检索的最佳策略本来就不一样。FTS-first + mustContain 会比默认 hybrid 更稳，尤其能压住 BM25/vector 的语义误报；只是现在仓库 retrieval 逻辑改动比较多，真正合之前最好基于最新主分支再 rebase 一次。

PR #56, opened 2026-03-05
fix(retrieval): make proj: tag queries exact (BM25-only + mustContain)

1034378361:
## What
When a query contains tag-style tokens like `proj:AIF`, treat it as an exact filter instead of a semantic query:
- Use BM25 (FTS-only) first
- Hard filter results: mustContain all extracted `proj:` tokens
- If no literal matches, fall back to existing hybrid retrieval (to avoid returning nothing)

## Why
Short tokens are prone to semantic false positives in hybrid retrieval because vector search dominates the candidate pool. For project tags, users usually expect exact matching, not related-content recall.

## Scope / Compatibility
- Only affects queries containing `proj:` tokens
- Normal natural-language queries keep the original hybrid path

## Implementation
`src/retriever.ts`:
- Added tag token extraction (`proj:[A-Za-z0-9._-]+`)
- Added `bm25OnlyRetrieval()` helper with mustContain filter
- Hooked in early in `retrieve()`

## Related issue
Fixes: win4r/memory-lancedb-pro#55

### DIFF
--- a/src/retriever.ts
+++ b/src/retriever.ts
@@ -280,6 +280,7 @@ function cosineSimilarity(a: number[], b: number[]): number {
 // ============================================================================
 
 export class MemoryRetriever {
+  private static TAG_QUERY_RE = /\bproj:[A-Za-z0-9][A-Za-z0-9._-]{0,63}\b/g;
   private accessTracker: AccessTracker | null = null;
 
   constructor(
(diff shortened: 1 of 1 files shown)


rwmjhb (commented):
Thanks for the PR! The approach is sound — routing tag-style queries to BM25 + mustContain is a clean solution for the false-positive problem described in #55.

I verified locally that LanceDB FTS treats `:` as a tokenizer (not a Tantivy field separator), so the BM25 stage correctly returns candidates and `mustContain` filters them down. The core logic works.

A few suggestions before merging:

### 1. Missing test coverage

+81 lines of new logic with no tests. At minimum, a unit test for `extractTagTokens()` and an integration-level test verifying that a `proj:AIF` query returns only entries literally containing that tag would be helpful.

### 2. `as RetrievalResult` type safety

```typescript
const mapped = literalFiltered.map(
  (result, index) =>
    ({
      ...result,
      sources: {
        bm25: { score: result.score, rank: index + 1 },
        fused: { score: result.score },
      },
    }) as RetrievalResult,
);
```

`sources.vector` is absent here. If any downstream code accesses `sources.vector.score`, it will throw. Worth checking that no consumer expects it, or add `vector: undefined` explicitly.

### 3. (Minor) Extensibility of tag patterns

Currently hardcoded to `proj:` only. Not a blocker, but consider making the tag prefix configurable (e.g., via retrieval config) or at least using an array of patterns, so adding `env:`, `team:`, etc. later doesn't require code changes.

Overall: approve once tests are added. Nice work! 👍

PR #75, opened 2026-03-07
feat: add cloud storage support

ddupg:
## Add cloud storage support for LanceDB                                                                                                                                                                       
                                                                                                                                                                                                                 
This PR adds support for connecting to cloud-based LanceDB instances by introducing a new storageOptions configuration parameter.                                                                           
                                                                                                                                                                                                                 
## Changes                                                                                                                                                                                                     
                                                                                                                                                                                                                 
- Plugin configuration: Added storageOptions field to accept key-value string pairs for LanceDB connection options                                                                                          
- Cloud URL detection: Automatically detects cloud storage URLs (containing ://) and skips local filesystem validation                                                                                      
- LanceDB connection: Passes storageOptions to the lancedb.connect() call for cloud authentication and configuration                                                                                        
- Schema updates: Updated plugin JSON schema and UI labels for the new configuration option                                                                                                                 
                                                                                                                                                                                                                 
Usage                                                                                                                                                                                                       
                                                                                                                                                                                                                 
Users can now configure cloud storage connections by providing storage options in the plugin configuration:                                                                                                 
```                                                                                                                                                                                                            
{                                                                                                                                                                                                           
  dbPath: s3://my-bucket/path,                                                                                                                                                                              
  storageOptions: {
    access_key_id: your-key,
    secret_access_key: your-secret,
    aws_endpoint: bucket endpoint,
  }                                                                                                                                                                                                         
}                                                                                                                                                                                                           
```                                                                                                                                                                                                            
This enables the plugin to work with LanceDB Cloud, S3-backed storage, and other remote storage backends without requiring local filesystem access.

### DIFF
--- a/cli.ts
+++ b/cli.ts
@@ -646,6 +646,7 @@ export function registerMemoryCLI(program: Command, context: CLIContext): void {
     .option("--dry-run", "Show what would be re-embedded without writing")
     .option("--skip-existing", "Skip entries whose id already exists in the target DB")
     .option("--force", "Allow using the same source-db as the target dbPath (DANGEROUS)")
+    .option("--storage-options <json>", "Storage options for LanceDB connection (JSON string, e.g. '{\"aws_access_key_id\":\"...\"}')")
     .action(async (options) => {
       try {
         if (!context.embedder) {
@@ -661,6 +662,23 @@ export function registerMemoryCLI(program: Command, context: CLIContext): void {
         const dryRun = options.dryRun === true;
         const skipExisting = options.skipExisting === true;
         const force = options.force === true;
+
+        let storageOptions: Record<string, string> = {};
+        if (options.storageOptions) {
+          try {
+            const parsed = JSON.parse(options.storageOptions as string);
+            if (typeof parsed === "object" && parsed !== null && !Array.isArray(parsed)) {
+              for (const [key, value] of Object.entries(parsed)) {
+                if (typeof value === "string") {
+                  storageOptions[key] = value;
+                }
+              }
+            }
+          } catch {
+            console.error("Invalid JSON in --storage-options");
+            process.exit(1);
+          }
+        }
 
         // Safety: prevent accidental in-place re-embedding
         let sourceReal = sourceDbPath;
@@ -678,7 +696,7 @@ export function registerMemoryCLI(program: Command, context: CLIContext): void {
         }
 
         const lancedb = await loadLanceDB();
-        const db = await lancedb.connect(sourceDbPath);
+        const db = await lancedb.connect(sourceDbPath, { storageOptions });
         const table = await db.openTable("memories");
 
         let query = table
--- a/index.ts
+++ b/index.ts
@@ -1449,7 +1449,7 @@ const memoryLanceDBProPlugin = {
     const config = parsePluginConfig(api.pluginConfig);
 
     const dbPath = config.dbPath || getDefaultDbPath();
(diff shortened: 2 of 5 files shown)


ddupg:
Would @win4r mind reviewing this PR when you get a chance?

ddupg:
Thanks @rwmjhb a lot for the detailed and thoughtful review! I really appreciate the feedback and have already addressed most of the points.

I have a question regarding #2 (passing storageOptions to migration). From my understanding, the LegacyData in the migrate refers to data written by the memory-lance plugin. Currently, the memory-lance plugin does not support cloud storage yet.Given that, I’m wondering if it’s necessary to pass storageOptions to loadLegacyData() and checkForLegacyData() at this moment.

Would it make sense to defer this change until the memory-lance plugin actually adds cloud storage support? We can track the migration compatibility in a separate issue or PR later, as there will likely be more details to handle — such as how to infer or pass the correct storageOptions for legacy data.

Let me know what you think!

ddupg:
@rwmjhb Gentle reminder for review, thanks!

rwmjhb:
I re-reviewed the current head. My updated take is:

- `storageOptions` env var resolution, the `sensitive` UI hint, and the primary store cloud connection path all look good now.
- CI is passing, so no concerns there.

My remaining concern is scope clarity: the cloud-storage support in this PR currently covers the primary store connection, but not the migration/source-db path.

Concretely:
- `src/migrate.ts` still scans only local directories in `findSourceDatabase()`
- `loadLegacyData()` still does a bare `lancedb.connect(sourceDbPath)` without `storageOptions`

So if the intended scope of this PR is “allow the primary LanceDB memory store to use a cloud backend”, I think the implementation is largely there.
But if the intended scope is broader “cloud storage support” in general, then the PR description is currently stronger than the actual coverage, because migration/cloud source-db is not implemented yet.

Also, this repo’s default branch is currently `master`, not `main`, so please retarget the PR base branch to `master` and re-sync with the latest `master`.

One more question: what is the main use case for putting the primary LanceDB memory store on cloud storage?
For example, is the goal cross-machine/shared memory, persistence in ephemeral environments, centralized backup, or multi-agent access to the same memory backend?
That context would help clarify whether migration/cloud-source support should be considered in-scope for this PR or left as follow-up work.

My suggestion:
1. Retarget the PR to `master`
2. If migration/cloud source support is out of scope, narrow the PR description to make that explicit
3. If you want to keep the current broader description, then the migration/source-db path still needs to be completed

If you go with (2), I don’t have a new blocker.

ddupg:
Thank you @rwmjhb  very much for re-reviewing and providing such constructive feedback.

As you correctly pointed out, this PR only adds cloud storage support for the primary store at the moment, and does not cover the migration path. The reasons are:
I understand migration/source-db refers to data written by the memory-lancedb plugin. If I’m wrong, please correct me. Since memory-lancedb itself does not yet support cloud storage, implementing cloud support for the source-db migration is not urgent right now.

Supporting cloud storage for migration would require more detailed design decisions: for example, how to get storageOptions for the legacy store. If we support cross-storage migration (local to cloud, or cross-cloud), the source and target would need different storageOptions, the migration process is triggered during plugin registration, a stage where it’s not possible to retrieve the storageOptions of the legacy store.

For these reasons, I prefer to implement storageOptions support in the migration path in a separate follow-up PR.

I have rebased against the master branch. Please follow up on #142 , and I will close this PR.

Thanks again for your help!

PR #86, opened 2026-03-07
feat: make autoRecall injection safer by default

### LINKED ISSUE #85: Make autoRecall injection safer and less likely to leak verbatim
## Summary
`autoRecall` is useful, but the current legacy XML wrapper (`<relevant-memories>`) can sometimes be echoed back by the model verbatim.

## Proposal
Introduce a safer injection format for auto-recalled memories:
- add `autoRecallFormat` config with `plain` (default) and `xml` (legacy compatibility)
- default to a plain internal block with stronger anti-leak instructions
- preserve `xml` as an opt-in compatibility mode
- update detection/filters so injected recall blocks are not re-captured as memory

## Why this matters
This improves the default safety/UX of autoRecall without removing the feature, and gives users an explicit compatibility switch if they still want the old XML wrapper.

mayisme:
## Summary
This PR makes `autoRecall` safer by default and less likely to leak injected memory blocks verbatim in user-facing replies.

## Changes
- add `autoRecallFormat` config with:
  - `plain` (new default)
  - `xml` (legacy compatibility)
- move recall context formatting into `src/recall-format.ts`
- use a plain internal recall block by default instead of the legacy `<relevant-memories>` wrapper
- preserve `xml` as an explicit opt-in mode
- update injection detection so the new recall block is also excluded from capture / re-ingestion
- add tests for recall formatting helpers
- update README / README_CN to document the safer default

## Why
The current XML-style wrapper is functional, but can sometimes be echoed back by the model. A safer default lowers that risk without removing `autoRecall` for users who want it.

## Testing
- `npm test`

Closes #85


rwmjhb:
Thanks for the contribution. The core idea here is good — making auto-recall safer by default is worth doing.

At this point, though, this PR is stale, conflicts with current `main`, and still has unresolved blocker feedback around format-aware paths outside the main injection site (reflection re-ingestion and JSONL distillation).

Rather than keep reviewing this old branch, the more practical path is to close it and, if still needed, re-open the work as a smaller follow-up PR on top of current `main` with all format consumers updated together.

PR #94, opened 2026-03-07
fix: flatten legacy migration regression test path

nyaxnyax:
## Summary

Follow-up to #90. This fixes the accidentally nested regression test path created via the GitHub web editor.

## Changes

- remove the nested path `test/migrate-legacy-schema.test.mjs/migrate-legacy-schema.test.mjs`
- restore the intended flat file `test/migrate-legacy-schema.test.mjs`

## Notes

The nested path created a file/directory conflict for the intended test location.

## Validation

- compare against `win4r:main` shows a pure path flattening with no content changes

Follow-up to #90


rwmjhb:
Thanks for the fix attempt! However, this issue has already been resolved in main — the file is already at the correct flat path `test/migrate-legacy-schema.test.mjs`. Closing as no longer needed.

PR #96, opened 2026-03-07
fix: rollback MemoryStore.update when add fails after delete

robinspt:
## Summary

Prevent silent memory loss in `MemoryStore.update()` when the post-delete add fails.

## What changed

- Preserve the original record before delete
- Attempt best-effort rollback if `add(updated)` fails after delete
- Return a descriptive error when rollback succeeds or fails

## Why

`update()` currently uses `delete + add`. If `add()` fails after delete, the original memory is lost.
This affects both direct memory updates and metadata write-backs such as `AccessTracker.flush()`.

## Validation

- Added `test/update-consistency-lancedb.test.mjs`
- Verified:
  - original record is restored when add fails after delete
  - concurrent update failure does not drop the record
  - access-tracker metadata write failure does not remove the record
  - successful updates still read correctly through `getById` and `list`

## Scope

This change focuses on preventing record loss when `add()` fails after delete.

### DIFF
--- a/cli.ts
+++ b/cli.ts
@@ -5,7 +5,11 @@
 import type { Command } from "commander";
 import { readFileSync } from "node:fs";
 import { loadLanceDB, type MemoryEntry, type MemoryStore } from "./src/store.js";
-import type { MemoryRetriever } from "./src/retriever.js";
+import type {
+  MemoryRetriever,
+  RetrievalExecution,
+  RetrievalTrace,
+} from "./src/retriever.js";
 import type { MemoryScopeManager } from "./src/scopes.js";
 import type { MemoryMigrator } from "./src/migrate.js";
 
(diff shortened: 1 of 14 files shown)


robinspt:
This failing check appears unrelated to this PR.

  I verified that `test/migrate-legacy-schema.test.mjs` does not exist in this PR branch, and it also does not exist in the current `main`
  branch. The `package.json` test script in both branches still runs only:

  `node test/embedder-error-hints.test.mjs && node test/cli-smoke.mjs`

  So the workflow seems to be picking up a non-existent test file outside the actual PR diff.

rwmjhb:
Thanks for checking this.

You are right that `test/migrate-legacy-schema.test.mjs` is not present on your PR head branch. However, the failing CI is running on the PR merge ref, not just the head branch. On the merge ref for this PR, the workflow uses the current base branch test script and test files, which is why `npm test` is picking up `test/migrate-legacy-schema.test.mjs`. So I agree the failing check looks unrelated to the actual diff in this PR, but it is not coming from “outside” the PR workflow either; it is coming from the merged PR/base context that GitHub Actions runs for `pull_request`.

On the PR itself, I still see two issues:

1. The rollback logic can overwrite a successful concurrent update with stale data  
   In `MemoryStore.update()`, the rollback path restores the snapshot read at the start of the failing update. In a concurrent case, that snapshot may already be stale. Your new test demonstrates this: after `update from A` succeeds, the later failed update restores `original memory`, which means the fix prevents record loss but does not preserve the latest committed value under concurrency.

2. The new regression test is not part of the default test script  
   `test/update-consistency-lancedb.test.mjs` is valuable coverage for this change, but it is not included in `package.json`'s `test` script. For a reliability fix like this, I think the new test should be part of the default test path.

So my current assessment is:
- the anti-loss fix is useful
- the unrelated CI failure should be handled separately
- but this PR still needs follow-up on the concurrent rollback semantics, and the new regression test should be wired into the default test run

robinspt:
Thanks for the review. I pushed a follow-up update addressing both points.

  Changes in the latest commit:
  - `MemoryStore.update()` now serializes updates per store instance and avoids restoring a stale snapshot over an already-preserved newer
  record after a failed write
  - `test/update-consistency-lancedb.test.mjs` is now included in the default `npm test` path

  I also updated the LanceDB-backed regression test so the concurrent failure case now asserts that the latest committed value is preserved
  rather than rolled back to the original value.

rwmjhb:
I resolved the current `main` merge conflict on an upstream branch and verified the merged state locally.

Conflict-resolved branch in upstream repo:
- `fix/update-rollback-on-add-failure`
- head commit: `78c0bf33686e89ca9898119decc68f531a5b2436`

What I changed in the merge resolution:
- kept the PR's `test/update-consistency-lancedb.test.mjs`
- kept main's newer `test/vllm-provider.test.mjs`
- updated `package.json` so `npm test` runs both

Verification:
- `npm test` passes locally after the merge resolution (`59/59`)

I could not update the PR branch directly because this PR still points to the fork head under `robinspt`. If you want, you can either:
1. cherry-pick `78c0bf33686e89ca9898119decc68f531a5b2436` onto the PR branch, or
2. open a replacement PR from the upstream branch above.

rwmjhb:
The default branch has moved from `main` to `master`. Since this PR targets `main` and has significant merge conflicts against `master`, I cherry-picked the two core commits onto a new branch based on `master` and opened #140 as a replacement.

#140 preserves all your changes (rollback logic, serialized updates, and the regression test) with the `package.json` conflict resolved to include the new test alongside `master`'s existing test suite.

Closing this PR in favor of #140. Thanks for the contribution @robinspt!

PR #97, opened 2026-03-07
feat(reflection): refine reflection persistence, governance extraction, and prompt/docs alignment

furedericca-lab:
## Summary
- split reflection persistence into separate Inherit and Derive entries while keeping `category=reflection`
- add logistic weighting for Derive injection instead of relying on only the latest reflection row
- keep legacy combined reflection rows readable/injectable without a required migration
- extract part of the reflection metadata/slice/store logic out of `index.ts`

## Why
- stable carry-over rules and short-lived next-run deltas should not live in the same reflection memory shape
- the previous derive path was too dependent on the latest reflection row
- reflection injection needs a decay-aware way to surface recent execution deltas without changing global retriever scoring
- the reflection path in `index.ts` had grown enough to justify a focused structural cleanup

## Changes
- store split reflection rows with `metadata.reflectionKind = "inherit" | "derive"`
- display new rows as `reflection:Inherit` and `reflection:Derive`
- preserve a legacy-safe read/display path for old combined reflection rows
- apply logistic derive weighting during reflection loading/injection only
- use defaults:
  - `midpointDays = 3`
  - `k = 1.2`
  - fallback derive base weight = `0.35`
- keep `<inherited-rules>` sourced from inherit memories only
- keep `<derived-focus>` sourced from derive memories only
- add focused tests for split persistence, legacy compatibility, subtype display tags, and derive weighting

## Verification
- `npm test`

## Risks / Compatibility
- compatibility is preserved by keeping `category=reflection` and continuing to read legacy combined reflection metadata
- derive injection behavior now aggregates multiple recent derive rows with decay instead of only using the latest row
- no global retriever scoring behavior was changed

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -186,6 +186,7 @@ const DEFAULT_REFLECTION_SESSION_TTL_MS = 30 * 60 * 1000;
 const DEFAULT_REFLECTION_MAX_TRACKED_SESSIONS = 200;
 const DEFAULT_REFLECTION_ERROR_SCAN_MAX_CHARS = 8_000;
 const REFLECTION_FALLBACK_MARKER = "(fallback) Reflection generation failed; storing minimal pointer only.";
+const DIAG_BUILD_TAG = "memory-lancedb-pro-diag-20260308-0058";
 
 type ReflectionErrorSignal = {
   at: number;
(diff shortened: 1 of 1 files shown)


furedericca-lab:
- Core issue: https://github.com/openclaw/openclaw/issues/37323
    (--json output can be polluted by startup/plugin logs on Gateway 2026.3.2)
  - Our merged hardening: https://github.com/win4r/memory-lancedb-pro/pull/97
    (merged as f0f5b64: reflection timeout guard + explicit early-return/stage diagnostics)

  So behavior is improved on our side, but full reliability still depends on the official OpenClaw fix for #37323.

PR #112, opened 2026-03-08
feat: Chinese synonym query expansion for BM25 search

AliceLJY:
## Summary

- Add a lightweight static synonym dictionary (`query-expander.ts`) that expands colloquial Chinese terms into technical equivalents before BM25 search
- Vector search stays unchanged — embeddings already handle semantic similarity
- Examples: "挂了" → adds "崩溃 crash error", "踩坑" → adds "bug 排查 troubleshoot"

## Why

Chinese users often search with colloquial terms ("挂了", "炸了", "卡死") while stored content uses technical terms ("crash", "OOM", "timeout"). Vector search bridges some of this gap, but BM25 needs exact term matches. This dictionary-based expansion layer boosts BM25 recall for fuzzy Chinese queries without adding any external dependencies.

## Changes

- `src/query-expander.ts` — static synonym dictionary + `expandQuery()` function
- `src/retriever.ts` — one-line change: wrap BM25 query with `expandQuery()`
- `src/__tests__/query-expander.test.ts` — 5 unit tests

## Test results

Tested locally on my own dataset (1000+ conversation entries):
- 3 new fuzzy Chinese test cases: **94% / 73% / 100%** pass rate
- 6 existing eval cases: **zero regression**
- Overall: 8/9 pass, avg 90.1%

## Design decisions

- **Static dictionary, no API calls** — zero latency, zero cost, zero dependencies
- **BM25 only** — vector search untouched (Jina embeddings handle semantics)
- **Reranker uses original query** — expansion only broadens candidate pool, final ranking judges against user intent
- Dictionary is easily extensible — users can add domain-specific entries


rwmjhb:
I cross-checked the review comments and would keep two formal findings:

1. The new tests are not integrated into the repository’s existing test pipeline.
This PR adds a `bun:test` file under `src/__tests__/`, but the project’s current test entrypoint is the `package.json` script using `node ...` / `node --test test/*.mjs`, so this new test is not exercised by the standard test flow. As it stands, the behavior change is effectively untested in CI / `npm test`.

2. The query expansion triggers are too broad and can cause false-positive retrieval.
The implementation uses substring matching via `includes()`, while the dictionary contains high-risk triggers such as `down`, `log`, `push`, and the single Chinese character `找`. That means unrelated queries like `download`, `blog`, or `dialog` can be expanded into failure/log/retrieval-related terms. Since the expanded query is sent directly into BM25, and BM25-only hits can participate in the fused result set, this is a real retrieval-quality regression rather than a minor dictionary issue.

I would not keep the other items as blocking findings:
- Having trigger terms also appear in expansions is more of a redundancy / maintainability issue than a standalone defect.
- “Needs rebase” is a workflow concern, not a proven code issue based on the current patch.
- Test naming, the short-query threshold comment, and documenting that this only affects hybrid mode are follow-up cleanup items, not merge blockers.

Conclusion: before merge, this PR should at minimum fix the test integration gap and narrow the expansion matching to avoid obvious false positives.

AliceLJY:
感谢恒哥的 review，两个 finding 都是实锤，我跑了一轮完整的静态+动态测试确认了：

**Finding 1（子串匹配误触发）**：我构造了 15 个无关查询测试，14 个被误触发。还跑了 Discord 端到端验证——问 "what do you know about my blog?"，因为 `blog` 包含 `log`，BM25 被扩展出 "日志 logging 输出 打印"，auto-recall 拉回了 3 条完全无关的记忆。是真实的检索退化，不是理论风险。

**Finding 2（测试孤岛）**：`npm test` 跑了 37 个测试全部通过，其中 0 个是 query-expander 的。我用 bun:test 写在 src/__tests__/ 里，但项目用的是 node --test test/*.mjs，完全脱节。

**修复计划**（已推送 [4dc2163](https://github.com/AliceLJY/memory-lancedb-pro/commit/4dc2163)）：
1. 英文触发词改用 `\b` 词边界正则匹配；移除/加长过短的触发词（`down`→`shutdown`、`log` 移除、单字 `找` 改为 `找不到`）
2. 测试迁移到 `test/query-expander.test.mjs`，用 `node:test` + `assert`，加入 package.json scripts
3. 加扩展上限（最多追加 5 个词），防止多重触发膨胀

根因反思：太急着写代码，跳过了调研——没先跑 `npm test` 看项目测试规范，没给匹配逻辑写反向用例。两个都是本该 5 分钟就能避免的低级错误。

再次感谢！

PR #113, opened 2026-03-08
feat: add exposeRetrievalMetadata and clean recall output

Disaster-Terminator:
## Background

`autoRecall` injection and `memory_recall` currently include retrieval metadata in their default output, for example:

`[category:scope] text (80%, vector+BM25+reranked)`

That information can be useful while debugging retrieval behavior, but it is noisy in normal operation:

- it adds unnecessary token cost to LLM-visible context;
- it exposes low-level retrieval details in user-facing output;
- it is not currently controlled by an explicit config flag.

This PR is a follow-up to #107. PR #107 was closed because unrelated upstream commits were accidentally mixed in during squash. This revision reapplies only the intended retrieval-metadata change on top of the current `main`.

### 1. Hide retrieval metadata by default
- remove `(score, sources)` suffixes from `autoRecall` injected lines;
- remove retrieval score/source annotations from the human/LLM-visible text returned by `memory_recall`;
- stop exposing `score` and `sources` in `details.memories` by default.

### 2. Add an explicit debug switch
Introduce a new `exposeRetrievalMetadata` config option, defaulting to `false`.

When enabled, retrieval metadata is exposed separately via `details.debug`, instead of being mixed into the default text output.

### 3. Preserve existing candidate response structure
For the candidate flows in `memory_forget` and `memory_update`, keep the existing `details.candidates` structure and attach `details.debug` only when needed. This avoids introducing an unnecessary response-shape change.

## Files changed

| File | Purpose |
| --- | --- |
| `index.ts` | clean up autoRecall output and pass `exposeRetrievalMetadata` into tool context |
| `src/tools.ts` | separate default output from debug metadata in recall / forget / update flows |
| `openclaw.plugin.json` | add the `exposeRetrievalMetadata` config option |
| `test/memory-recall-metadata.test.mjs` | add regression coverage for metadata visibility and candidate-structure compatibility |
| `package.json` | include the new test in the test script |

## Validation

```bash
npm ci
npm test
```

Result: **41/41 passing**.

## Scope

This revision is intentionally limited to:
- removing retrieval metadata from default recall output;
- adding an explicit debug toggle;
- preserving compatibility for existing candidate response payloads.

### DIFF
--- a/test/memory-recall-metadata.test.mjs
+++ b/test/memory-recall-metadata.test.mjs
@@ -107,16 +107,20 @@ describe("memory metadata exposure", () => {
   it("exposes debug metadata without polluting memory_recall text when enabled", async () => {
     const tool = createTool(
       registerMemoryRecallTool,
-      makeContext({ expose: true, results: makeResults(1) }),
+      makeContext({ expose: true, results: makeResults(2) }),
     );
 
     const res = await tool.execute(null, { query: "test" });
 
-    assert.equal(res.details.count, 1);
+    assert.equal(res.details.count, 2);
     assert.ok(Array.isArray(res.details.memories));
     assert.ok(Array.isArray(res.details.debug));
+    assert.equal(res.details.debug.length, res.details.memories.length);
     assert.equal(typeof res.details.debug[0].score, "number");
     assert.ok(res.details.debug[0].sources.vector);
+    assert.ok(res.details.debug[0].sources.bm25);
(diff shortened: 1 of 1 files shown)


win4r:
@claude

Disaster-Terminator:
Thanks — I’ve moved this into a fresh PR rebased onto the current `master`: #159.

This older PR is now superseded and can be safely closed. The new PR keeps the non-breaking text cleanup only, with `details.memories[*].score` and `sources` fully preserved as requested.

PR #123, opened 2026-03-08
fix: relax autoRecall skip on first turn of fresh session

### LINKED ISSUE #101: Fresh-session continuity break: weak first message (e.g. 🤔) skips autoRecall after restart
## Summary

After restarting OpenClaw Gateway (with `memory-lancedb-pro` enabled), the **first user message in the fresh session** can be a very low-signal continuation such as:

- `🤔`
- `?`
- `...`
- `嗯`
- `继续`

In this case, `autoRecall` is **silently skipped** by adaptive retrieval, which makes the assistant lose continuity at exactly the moment when continuity matters most.

This is especially confusing when the user is clearly continuing the conversation right after restart: the assistant behaves as if it suddenly forgot the previous topic and replies with a generic opener.

---

## Environment

- OpenClaw: `2026.3.2`
- Plugin: `memory-lancedb-pro`
- Memory slot: `memory-lancedb-pro`
- Config:
  - `autoRecall: true`
  - `autoCapture: true`
  - `sessionStrategy: "memoryReflection"`
- Embedding:
  - OpenAI-compatible / OpenRouter
  - model: `qwen/qwen3-embedding-8b`

---

## Reproduction

1. Enable `memory-lancedb-pro` with `autoRecall: true`.
2. Have an active conversation with clear topic continuity.
3. Restart OpenClaw Gateway (or otherwise cause a fresh session to start).
4. Send a **low-signal first message** in the new session, for example: `🤔`
5. Observe the assistant response.

### Observed behavior

The assistant responds as if the conversation context was not continued, e.g. with something like:

> “怎么啦，在想什么？”

instead of understanding that the user is continuing the immediately previous topic.

---

## Root cause analysis

From local code inspection:

### 1. `before_agent_start` autoRecall gate
In `index.ts`, autoRecall runs only if:

```ts
!shouldSkipRetrieval(event.prompt, config.autoRecallMinLength)
```

### 2. `normalizeQuery()` strips OpenClaw wrappers
In `src/adaptive-retrieval.ts`, the query is normalized by stripping:

- `Conversation info (untrusted metadata): ...`
- `Sender (untrusted metadata): ...`
- timestamp wrappers
- cron prefixes

This is correct and desirable.

### 3. But after stripping, the first message may collapse to just `🤔`
Then `shouldSkipRetrieval()` returns `true` because of rules like:

- pure emoji skip
- `trimmed.length < 5`
- short non-question skip

So on a fresh post-restart session, the plugin effectively says:

> “this message is too weak to retrieve memory for”

But from the **user experience** perspective, this is exactly the moment where continuity recovery is needed most.

### 4. `memoryReflection` does not solve this
`memoryReflection` injects inheritance/derived guidance, but it does **not** restore the immediate prior conversational thread.
So the result is:

- fresh session
- no immediate continuity state
- first message too weak
- autoRecall skipped
- user sees an apparent “amnesia” moment

---

## Why this matters

This is not just a minor retrieval heuristic edge case.
It directly affects **trust in continuity**:

- users often send very short continuation messages after restart
- on chat surfaces, emoji / `?` / `继续` / `嗯` are normal follow-ups
- the current behavior makes the assistant feel like it “suddenly forgot everything”

That user-visible effect is much worse than a normal recall miss.

---

## Expected behavior

At least one of these should happen for the **first turn of a fresh session**:

1. **Do not blindly skip autoRecall for weak first-turn messages**
   - e.g. relax skip rules on session-first turn

2. **Fallback continuity recovery**
   - inspect recent same-chat session tail / recent reflection / session summary before deciding to skip

3. **Special handling for restart/new-session continuity turns**
   - if the session is fresh and the message is weak, prefer continuity heuristics over generic skip heuristics

---

### Option A — session-aware skip logic
In `shouldSkipRetrieval()` or its caller, avoid skipping when all of the following are true:

- this is the first user turn of a fresh session
- the message is weak / ultra-short / pure emoji
- there was a recent prior session for the same chat/session key

### Option B — continuity fallback before skip
If a fresh-session first turn is weak, try a small continuity recovery path before returning early:

- recent session tail
- latest session summary / reflection-derived continuity summary
- last-turn topic carryover

### Option C — dedicated config switch
Add something like:

```json
{
  "autoRecall": true,
  "continuityFirstTurnFallback": true
}
```

for users who want safer cross-restart continuity.

---

### In this repo
- #100 — autoRecall retrieval query not stripped of OpenClaw metadata — wrong memories returned
- #98 — memoryReflection does not reliably fire on `/new` / `/reset`
- #47 — normalizeQuery strips OpenClaw metadata and timestamps
- #50 — strip metadata noise in autoCapture and normalizeQuery

### In OpenClaw core
- openclaw/openclaw#18539 — Context primer not auto-injected on session restart after compaction
- openclaw/openclaw#33477 — continuity loss after provider restart causes fresh context
- openclaw/openclaw#33646 — recover continuity when session key lookup misses
- openclaw/openclaw#39049 — continuity context engine

This issue is related to the same continuity problem space, but is specifically about **adaptive autoRecall skip behavior on the first weak-signal message after restart/new session**.

sinskl:
## Summary

Fixes #101 — fresh-session continuity break when the first user message is a weak signal (emoji, `?`, `继续`, etc.).

## Problem

After a session restart, users often send low-signal continuation messages like `🤔`, `?`, `嗯`, or `继续`. The existing `shouldSkipRetrieval()` suppresses autoRecall for these, causing the assistant to appear to "forget" the previous conversation at exactly the moment when continuity matters most.

### `src/adaptive-retrieval.ts`
- Added `shouldSkipRetrievalFirstTurn()` — a relaxed variant that only skips retrieval for truly non-user messages:
  - Empty / whitespace-only
  - Slash commands (`/...`)
  - Heartbeat / system messages
- All other messages (including short text, pure emoji, affirmations) are allowed through on the first turn.

### `src/recall-engine.ts`
- `orchestrateDynamicRecall()` now detects whether the current turn is the first turn of a fresh session (`turnCounter === 1`).
- First turn → uses `shouldSkipRetrievalFirstTurn()` (relaxed)
- Subsequent turns → uses `shouldSkipRetrieval()` (unchanged behavior)

## Why this is safe

- The relaxed logic only applies to **turn 1** of a fresh session. All subsequent turns use the existing adaptive retrieval logic unchanged.
- The only messages skipped on first turn are system/infrastructure messages that would never need continuity recovery.
- No new config options needed — this is a behavioral improvement within the existing `autoRecall: true` flow.

## Testing

Verified locally that:
1. Fresh session + `🤔` → autoRecall now fires (previously skipped)
2. Fresh session + `/status` → still correctly skipped
3. Turn 2+ with weak messages → still uses standard skip logic (no behavior change)

### DIFF
--- a/test/adaptive-retrieval.test.mjs
+++ b/test/adaptive-retrieval.test.mjs
@@ -1,0 +1,142 @@
+import { describe, it } from "node:test";
+import assert from "node:assert/strict";
+import jitiFactory from "jiti";
+
+const jiti = jitiFactory(import.meta.url, { interopDefault: true });
+
+const {
+  shouldSkipRetrieval,
+  shouldSkipRetrievalFirstTurn,
+} = jiti("../src/adaptive-retrieval.ts");
+
+// ============================================================================
+// shouldSkipRetrieval — existing behavior (sanity checks)
+// ============================================================================
+
+describe("shouldSkipRetrieval", () => {
+  it("skips pure emoji", () => {
+    assert.equal(shouldSkipRetrieval("🤔"), true);
+    assert.equal(shouldSkipRetrieval("😂👍"), true);
+  });
+
+  it("skips very short messages", () => {
+    assert.equal(shouldSkipRetrieval("hi"), true);
+    assert.equal(shouldSkipRetrieval("?"), true);
+    assert.equal(shouldSkipRetrieval("ok"), true);
+  });
+
+  it("skips greetings", () => {
+    assert.equal(shouldSkipRetrieval("hello"), true);
+    assert.equal(shouldSkipRetrieval("good morning"), true);
+  });
+
+  it("skips continuation prompts", () => {
+    assert.equal(shouldSkipRetrieval("继续"), true);
+    assert.equal(shouldSkipRetrieval("go ahead"), true);
+    assert.equal(shouldSkipRetrieval("好的"), true);
+  });
+
+  it("skips slash commands", () => {
+    assert.equal(shouldSkipRetrieval("/status"), true);
+    assert.equal(shouldSkipRetrieval("/new"), true);
+  });
+
+  it("does not skip memory-related queries", () => {
+    assert.equal(shouldSkipRetrieval("你记得吗"), false);
+    assert.equal(shouldSkipRetrieval("do you remember my name"), false);
+    assert.equal(shouldSkipRetrieval("what did I say last time"), false);
+  });
+
+  it("does not skip meaningful questions", () => {
+    assert.equal(shouldSkipRetrieval("what is the capital of France?"), false);
+    assert.equal(shouldSkipRetrieval("帮我查一下明天的天气怎么样"), false);
+  });
+
+  it("strips OpenClaw metadata before evaluating", () => {
+    const wrapped = `Conversation info (untrusted metadata):\n\`\`\`json\n{"message_id":"123"}\n\`\`\`\n\n🤔`;
+    assert.equal(shouldSkipRetrieval(wrapped), true);
+  });
+});
+
+// ============================================================================
+// shouldSkipRetrievalFirstTurn — relaxed first-turn logic
+// ============================================================================
+
+describe("shouldSkipRetrievalFirstTurn", () => {
+  it("allows pure emoji through on first turn", () => {
+    assert.equal(shouldSkipRetrievalFirstTurn("🤔"), false);
+    assert.equal(shouldSkipRetrievalFirstTurn("😂"), false);
+    assert.equal(shouldSkipRetrievalFirstTurn("👍"), false);
+  });
+
+  it("allows short continuation messages on first turn", () => {
+    assert.equal(shouldSkipRetrievalFirstTurn("继续"), false);
+    assert.equal(shouldSkipRetrievalFirstTurn("嗯"), false);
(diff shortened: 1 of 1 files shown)


furedericca-lab:
Thanks — the continuity goal makes sense: first-turn weak signals after `/new` or `/reset` should not make the assistant feel like it forgot everything.

I do see a conflict with a recent fix on our side: system/control prompts must never trigger memory recall injection (for either generic auto-recall or reflection recall).

The current `shouldSkipRetrievalFirstTurn()` looks too permissive because it only skips:
- empty input
- slash commands
- heartbeat
- `[System]...`

but it does not appear to preserve the newer control-prompt exclusions such as:
- session-start boilerplate (`A new session was started via /new or /reset`)
- `Execute your Session Startup sequence now`
- wrapped `/note` handoff/control prompts

So in its current form, this may reintroduce control-prompt recall pollution on fresh-session turn 1.

My suggestion:
1. keep the first-turn relaxation for genuine weak user signals (`🤔`, `?`, `继续`, `嗯`, etc.)
2. factor system/control exclusions into a shared skip gate that both standard and first-turn paths must respect
3. add integration tests proving the same control prompt is skipped by both generic auto-recall and reflection recall
4. add first-turn tests proving weak user signals are allowed without allowing startup/control wrappers through
5. ensure the new tests are actually included in the main test command / CI path

AliceLJY:

PR #141, opened 2026-03-10
fix: respect framework-level memorySearch.enabled switch in autoRecall

### LINKED ISSUE #58: autoRecall 绕过了框架层 memorySearch.enabled 开关
## 问题描述

当 `agents.defaults.memorySearch.enabled = false` 时，插件仍然会执行自动召回。

## 配置情况

```json
// 框架层配置 - 已关闭
"agents": {
  "defaults": {
    "memorySearch": {
      "enabled": false
    }
  }
}

// 插件层配置 - 开启
"plugins": {
  "entries": {
    "memory-lancedb-pro": {
      "config": {
        "autoRecall": true
      }
    }
  }
}
```

## 预期行为

当框架层 `memorySearch.enabled = false` 时，插件应该尊重这个设置，不执行自动召回。

## 实际行为

插件在 `before_agent_start` 事件中直接执行召回，绕过了框架层的开关控制。

## 代码位置

`index.ts:469` - 插件直接监听 `before_agent_start` 事件执行召回，没有检查框架层的 `memorySearch.enabled` 状态。

## 建议

插件应该检查框架层的 `memorySearch.enabled` 配置，或者至少在文档中说明这两个开关的优先级关系，避免用户困惑。

## 环境

- OpenClaw: 2026.3.1
- memory-lancedb-pro: 1.0.30

eisen0419:
## Summary

Fixes #58 — `autoRecall` now respects the framework-level `memorySearch.enabled` switch.

Previously, when `agents.defaults.memorySearch.enabled = false` in OpenClaw config, the plugin still executed auto-recall because the `before_agent_start` handler did not check the framework-level setting.

## Changes

- Added `resolveFrameworkMemorySearchEnabled(cfg, agentId)` helper that reads `memorySearch.enabled` from `api.config`, checking agent-specific override first (`agents.list[].memorySearch.enabled`), then falling back to `agents.defaults.memorySearch.enabled`
- Added early return in `before_agent_start` handler when framework `memorySearch.enabled = false`, with a debug log message
- Moved `agentId` resolution earlier in the handler to support the check (no behavioral change for the existing code path)

## Precedence

| agents.defaults | agent-specific | Result |
|-----------------|---------------|--------|
| `false` | not set | skip recall |
| `false` | `true` | allow recall |
| `true` | `false` | skip recall |
| not set | not set | allow recall (unchanged) |

## Test plan

- [x] New standalone test file `test/issue-58-framework-memory-search.test.mjs` with 3 cases:
  - `defaults.memorySearch.enabled=false` → skips recall, 0 embedding requests
  - `defaults=false` + agent override `true` → allows recall
  - `defaults=true` + agent override `false` → skips recall
- [x] Existing `test/config-session-strategy-migration.test.mjs` still passes

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -574,6 +574,38 @@ function resolveAgentPrimaryModelRef(cfg: unknown, agentId: string): string | un
   return undefined;
 }
 
+function resolveFrameworkMemorySearchEnabled(cfg: unknown, agentId: string): boolean | undefined {
+  try {
+    const root = cfg as Record<string, unknown>;
+    const agents = root.agents as Record<string, unknown> | undefined;
+    const target = agentId.trim();
+    const list = agents?.list as unknown;
+
+    if (target && Array.isArray(list)) {
+      const found = list.find((x) => {
+        if (!x || typeof x !== "object") return false;
+        return (x as Record<string, unknown>).id === target;
+      }) as Record<string, unknown> | undefined;
+      const memorySearch = found?.memorySearch as Record<string, unknown> | undefined;
+      if (typeof memorySearch?.enabled === "boolean") return memorySearch.enabled;
+    }
+
+    const defaults = agents?.defaults as Record<string, unknown> | undefined;
+    const defaultMemorySearch = defaults?.memorySearch as Record<string, unknown> | undefined;
+    if (typeof defaultMemorySearch?.enabled === "boolean") return defaultMemorySearch.enabled;
+  } catch {
+    // ignore
+  }
+  return undefined;
+}
+
+function isFrameworkMemorySearchDisabled(
+  api: OpenClawPluginApi,
+  agentId: string,
+): boolean {
+  return resolveFrameworkMemorySearchEnabled(api.config, agentId) === false;
+}
+
 function isAgentDeclaredInConfig(cfg: unknown, agentId: string): boolean {
   const target = agentId.trim();
   if (!target) return false;
@@ -1985,6 +2017,14 @@ const memoryLanceDBProPlugin = {
     // Default is OFF to prevent the model from accidentally echoing injected context.
     if (config.autoRecall === true) {
       api.on("before_agent_start", async (event, ctx) => {
+        const agentId = resolveHookAgentId(ctx?.agentId, (event as any).sessionKey);
+        if (isFrameworkMemorySearchDisabled(api, agentId)) {
+          api.logger.debug?.(
+            `memory-lancedb-pro: skipping auto-recall because framework memorySearch.enabled=false for agent ${agentId}`,
+          );
+          return;
+        }
+
         if (
           !event.prompt ||
           shouldSkipRetrieval(event.prompt, config.autoRecallMinLength)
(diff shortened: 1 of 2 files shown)


AliceLJY:
## Review: fix: respect framework-level memorySearch.enabled switch in autoRecall

**Verdict: Fix-then-merge** — same blocking item as #148.

---

### ✅ What's working

- Priority chain is correct: agent-level override → `agents.defaults` → `undefined` (allow recall). Existing users with no `memorySearch` config see zero behavior change
- Early return placement is right — framework permission check happens *before* `shouldSkipRetrieval()`, not after
- Test infrastructure is solid: real HTTP embedding server mock makes the assertions trustworthy (no false passes from over-mocking)
- All 3 test cases pass, including the agent-override-wins scenarios
- `agentId` move to top of handler is a clean refactor; `resolveHookAgentId` is pure, no side effects

### 🔴 Blocking (same issue as #148)

**New test file is not wired into `npm test`.** `issue-58-framework-memory-search.test.mjs` is missing from `package.json` scripts — CI never runs it.

Fix: append to the test script in `package.json`:
```
&& node --test test/issue-58-framework-memory-search.test.mjs
```

### 🟡 Suggested before merge

**Missing backward-compat test** — the most common real-world scenario (user upgrades without adding `memorySearch` to config) has no test coverage. Adding one case would close the gap:

```js
test("issue #58: no memorySearch config → recall proceeds normally", async () => {
  const outcome = await runAutoRecallScenario({}); // empty OpenClaw config
  assert.ok(outcome.requests > 0);
});
```

### ⚪ Non-blocking

- `NODE_PATH` hardcodes `/opt/homebrew` — harmless in Docker (jiti resolves without it), but worth making portable
- `catch { // ignore }` in `resolveFrameworkMemorySearchEnabled` is intentionally fail-open; a short comment would help future readers

---

### Merge order note

This should merge before #123 (fresh-session continuity fix) since both touch the autoRecall skip logic. #148 can go in either order — different files.

Good fix. The two-level config lookup matches how OpenClaw models other agent settings (e.g. `resolveAgentPrimaryModelRef` follows the same pattern).

chenjiyong:
Yes, PR 141 has value for `master`, but the value is in the fix itself, not in merging the PR branch as-is.

What I verified:

- The PR fixes a real behavior gap: when framework-level `memorySearch.enabled=false`, the plugin should skip auto-recall. That logic is added in [index.ts](/Users/chenjiyong/learning/memory-lancedb-pro/validate-memory-lancedb-pro-pr/pr-141/index.ts#L577) and [index.ts](/Users/chenjiyong/learning/memory-lancedb-pro/validate-memory-lancedb-pro-pr/pr-141/index.ts#L2018).
- On the original PR 141 head (`55eadd5`), the OpenClaw validation flow passed, including both `openclaw host functional` and `openclaw agent e2e`.
- The new regression test in [test/issue-58-framework-memory-search.test.mjs](/Users/chenjiyong/learning/memory-lancedb-pro/validate-memory-lancedb-pro-pr/pr-141/test/issue-58-framework-memory-search.test.mjs#L145) also passed all 3 subtests.

What also matters:

- PR 141 is not current with `cortex/master`. It is `1 ahead / 22 behind`.
- On the PR branch itself, `npm test` fails at [test/smart-extractor-branches.mjs](/Users/chenjiyong/learning/memory-lancedb-pro/validate-memory-lancedb-pro-pr/pr-141/test/smart-extractor-branches.mjs#L470) with `3 !== 4`.
- I then cherry-picked the PR 141 commit onto the current `cortex/master` in a separate worktree, and `npm test` passed there.

So the conclusion is: PR 141 does have value for `master`, because the fix appears valid and passes when applied on top of current `master`. But the PR branch itself is stale, so the value is in bringing the fix forward, not in merging the outdated branch unchanged.

PR #148, opened 2026-03-10
fix: skip autoRecall for heartbeat/NO_REPLY turns

### LINKED ISSUE #137: Heartbeat / NO_REPLY turns may still trigger autoRecall injection and cause compaction-related latency
## Summary

We observed severe reply delays in OpenClaw Web UI that were strongly correlated with `heartbeat` turns and `memory-lancedb-pro` auto-recall injection.

After investigation, the issue appears to be caused by a combination of:

- heartbeat turns being processed too much like normal conversation turns
- `memory-lancedb-pro` auto-injecting `<relevant-memories>` during `before_agent_start`
- compaction being triggered but then cancelled by safeguard because there are "no real conversation messages to summarize"

Disabling **heartbeat** and setting `autoRecall: false` made the system recover immediately: no more repeated `compacting context`, and normal reply latency returned.

### 1) Heartbeat turns seem to enter the normal prompt-building path
Our heartbeat prompt is `NO_REPLY`.

In practice, these heartbeat turns did **not** behave like a minimal health-check / `HEARTBEAT_OK` path. Instead, they seemed to be treated like normal turns, and we could see suspiciously long context being built around them.

### 2) Auto-recall appears to inject memories even for heartbeat-like turns
We observed that `NO_REPLY` heartbeat turns appeared to receive LanceDB memory injection, resulting in unexpectedly long prompt context.

This is especially surprising because heartbeat turns are low-value system pulses and should ideally not trigger long-term memory recall at all.

### 3) Compaction safeguard repeatedly cancelled compaction
We captured logs like:

```text
Compaction safeguard: cancelling compaction with no real conversation messages to summarize.
memory-lancedb-pro: injecting 3 memories into context for agent main
```

This suggests a bad interaction:

1. context grows
2. compaction triggers
3. safeguard sees too few real user/assistant messages
4. compaction is cancelled
5. auto-recall continues injecting memories
6. latency becomes minutes long

## Related existing issues

This seems related in spirit to several existing issues:

- #13 `periodically showing long-term memories in replies`
- #58 `autoRecall bypasses framework-level memorySearch.enabled`
- #85 `Make autoRecall injection safer and less likely to leak verbatim`
- #100 `autoRecall retrieval query not stripped of OpenClaw metadata`

So while this report is specifically about heartbeat interaction, it may share the same root family: **autoRecall boundary/gating is too broad, and metadata/system turns can pollute recall behavior**.

## Current workaround

The following workaround restored normal behavior for us:

```json
{
  "agents": {
    "defaults": {
      "heartbeat": {
        "every": "0",
        "model": "aihubmix-router/gpt-5-nano",
        "prompt": "NO_REPLY",
        "lightContext": true
      }
    }
  },
  "plugins": {
    "entries": {
      "memory-lancedb-pro": {
        "config": {
          "autoRecall": false
        }
      }
    }
  }
}
```

## Suggested fix / expected behavior

I think heartbeat / system-pulse turns should be excluded from generic auto-recall by default.

Suggested behavior:

1. **Do not trigger `autoRecall` for heartbeat / `NO_REPLY` / system-event turns**
2. **Prefer `lightContext: true` or equivalent minimal context path for heartbeat**
3. **Strip framework metadata before any recall gating / retrieval query construction**
4. **Improve logs when compaction is cancelled**, e.g. indicate how much of recent context comes from system events / memory injections / heartbeat turns

## Why this matters

Heartbeat is supposed to be a tiny health pulse. If it can trigger memory injection and normal prompt-building, it can become a silent context amplifier that degrades the main session over time.

In our case, turning off `autoRecall` solved the practical problem. But if heartbeat still enters the normal path, there may still be a framework-side issue worth fixing.

Happy to provide more environment details or logs if useful.

eisen0419:
## Summary
Heartbeat / `NO_REPLY` turns were still triggering `autoRecall` injection, causing compaction-related latency spikes (minutes-long delays).

## Changes
- **`src/adaptive-retrieval.ts`**:
  - Added `NO_REPLY`, `HEARTBEAT_OK`, `health_check`, `system_check` and variants to `SKIP_PATTERNS`
  - Added heartbeat/healthcheck wrapper stripping in `normalizeQuery()` (e.g. `[heartbeat] ...`, `heartbeat: ...`)
- **`test/adaptive-retrieval-skip-heartbeat.test.mjs`**: 9 test cases covering all heartbeat variants
- **`package.json`**: added new test to `npm test` pipeline

## Root cause
`shouldSkipRetrieval()` already had `/HEARTBEAT/i` but missed:
- `NO_REPLY` (common heartbeat prompt)
- `HEARTBEAT_OK` / `heartbeat ok` variants
- `health_check` / `system_check` patterns
- Prefixed formats like `[heartbeat] NO_REPLY` or `heartbeat: NO_REPLY`

## Test plan
- [x] All 9 new heartbeat skip tests pass
- [ ] Verify existing tests still pass (note: `smart-extractor-branches.mjs` has a pre-existing failure unrelated to this change)

Closes #137

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     ]
   },
   "scripts": {
-    "test": "node test/embedder-error-hints.test.mjs && node test/migrate-legacy-schema.test.mjs && node --test test/config-session-strategy-migration.test.mjs && node test/update-consistency-lancedb.test.mjs && node test/cli-smoke.mjs && node test/functional-e2e.mjs && node test/retriever-rerank-regression.mjs && node test/smart-memory-lifecycle.mjs && node test/smart-extractor-branches.mjs && node test/plugin-manifest-regression.mjs",
+    "test": "node test/embedder-error-hints.test.mjs && node test/migrate-legacy-schema.test.mjs && node --test test/config-session-strategy-migration.test.mjs && node test/update-consistency-lancedb.test.mjs && node test/cli-smoke.mjs && node test/functional-e2e.mjs && node test/retriever-rerank-regression.mjs && node test/smart-memory-lifecycle.mjs && node test/smart-extractor-branches.mjs && node test/plugin-manifest-regression.mjs && node --test test/adaptive-retrieval-skip-heartbeat.test.mjs",
     "test:openclaw-host": "node test/openclaw-host-functional.mjs"
   },
   "devDependencies": {
--- a/src/adaptive-retrieval.ts
+++ b/src/adaptive-retrieval.ts
@@ -20,6 +20,8 @@ const SKIP_PATTERNS = [
   /^[\p{Emoji}\s]+$/u,
   // Heartbeat/system (match anywhere, not just at start, to handle prefixed formats)
   /HEARTBEAT/i,
+  /^(no[_-]?reply|noreply)\s*[.!]?$/i,
+  /^(heartbeat(_ok|_fail|_err|_error)?|heartbeat\s*(ok|fail|error)|health[_ -]?check|system[_ -]?check|status[_ -]?check)\s*[.!]?$/i,
   /^\[System/i,
   // Single-word utility pings
   /^(ping|pong|test|debug)\s*[.!?]?$/i,
@@ -56,6 +58,10 @@ function normalizeQuery(query: string): string {
 
   // 3. Strip OpenClaw timestamp prefix [Mon 2026-03-02 04:21 GMT+8].
   s = s.trim().replace(/^\[[A-Za-z]{3}\s\d{4}-\d{2}-\d{2}\s\d{2}:\d{2}\s[^\]]+\]\s*/, "");
+
+  // 4. Strip heartbeat/healthcheck wrappers.
+  s = s.trim().replace(/^\[(heartbeat|health[-_ ]?check|system[-_ ]?check)[^\]]*\]\s*/i, "");
+  s = s.trim().replace(/^(heartbeat|health[-_ ]?check|system[-_ ]?check)\s*[:\-]\s*/i, "");
 
   const result = s.trim();
   return result;
--- a/test/adaptive-retrieval-skip-heartbeat.test.mjs
+++ b/test/adaptive-retrieval-skip-heartbeat.test.mjs
@@ -1,0 +1,27 @@
+import { describe, it } from "node:test";
+import assert from "node:assert/strict";
+import jitiFactory from "jiti";
+
+const jiti = jitiFactory(import.meta.url, { interopDefault: true });
+
+const { shouldSkipRetrieval } = jiti("../src/adaptive-retrieval.ts");
+
+describe("shouldSkipRetrieval heartbeat/NO_REPLY", () => {
+  const cases = [
+    "NO_REPLY",
+    "no_reply",
+    "no-reply",
+    "HEARTBEAT_OK",
+    "heartbeat ok",
+    "heartbeat: NO_REPLY",
+    "[heartbeat] NO_REPLY",
+    "health_check",
+    "system-check",
+  ];
+
+  for (const input of cases) {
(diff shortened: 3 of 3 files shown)


eisen0419:
## CI failure note

The CI failure is a **pre-existing issue on `master`**, unrelated to this PR's changes.

The `plugin-manifest-regression.mjs` test fails because `openclaw.plugin.json` has version `1.1.0-beta.4` while `package.json` has `1.1.0-beta.5`:

```
AssertionError [ERR_ASSERTION]: openclaw.plugin.json version should stay aligned with package.json
+ '1.1.0-beta.4'
- '1.1.0-beta.5'
```

This version mismatch exists on the current `master` branch and affects all PRs based on it.

AliceLJY:
## Review: fix: skip autoRecall for heartbeat/NO_REPLY turns

**Verdict: Fix-then-merge** — one blocking item before merge.

---

### ✅ What's working

- All 9 new unit tests pass (`adaptive-retrieval-skip-heartbeat.test.mjs`)
- New test file is correctly wired into `npm test` pipeline
- `normalizeQuery()` + SKIP_PATTERNS two-layer design is clean: strip wrapper prefix first, then pattern-match the residual
- The existing unanchored `/HEARTBEAT/i` acts as a safety net for compound strings; the new anchored patterns complement it without conflicting
- Dynamic test (simulated via Discord): 3/4 heartbeat-style messages correctly skipped recall

### 🔴 Blocking (please fix before merge)

**No negative test cases.** The test suite has 9 "should skip" cases but zero "should NOT skip" cases. Without negative coverage, a future regex change could silently over-match real user messages and no test would catch it.

Suggested additions to `adaptive-retrieval-skip-heartbeat.test.mjs`:

```js
// These should NOT be skipped
assert.equal(shouldSkipRetrieval("I got no reply from the server"), false);
assert.equal(shouldSkipRetrieval("noreply@example.com"), false);
assert.equal(shouldSkipRetrieval("system check required before deployment"), false);
assert.equal(shouldSkipRetrieval("my health check results came back"), false);
```

### 🟡 Non-blocking notes

1. **`no reply` (space variant) not covered** — `no[_-]?reply` matches `no_reply`/`no-reply` but not `no reply`. In practice this falls through to the 15-char length filter so it won't cause latency, but worth a comment noting the intentional gap.

2. **Pre-existing CI failure** — `plugin-manifest-regression` fails due to `openclaw.plugin.json` (v1.1.0-beta.4) vs `package.json` (v1.1.0-beta.5) version mismatch on `master`. Unrelated to this PR; should be fixed on `master` separately.

3. **Merge order** — PRs #141 and #123 both touch `shouldSkipRetrieval()` logic. Suggest merging this one first as it's the most self-contained.

---

Good catch on the `HEARTBEAT_OK` / `health_check` gaps — the original `/HEARTBEAT/i` was clearly not enough. Surgical fix, proportional to the problem.

chenjiyong:
Thanks for the fix. I validated PR #148 against the latest `master`.

What I confirmed:
- The adaptive retrieval change is useful. It fixes cases that current `master` still misses, especially wrapped heartbeat/system inputs such as:
  - `[health_check worker=main run=42] NO_REPLY`
  - `health_check: NO_REPLY`
  - `system_check: NO_REPLY`
  - `[Mon 2026-03-02 04:21 GMT+8] [health_check worker=main] NO_REPLY`
- The OpenClaw host functional harness passed.
- The real `openclaw agent` E2E also passed.

What still blocks merge:
- The PR does not merge cleanly into current `master` because `package.json` now conflicts with the updated test script on `master`.
- The new test file does not cover the behavior that actually changed. The 9 added cases already pass on current `master`, so they would not catch the regression this PR is fixing.

I think this should still land, but it needs to be refreshed on top of latest `master` and the tests should include the wrapped `health_check` / `system_check` cases above.

PR #149, opened 2026-03-10
fix: add structural noise filtering for system traces and raw blobs

### LINKED ISSUE #127: Noise filter misses structural memory contamination at write time (System traces / raw blobs / fragments)
## Summary
Noise filtering in `memory-lancedb-pro` appears to be too narrow at write time: it can block greetings / denials / meta-questions, but still allows structurally noisy memories into LanceDB, such as:
- `System:` / compaction / model-switch traces
- long raw quoted conversation blobs
- concatenated fragment strings / malformed snippets
- low-value duplicate-ish user utterances that were not distilled into atomic memory entries

This makes the retrieval engine look “healthy” while the memory corpus itself slowly accumulates contaminants.

## Observed Symptoms
In real-world usage, recent memories included examples like:
- quoted raw user text stored nearly verbatim
- system traces such as compaction/model-switch remnants ending up in memory rows
- concatenated fragment-like entries such as mixed filename/text shards

These were not simple greeting/boilerplate cases, so the current `noise-filter.ts` patterns did not catch them.

## Why this matters
The plugin’s retrieval stack is strong (hybrid retrieval, rerank, decay, normalization, diversity), but corpus quality still depends on ingress quality.
If ingress filtering is too weak, the plugin can remain operational while recall quality degrades over time because the stored memories are not clean / atomic / semantically stable.

## Current likely gap
From reading the code/docs, the likely ingress points are:
1. `src/tools.ts` → `memory_store`
2. `index.ts` auto-capture path (`agent_end` hook) before final persistence

`src/noise-filter.ts` currently focuses on:
- greetings / boilerplate
- denials
- meta-questions

But it does **not** appear to explicitly reject:
- `System:`-prefixed traces or internal runtime artifacts
- compaction / session-management / model-switch messages
- overly long raw conversation blobs
- malformed concatenated fragments / accidental shards
- “not yet distilled” entries that should have been compressed into a short fact/decision/preference memory instead of being stored verbatim

## Suggestion
Consider adding a stricter **ingress hygiene layer** before persistence, applied both to manual tool-store and auto-capture:

### 1) Source/artifact rejection
Reject entries matching patterns like:
- `^System:`
- compaction / model switched / session reset / tool transcript artifacts
- known internal control markers / tags

### 2) Atomicity / length gate
Reject or require transformation when:
- text is over a configurable character threshold
- contains multi-sentence raw dialogue / quote blocks
- contains suspicious concatenation signatures / repeated quote wrapping / filename-shard blobs

### 3) Distillation gate
For auto-capture especially, require candidate memory items to resemble atomic memory forms, e.g.:
- preference
- fact (pitfall/cause/fix/prevention)
- decision principle
- entity

Instead of allowing near-verbatim conversation carryover.

### 4) Optional config flags
Example ideas:
- `store.rejectSystemArtifacts: true`
- `store.maxRawLength: 500`
- `store.requireAtomicMemoryShape: true`
- `store.rejectConversationBlob: true`
- `store.rejectMalformedFragments: true`

### 5) Post-write verification hook (optional)
An optional callback / validation stage that checks whether the stored item is likely retrievable and non-noisy before accepting it permanently.

## Key point
This seems less like a retrieval problem and more like an **ingress validation / corpus hygiene** problem.

The current noise filter is useful, but from field behavior it feels optimized for **surface-level low-quality chatter**, not **structural memory contamination**.

If helpful, I can also prepare a concrete patch proposal for:
- `src/tools.ts` write-time validation
- `index.ts` auto-capture final gate
- new `noise-filter.ts` heuristics for system artifacts / raw blob rejection

eisen0419:
## Summary
`noise-filter.ts` previously only filtered surface-level noise (greetings, denials, meta-questions), but missed **structural memory contamination**: runtime traces, raw conversation blobs, and metadata wrappers that silently degrade memory corpus quality over time.

## Changes
- **`src/noise-filter.ts`**: Added `STRUCTURAL_NOISE_PATTERNS` array covering:
  - `System:` prefixed runtime messages
  - Compaction/model-switch/session-management traces
  - OpenClaw `(untrusted metadata):` wrapper remnants
  - Pure JSON objects (`{...}`)
  - XML-wrapped content (`<tag>...</tag>`)
  - Multi-line blockquote blobs (>= 3 quoted lines)
- New `filterStructuralNoise` option in `NoiseFilterOptions` (default: `true`)
- **`test/noise-filter-structural.test.mjs`**: 10 test cases covering all new patterns + opt-out behavior

## Test plan
- [x] All 10 new structural noise tests pass
- [x] Existing `isNoise()` behavior preserved (new patterns are additive)
- [x] Option `filterStructuralNoise: false` disables the new filters

Closes #127

### DIFF
--- a/src/noise-filter.ts
+++ b/src/noise-filter.ts
@@ -50,6 +50,18 @@ const DIAGNOSTIC_ARTIFACT_PATTERNS = [
   /\bno explicit solution\b/i,
 ];
 
+// Structural noise: runtime traces, raw blobs, wrappers, and malformed fragments
+const STRUCTURAL_NOISE_PATTERNS = [
+  /^System\s*:/i,
+  /compacti(ng|on)\s*(context|safeguard)/i,
+  /model\s*(switch|switched|changed|swap)/i,
+  /session\s*(reset|restart|start|end)/i,
+  /\(untrusted metadata\)\s*:/i,
+  /^\{[\s\S]*\}$/,
+  /<[a-z-]+>[\s\S]*<\/[a-z-]+>/i,
+  /(?:^>.*\n){3,}/m,
+];
+
 export interface NoiseFilterOptions {
   /** Filter agent denial responses (default: true) */
   filterDenials?: boolean;
@@ -57,12 +69,15 @@ export interface NoiseFilterOptions {
   filterMetaQuestions?: boolean;
   /** Filter session boilerplate (default: true) */
   filterBoilerplate?: boolean;
+  /** Filter structural noise (default: true) */
+  filterStructuralNoise?: boolean;
 }
 
 const DEFAULT_OPTIONS: Required<NoiseFilterOptions> = {
   filterDenials: true,
   filterMetaQuestions: true,
   filterBoilerplate: true,
+  filterStructuralNoise: true,
 };
 
 /**
@@ -78,6 +93,7 @@ export function isNoise(text: string, options: NoiseFilterOptions = {}): boolean
   if (opts.filterDenials && DENIAL_PATTERNS.some(p => p.test(trimmed))) return true;
   if (opts.filterMetaQuestions && META_QUESTION_PATTERNS.some(p => p.test(trimmed))) return true;
   if (opts.filterBoilerplate && BOILERPLATE_PATTERNS.some(p => p.test(trimmed))) return true;
+  if (opts.filterStructuralNoise && STRUCTURAL_NOISE_PATTERNS.some(p => p.test(trimmed))) return true;
   if (DIAGNOSTIC_ARTIFACT_PATTERNS.some(p => p.test(trimmed))) return true;
 
   return false;
--- a/test/noise-filter-structural.test.mjs
+++ b/test/noise-filter-structural.test.mjs
@@ -1,0 +1,32 @@
+import { describe, it } from "node:test";
+import assert from "node:assert/strict";
+import jitiFactory from "jiti";
+
+const jiti = jitiFactory(import.meta.url, { interopDefault: true });
+
+const { isNoise } = jiti("../src/noise-filter.ts");
+
+describe("noise-filter structural patterns", () => {
+  const samples = [
+    "System: compaction safeguard engaged",
+    "Compaction context safeguard tripped",
+    "model switch detected",
+    "model changed to gpt-5",
+    "session reset due to inactivity",
+    "(untrusted metadata): Sender (untrusted metadata): foo",
+    "{\"type\":\"meta\",\"payload\":{\"note\":\"wrapper\"}}",
+    "<relevant-memories>foo</relevant-memories>",
+    "> quote one\n> quote two\n> quote three\n> quote four\n",
+  ];
+
+  for (const input of samples) {
+    it(`filters structural noise: ${input.slice(0, 40)}`, () => {
+      assert.equal(isNoise(input), true);
+    });
+  }
+
+  it("allows structural noise when disabled", () => {
+    const input = "System: model switch detected";
+    assert.equal(isNoise(input, { filterStructuralNoise: false }), false);
+  });
+});


eisen0419:
## CI failure note

The CI failure is a **pre-existing issue on `master`**, unrelated to this PR's changes.

The `plugin-manifest-regression.mjs` test fails because `openclaw.plugin.json` has version `1.1.0-beta.4` while `package.json` has `1.1.0-beta.5`:

```
AssertionError [ERR_ASSERTION]: openclaw.plugin.json version should stay aligned with package.json
+ '1.1.0-beta.4'
- '1.1.0-beta.5'
```

This version mismatch exists on the current `master` branch and affects all PRs based on it.

AliceLJY:
## Review: fix: add structural noise filtering for system traces and raw blobs

**Verdict: Fix-then-merge** — two pattern-level issues need addressing before this is safe to ship.

---

### ✅ What's working

- `filterStructuralNoise` opt-out flag is correctly wired and tested
- `compacti(ng|on)\s*(context|safeguard)` and `(untrusted metadata)\s*:` are highly specific — negligible false positive risk
- `/^\{[\s\S]*\}$/` anchored to full string correctly targets raw JSON blob memories
- `/(?:^>.*\n){3,}/m` (3+ blockquote lines) is reasonable
- All 10 tests pass

### 🔴 Blocking

**1. `npm test` not updated** — `noise-filter-structural.test.mjs` missing from `package.json` scripts.

**2. XML/HTML pattern is not anchored and will false-positive on real user content**

`/<[a-z-]+>[\s\S]*<\/[a-z-]+>/i` matches *anywhere* in a string and doesn't require matching tags. Any memory mentioning HTML/XML gets silently dropped:

```
"I learned that <div>content</div> structures HTML"   → filtered ❌
"response wrapped in <answer>...</answer> tags"        → filtered ❌
"The <tool_call> format used by Claude"               → filtered ❌ (if any closing tag exists later)
```

For the intended use case (filtering OpenClaw's `<relevant-memories>` injection), a targeted pattern is safer:
```js
/^<(relevant-memories|memory-context|system-context)>[\s\S]*<\/\1>/i
```
Or anchor to the full string (`^` + `$`) if broader XML filtering is intended.

### 🟡 Suggested before merge

**3. `model` and `session` patterns lack word boundaries — risk of over-filtering**

`/model\s*(switch|switched|changed|swap)/i` and `/session\s*(reset|restart|start|end)/i` have no anchors and no `\b` boundaries. Real user memories that would be silently dropped:

```
"Our business model changed after the pivot"  → matches /model\s*changed/  ❌
"The session started at 9am"                  → matches /session\s*start/  ❌
  ("started" begins with "start" — no word boundary)
"We need to model switching costs"            → matches /model\s*switch/   ❌
```

Suggested fix:
```js
/\bmodel\s+(switched?|changed|swapped?)\b/i
/\bsession\s+(reset|restarted?|started?|ended?)\b/i
```

**4. No negative test cases for the three risky patterns** — especially important here given the false positive surface.

---

The `compaction` and `untrusted metadata` patterns are solid. The JSON blob pattern is clean. The XML and model/session patterns need scoping before this is ready to ship.

AliceLJY:
**Correction to my previous comment** — the suggested regex fix was wrong on two counts:

1. `switched?` does not match `"model switch"` (misses true positive)
2. `started?` still matches `"session started"` (false positive remains)

The correct minimal fix is to add `\b` at the **end** of each pattern without changing the word list:

```js
/\bmodel\s+(switch|switched|changed|swap)\b/i
/\bsession\s+(reset|restart|start|end)\b/i
```

With this:
- `"model switching costs"` → no match ✅ (`\b` fails mid-word)
- `"session started at 9am"` → no match ✅ (`\b` fails because `ed` follows)
- `"model switch detected"` → match ✅
- `"session reset"` → match ✅

One remaining limitation: `"Our business model changed after the pivot"` still matches `/model\s+changed\b/` because `model changed` happens to appear as a valid word sequence. This is a semantic ambiguity that `\b` alone cannot resolve — worth noting as an accepted trade-off or handled with a deny-list of preceding words (e.g. `business model`, `mental model`).

chenjiyong:
Please close this PR for now and rebase it onto the latest `master`, then open a new PR. I tested the current branch against today’s `master` on March 12, 2026, and the branch is too far behind to evaluate merge value accurately in its current state.

PR #151, opened 2026-03-10
fix: allow apiKey omission for local embedding providers

### LINKED ISSUE #133: Follow-up: harden no-apiKey embedding path for local providers
## Summary

PR #93 fixed issue #71 by making `embedding.apiKey` optional so local providers like Ollama can start without authentication. The core fix is useful and should remain, but there are two follow-up areas worth tightening.

## Follow-up items

1. Add stable test coverage for the no-apiKey path
- The PR added `test/ollama-no-apikey.test.mjs`, but it depends on a local Ollama instance.
- Add unit-level coverage that does not require a running local server, so CI can reliably lock in the expected config/embedder behavior.

2. Revisit the `no-key-required` fallback strategy
- Today, missing `embedding.apiKey` falls back to a dummy value and emits a warning.
- This is convenient for local providers, but it also changes failure behavior for cloud providers from config-time failure to runtime failure.
- Consider a more provider-aware path so local providers can omit keys cleanly while cloud providers still fail early and clearly.

## Context

This is not a regression report against PR #93. It is a follow-up hardening task so the Ollama/local-provider fix has better long-term test and behavior guarantees.

eisen0419:
## Summary
Follow-up to PR #93 — hardens the no-apiKey path with provider-aware validation and unit test coverage that doesn't require a running local server.

## Changes
- **`index.ts`**: Added `isLocalProvider()` helper to detect local endpoints (localhost/127.0.0.1/[::1]/0.0.0.0)
  - Local providers: apiKey is optional, falls back to dummy `'ollama'` key with console warning
  - Cloud providers: fail early with clear error message at config time
- **`test/no-apikey-local-provider.test.mjs`**: 4 test cases covering:
  - Local provider without apiKey → uses dummy key
  - Cloud provider without apiKey → throws
  - Local provider with explicit apiKey → uses provided key
  - No baseURL + no apiKey → throws

## Test plan
- [x] All 4 new tests pass (no local server required)
- [x] Existing behavior preserved for all configured providers
- [ ] CI note: `plugin-manifest-regression` failure is a pre-existing version mismatch on `master`

Closes #133

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -205,6 +205,11 @@ function parsePositiveInt(value: unknown): number | undefined {
   return undefined;
 }
 
+function isLocalProvider(baseURL?: string): boolean {
+  if (!baseURL) return false;
+  return /localhost|127\.0\.0\.1|\[::1\]|0\.0\.0\.0/i.test(baseURL);
+}
+
 function resolveHookAgentId(
   explicitAgentId: string | undefined,
   sessionKey: string | undefined,
@@ -3160,8 +3165,12 @@ export function parsePluginConfig(value: unknown): PluginConfig {
     throw new Error("embedding config is required");
   }
 
+  const resolvedBaseURL =
+    typeof embedding.baseURL === "string" ? resolveEnvVars(embedding.baseURL) : undefined;
+  const isLocal = isLocalProvider(resolvedBaseURL);
+
   // Accept single key (string) or array of keys for round-robin rotation
-  let apiKey: string | string[];
+  let apiKey: string | string[] | undefined;
   if (typeof embedding.apiKey === "string") {
     apiKey = embedding.apiKey;
   } else if (Array.isArray(embedding.apiKey) && embedding.apiKey.length > 0) {
@@ -3183,7 +3192,16 @@ export function parsePluginConfig(value: unknown): PluginConfig {
   }
 
   if (!apiKey || (Array.isArray(apiKey) && apiKey.length === 0)) {
-    throw new Error("embedding.apiKey is required (set directly or via OPENAI_API_KEY env var)");
+    if (isLocal) {
+      apiKey = "ollama";
+      console.warn(
+        "[memory-lancedb-pro] No embedding.apiKey provided for local provider; using dummy key. This is expected for Ollama and similar local endpoints.",
+      );
+    } else {
+      throw new Error(
+        "embedding.apiKey is required for cloud providers (set directly or via OPENAI_API_KEY env var)",
+      );
+    }
   }
 
   const memoryReflectionRaw = typeof cfg.memoryReflection === "object" && cfg.memoryReflection !== null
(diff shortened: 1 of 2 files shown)


AliceLJY:
## Review: fix: allow apiKey omission for local embedding providers

**Verdict: Fix-then-merge** — two items to address.

---

### ✅ What's working

- `isLocalProvider()` correctly handles all four loopback forms: `localhost`, `127.0.0.1`, `[::1]`, `0.0.0.0`
- `resolveEnvVars()` is called on `baseURL` before the check — env-var-based URLs (e.g. `${LOCAL_EMBEDDING_URL}`) resolve correctly
- Fallback dummy key `"ollama"` is a reasonable default, warning is clear
- Cloud provider path still throws — no regression for existing users
- All 4 tests pass; test covers the key branches without requiring a running server

### 🔴 Blocking

**`npm test` not updated** — `no-apikey-local-provider.test.mjs` is missing from `package.json` scripts. Third PR in a row with this issue.

Fix: append `&& node --test test/no-apikey-local-provider.test.mjs` to the test script.

### 🟡 Suggested before merge

**`openclaw.plugin.json` schema not updated.** `embedding.required` still includes `"apiKey"`, but the fix makes it optional for local providers. This causes a mismatch between the schema (used by UI and docs) and the actual runtime behavior — users reading the schema will think apiKey is always mandatory.

Suggested fix: make `apiKey` optional in the schema and update the `uiHints` description:
```json
// Remove "apiKey" from embedding.required (or remove required entirely)
// Update uiHints.embedding.apiKey.help to mention local providers explicitly
```

### ⚪ Non-blocking

`isLocalProvider()` uses `.test()` on the full URL string — a baseURL like `http://myserver-localhost.corp:8080` would match `localhost` and be treated as local (skipping apiKey check). Low real-world risk since embedding baseURLs are almost always clean loopback addresses, but anchoring to the hostname portion would be safer:
```js
/^https?:\/\/(localhost|127\.0\.0\.1|\[::1\]|0\.0\.0\.0)(:\d+)?(\\/|$)/i
```

---

Clean fix overall. The `resolveEnvVars` call on `baseURL` before the local-provider check is a nice touch.

chenjiyong:
Thanks. I verified PR #151 locally.

The change itself looks useful: it makes `embedding.apiKey` optional for local embedding endpoints (`localhost` / `127.0.0.1` / `[::1]` / `0.0.0.0`) while still failing early for cloud providers, and the new unit test passes.

For my local OpenClaw setup, the direct impact is limited because I already have an explicit dummy `embedding.apiKey` configured for a local `127.0.0.1` embedding endpoint, so behavior does not change on my side. It is still a worthwhile ergonomics improvement for users of local embedding servers.

One note: the PR branch is currently 24 commits behind `master`. Validating the PR branch as-is failed an agent E2E in my environment, but cherry-picking this commit onto the latest `master` passed both host functional and agent E2E. So the patch looks fine to me, but it would be better to rebase onto the latest `master` before merge.

PR #165, opened 2026-03-11
feat(retriever): upgrade applyTimeDecay to Weibull decay

AliceLJY:
## Summary

Upgrade `applyTimeDecay()` from simple exponential to **Weibull stretched-exponential** decay, using tier-aware shape parameters.

This benefits users who haven't enabled the full smart-memory lifecycle (`DecayEngine`). The existing `decay-engine.ts` already implements Weibull with tier-specific β values, but `applyTimeDecay` (the fallback path) was still using `0.5 + 0.5 * exp(-t/HL)`.

### What changed

| Tier | β (shape) | Behavior | Use case |
|------|-----------|----------|----------|
| Core | 0.8 | Sub-exponential (slow forgetting) | Important, pinned memories |
| Working | 1.0 | Standard exponential | Default — **identical to current behavior** |
| Peripheral | 1.3 | Super-exponential (fast forgetting) | Casual, low-importance memories |

- Entries without `tier` in metadata default to `"working"` (β=1.0), so **existing behavior is preserved exactly**.
- Access reinforcement via `computeEffectiveHalfLife()` is unchanged.
- Tier β values are kept inline (not imported from `decay-engine.ts`) to avoid coupling the retriever to the full decay engine module.

### Background

This approach has been running in [RecallNest](https://github.com/AliceLJY/recallnest) (v1.3.0+) as the default time decay method, validated against 32,000+ memory entries across Claude Code, Codex, and Gemini CLI transcripts.

## Test plan

- [ ] Entries without tier metadata should behave identically to before (β=1.0 = simple exponential)
- [ ] Entries with `tier: "core"` should decay slower than working/peripheral
- [ ] Entries with `tier: "peripheral"` should decay faster
- [ ] Access reinforcement (effective half-life extension) still applies correctly
- [ ] No new dependencies introduced

### DIFF
--- a/src/retriever.ts
+++ b/src/retriever.ts
@@ -136,6 +136,32 @@ function clamp01WithFloor(value: number, floor: number): number {
 }
 
 // ============================================================================
+// Weibull Tier Parameters (lightweight fallback for applyTimeDecay)
+// ============================================================================
+
+/** Weibull shape parameter per tier. Reuses the same values as decay-engine.ts
+ *  but kept inline to avoid coupling retriever to the full decay engine. */
+const TIER_BETA: Record<string, number> = {
+  core: 0.8,        // Sub-exponential: slow forgetting
+  working: 1.0,     // Standard exponential
+  peripheral: 1.3,  // Super-exponential: fast forgetting
+};
+
+/** Extract tier from entry metadata. Defaults to "working" (β=1.0) for
+ *  backward compatibility — existing entries without tier metadata behave
+ *  exactly as before (standard exponential). */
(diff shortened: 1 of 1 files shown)


AliceLJY:
Superseded by #174. The original head branch for #165 lives in `AliceLJY/memory-lancedb-pro`, which is now archived/read-only, so it can no longer be updated in place. I pushed the same change set plus the fork-PR CI fix to a writable branch in the upstream repo and opened #174 as the replacement PR.

AliceLJY:
Closing in favor of #174, which supersedes this PR on a writable upstream branch.

PR #183, opened 2026-03-13
feat: add temporal supersede semantics for mutable facts

AliceLJY:
## Summary
- add temporal metadata for mutable facts (`fact_key`, `valid_from`, `invalidated_at`, `supersedes`, `superseded_by`)
- add `supersede` handling in smart extraction so updated preferences/entities create a new current fact instead of overwriting history
- filter invalidated memories out of retrieval results and cover the new flow with an end-to-end test

## Scope
This PR intentionally stays narrow:
- metadata-only temporal versioning
- applies to mutable fact categories (`preferences`, `entities`)
- no graph/triple extraction
- no as-of query API yet

## Behavior
When the dedup LLM decides a candidate should supersede an existing mutable fact:
- the old memory is kept
- the old memory is marked with `invalidated_at` and `superseded_by`
- a new memory is created with `valid_from` and `supersedes`
- retrieval returns the new active fact by default

For compatibility, if the dedup LLM returns `contradict` for a mutable fact with no specific context label (or `general`), the implementation normalizes that to `supersede`. This is intentional fallback behavior so truth-replacement still works even before the prompt behavior is perfectly stable.

## Tests
- `npm test`

### DIFF
--- a/src/retriever.ts
+++ b/src/retriever.ts
@@ -354,15 +354,15 @@ export class MemoryRetriever {
       limit,
       this.config.minScore,
       scopeFilter,
+      { excludeInactive: true },
     );
 
     // Filter by category if specified
     const filtered = category
       ? results.filter((r) => r.entry.category === category)
       : results;
-    const active = this.filterActiveResults(filtered);
 
-    const mapped = active.map(
+    const mapped = filtered.map(
       (result, index) =>
         ({
           ...result,
@@ -470,15 +470,15 @@ export class MemoryRetriever {
       limit,
       0.1,
       scopeFilter,
+      { excludeInactive: true },
     );
 
     // Filter by category if specified
     const filtered = category
       ? results.filter((r) => r.entry.category === category)
       : results;
-    const active = this.filterActiveResults(filtered);
 
-    return active.map((result, index) => ({
+    return filtered.map((result, index) => ({
       ...result,
       rank: index + 1,
     }));
@@ -490,15 +490,14 @@ export class MemoryRetriever {
     scopeFilter?: string[],
     category?: string,
   ): Promise<Array<MemorySearchResult & { rank: number }>> {
-    const results = await this.store.bm25Search(query, limit, scopeFilter);
+    const results = await this.store.bm25Search(query, limit, scopeFilter, { excludeInactive: true });
 
     // Filter by category if specified
     const filtered = category
       ? results.filter((r) => r.entry.category === category)
       : results;
-    const active = this.filterActiveResults(filtered);
 
-    return active.map((result, index) => ({
+    return filtered.map((result, index) => ({
       ...result,
       rank: index + 1,
     }));
--- a/src/store.ts
+++ b/src/store.ts
@@ -483,18 +483,22 @@ export class MemoryStore {
     query: string,
     limit = 5,
     scopeFilter?: string[],
+    options?: { excludeInactive?: boolean },
   ): Promise<MemorySearchResult[]> {
     await this.ensureInitialized();
 
     const safeLimit = clampInt(limit, 1, 20);
+    const inactiveFilter = options?.excludeInactive ?? false;
+    // Over-fetch when filtering inactive records to avoid crowding
+    const fetchLimit = inactiveFilter ? Math.min(safeLimit * 20, 200) : safeLimit;
 
     if (!this.ftsIndexCreated) {
-      return this.lexicalFallbackSearch(query, safeLimit, scopeFilter);
+      return this.lexicalFallbackSearch(query, safeLimit, scopeFilter, options);
     }
 
     try {
       // Use FTS query type explicitly
-      let searchQuery = this.table!.search(query, "fts").limit(safeLimit);
+      let searchQuery = this.table!.search(query, "fts").limit(fetchLimit);
 
       // Apply scope filter if provided
       if (scopeFilter && scopeFilter.length > 0) {
@@ -527,8 +531,7 @@ export class MemoryStore {
         const normalizedScore =
           rawScore > 0 ? 1 / (1 + Math.exp(-rawScore / 5)) : 0.5;
 
-        mapped.push({
-          entry: {
+        const entry: MemoryEntry = {
             id: row.id as string,
             text: row.text as string,
             vector: row.vector as number[],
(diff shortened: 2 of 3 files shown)


rwmjhb:
Directionally this is the right phase-1 cut.

What this PR gets right:
- mutable facts now preserve history instead of overwriting the old record
- retrieval returns the current truth by default by filtering inactive versions
- scope is intentionally narrow (`preferences` / `entities`), which is a reasonable first step
- normalizing general-context `contradict` into `supersede` for mutable facts is pragmatic and matches the intended truth-replacement behavior

My main merge-before-landing concern:
- `memory_update` still mutates records in place. For temporal-versioned categories, that bypasses the new `supersedes` / `superseded_by` chain and can destroy the historical record this PR is trying to preserve. I think `preferences` / `entities` should reject raw update or route through the same supersede path.

Follow-up gaps I think should be called out explicitly:
- there is still no `as_of` query path, so we can answer “what is true now” but not “what was true last year”
- temporal state is still metadata-only, so filtering happens after retrieval rather than as a first-class store constraint
- `valid_from` / `invalidated_at` currently use `Date.now()` in the supersede flow, so this is ingestion-time versioning, not fact-effective-time semantics
- `fact_key` is still a heuristic derived from the abstract/topic prefix, not a stable claim identity at the subject-slot-value level
- test coverage proves the current-truth flow, but does not yet cover `as_of`, the `memory_update` bypass, or explicit historical time expressions in the source text

If we fix the `memory_update` bypass, I’m comfortable treating the rest as documented follow-up work for later iterations.

rwmjhb:
Clarifying review scope: this feedback is scoped to `master...pr-183`, i.e. behavior introduced by this PR. Some of the concrete failure points show up in files like `src/store.ts` and `src/
  tools.ts`, even though those files are not directly modified here. I am not calling those out as unrelated legacy issues. The issue is that this PR introduces new temporal-versioning / supersede
  semantics, and those semantics interact with the existing retrieval and dedup paths in ways that now cause regressions. That makes them PR-scope problems.

  Concretely, I see at least these PR-level issues:
  1. We now filter inactive memories only after the store has already truncated to top-N. Once enough superseded history accumulates, historical rows can crowd the active fact out of the candidate
  window. That causes empty recall results or duplicate “current fact” creation because the extractor no longer sees the existing active record.
  2. A `contradict` decision with a missing `context_label` is currently treated as a general contradiction and escalated into `supersede`, which invalidates the existing fact. Since
  `context_label` is LLM-generated output, “missing field” is not a safe proxy for “general contradiction”.
  3. The manual `memory_store` duplicate pre-check still looks at raw nearest neighbors without excluding inactive historical versions, so a superseded fact can incorrectly block writing the new
  current fact.

  So the concern here is not “these old files have issues”; it is “this PR’s new semantics create reproducible regressions when combined with the current retrieval/dedup implementation,” which is
  why I consider them in scope for this review.

AliceLJY:
Independently verified the three regression points from @rwmjhb's review:

**All three confirmed with targeted regression tests:**

1. **top-N truncation + active filter** — Created 8 superseded + 1 active record; `vectorSearch(limit=5)` returned 5 historical, 0 active after filtering. Also affects extractor dedup (would create duplicate active records).
2. **`context_label` missing → false supersede** — `undefined`/`null`/`""` all route to supersede. Missing ≠ general; should default to contradict.
3. **dedup pre-check ghost** — Inactive record matched at score=1.0, blocking legitimate new write.

Agree with fix-then-merge: the supersede guard in `memory_update` (first-comment blocker) is correctly implemented. Issues 1 and 3 are blocking; issue 2 is a safer follow-up.

Regression test script available if helpful.

rwmjhb:
Clarifying scope: this feedback is scoped to `master...pr-183`, i.e. behavior introduced by this PR. Some of the concrete failure points are in existing files like `src/store.ts` / `src/tools.ts`, but I’m not calling those out as unrelated legacy issues. The issue is that this PR introduces temporal-versioning / supersede semantics, and those semantics now interact with the current retrieval and dedup paths in ways that produce regressions. That makes them in-scope for this review.

I currently see four PR-level issues:

1. Blocking: active/inactive filtering happens after the store has already truncated to top-N. Once enough superseded history exists, inactive rows can crowd the current fact out of the candidate window. That means recall can return no current fact, and extractor dedup can miss the active record and create a duplicate current fact. This needs to filter before capping results, or over-fetch until N active rows are available.

2. `contradict` without `context_label` currently falls into the same branch as an explicit `general` contradiction and triggers `handleSupersede()`. Since `context_label` is LLM-generated output, “missing field” is not a safe proxy for “general contradiction”. A partial model response can therefore invalidate the base fact instead of storing a contextual contradiction.

3. The manual `memory_store` duplicate pre-check still uses raw nearest-neighbor results and does not exclude inactive / superseded rows. With temporal versioning enabled, a historical fact can become the top hit and block writing the new current fact with “Similar memory already exists”. This path needs the same active-only semantics as retrieval/extractor.

4. `supersede` still defaults a missing/invalid `match_index` to `topSimilar[0]`. That fallback is destructive now: it can invalidate the wrong memory and propagate the wrong `fact_key` onto the new record. For destructive decisions like `supersede`, I don’t think defaulting to the first candidate is safe; missing `match_index` should degrade to a non-destructive fallback instead.

AliceLJY:
@rwmjhb 四个问题已全部修复，commit `6424587`：

**1. top-N 截断 + active 过滤（blocking）**
- `store.ts` `vectorSearch` 新增 `excludeInactive` 选项，启用时 over-fetch 从 10x 提到 20x，在循环内用 `isMemoryActiveAt` 过滤，凑够 N 条 active 才停
- `smart-extractor.ts` `deduplicate()` 改为传 `{ excludeInactive: true }` 给 store 层，去掉了应用层手动 filter

**2. `context_label` 缺失 → 误触发 supersede**
- 条件从 `!dedupResult.contextLabel || dedupResult.contextLabel === "general"` 改为严格匹配 `dedupResult.contextLabel === "general"`
- 缺失 context_label 现在走普通 contradict，不再升级为 supersede

**3. `memory_store` 去重预检没排除 inactive**
- `tools.ts` 去重预检的 `vectorSearch` 调用加了 `{ excludeInactive: true }`

**4. `match_index` fallback 太危险**
- 对破坏性决策（supersede + contradict），`match_index` 缺失/无效时降级为 create，不再默认取 `topSimilar[0]`
- contradict 也加入了 destructiveDecisions，因为 contradict + general 会走 supersede 路径

另外清理了 smart-extractor.ts 中 `isMemoryActiveAt` 的死导入（过滤已下沉到 store 层）。

全部测试通过，请 review。

PR #194, opened 2026-03-13
fix: declare apache-arrow as a direct runtime dependency

sd1114820:
## Summary

Declare `apache-arrow` as a direct runtime dependency of `memory-lancedb-pro`.

## Why

`@lancedb/lancedb` requires `apache-arrow` at runtime, but currently exposes it via `peerDependencies`.
That works in some regular npm environments, but it can be missed by isolated plugin installation flows.

One reproducible case is OpenClaw's plugin installer, which installs plugin dependencies with:

```bash
npm install --omit=dev --omit=peer --ignore-scripts
```

In that environment, `apache-arrow` is not installed, and the plugin later fails when LanceDB is loaded:

```text
memory-lancedb-pro: failed to load LanceDB
Cannot find module 'apache-arrow'
```

Declaring `apache-arrow` directly fixes both:
- regular npm installation
- isolated OpenClaw plugin installation

## Changes

- add `apache-arrow@18.1.0` to `dependencies`
- add a regression assertion so the package keeps declaring it directly
- refresh `package-lock.json`

## Verification

Ran in a clean container:

```bash
npm ci
node test/plugin-manifest-regression.mjs
node --test test/sync-plugin-version.test.mjs
```

All passed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
   "dependencies": {
     "@lancedb/lancedb": "^0.26.2",
     "@sinclair/typebox": "0.34.48",
+    "apache-arrow": "18.1.0",
     "openai": "^6.21.0"
   },
   "openclaw": {
--- a/test/plugin-manifest-regression.mjs
+++ b/test/plugin-manifest-regression.mjs
@@ -84,6 +84,11 @@ assert.equal(
   pkg.version,
   "openclaw.plugin.json version should stay aligned with package.json",
 );
+assert.equal(
+  pkg.dependencies["apache-arrow"],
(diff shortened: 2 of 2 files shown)


rwmjhb:
感谢这个聚焦修复。

我这边复现了 `--omit=peer` 安装路径下缺少 `apache-arrow` 导致 LanceDB 运行时加载失败的问题；把它声明为直接依赖可以把这个缺口补上。

本地已在隔离目录上应用 PR 补丁并跑完整 `npm test`，全部通过。现在执行合并。

PR #209, opened 2026-03-14
feat: add native support for Azure OpenAI embeddings

Restry:
This PR adds native support for Azure OpenAI endpoints. When the `baseURL` contains `.openai.azure.com`, it automatically injects the required `api-key` header and `api-version` query parameter into the underlying OpenAI client. This allows users to configure Azure OpenAI directly without needing an intermediate proxy like LiteLLM or OneAPI.

### How to configure
Users can simply use the standard `openai-compatible` provider and paste their Azure deployment URL:

```json
"embedding": {
  "baseURL": "https://<YOUR_RESOURCE>.openai.azure.com/openai/deployments/<YOUR_DEPLOYMENT>",
  "apiKey": "<YOUR_AZURE_API_KEY>",
  "model": "text-embedding-3-large",
  "dimensions": 3072
}
```

### DIFF
--- a/src/embedder.ts
+++ b/src/embedder.ts
@@ -302,9 +302,14 @@ export class Embedder {
     this._autoChunk = config.chunking !== false;
 
     // Create a client pool — one OpenAI client per key
+    const isAzure = config.baseURL?.includes('.openai.azure.com');
     this.clients = resolvedKeys.map(key => new OpenAI({
       apiKey: key,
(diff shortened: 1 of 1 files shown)


rwmjhb:
Closing: Azure OpenAI can be configured via proxy (LiteLLM, OneAPI, etc.) without code changes. This implementation also hardcodes api-version and lacks tests. If native Azure support is needed, it should be a proper provider with configurable api-version and test coverage.

PR #210, opened 2026-03-14
ci: skip Claude Code Review on fork PRs

mouseroser:
## Summary
This small follow-up PR fixes a false-negative GitHub Actions failure in `Claude Code Review` for fork-based pull requests.

## Root cause
The current `claude-code-review.yml` workflow runs on `pull_request` and invokes `anthropics/claude-code-action@v1`.

On fork PRs, that job can fail before doing any review work because the action cannot obtain the auth/OIDC context it expects, resulting in errors such as:
- `Unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable`
- `Could not fetch an OIDC token`

## Fix
Skip the `claude-review` job for fork PRs:
- same-repo PRs still run Claude Code Review normally
- fork PRs avoid the misleading red check caused by missing auth context

## Why this is safe
This change does **not** broaden permissions and does **not** switch to `pull_request_target`.
It is a conservative mitigation that removes false failures without introducing new risk.

## Validation
- workflow YAML updated in `.github/workflows/claude-code-review.yml`
- change was already validated in practice on a fork PR where the previous failure mode occurred

## Related context
This was discovered while working on PR #206.

### DIFF
--- a/.github/workflows/auto-assign.yml
+++ b/.github/workflows/auto-assign.yml
@@ -20,7 +20,7 @@ jobs:
           numOfAssignee: 1
 
   assign-prs:
-    if: github.event_name == 'pull_request'
+    if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     ]
   },
   "scripts": {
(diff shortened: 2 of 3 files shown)


rwmjhb:
I pushed one follow-up commit to this branch to apply the same fork-PR guard to `assign-prs`, since that job was still failing with `Resource not accessible by integration` on fork PRs.

Thanks for surfacing the `claude-review` fork/OIDC failure mode first. This makes the fork-PR behavior consistent across both workflows.

mouseroser:
Thanks — I saw the follow-up for `assign-prs` too. That closes the remaining fork-PR permissions gap nicely and makes the workflow behavior consistent across both jobs.

Appreciate you pushing that directly to the branch and getting this merged quickly.

PR #215, opened 2026-03-14
fix: prevent infinite recursion in embedSingle() for CJK text

### LINKED ISSUE #214: [Bug] Infinite recursion in embedSingle() when smartChunk produces single chunk equal to input
## Plugin Version
v1.1.0-beta.8

## OpenClaw Version
v2026.3.8

## Bug Description

When a Discord message contains a large text file attachment (14KB+ Chinese `.md` file), `embedSingle()` enters an infinite recursion loop, producing ~50,000 embedding error calls in 12 minutes and completely blocking the Node.js event loop. All agents on the gateway become unresponsive.

### Root Cause

The recursion occurs because of a **token/character mismatch** in `smartChunk()`:

1. `EMBEDDING_CONTEXT_LIMITS["nomic-embed-text"]` = 8192 (tokens)
2. `smartChunk()` uses this as **character limit**: `maxChunkSize = 8192 * 0.7 = 5740 chars`
3. For CJK text, each character ≈ 2-3 tokens → 5740 Chinese chars ≈ 14,000 tokens → **still exceeds** 8192 token limit
4. The chunk is already ≤ 5740 chars, so `chunkDocument()` cannot split it further
5. `smartChunk()` returns **1 chunk identical to the input** (log: `Split document into 1 chunks`)
6. `embedSingle()` recursively calls itself with the same text → same failure → infinite loop

### Three defects:

1. **Token/Character mismatch**: `smartChunk()` treats token limits as character limits. CJK characters use 2-3x more tokens than ASCII.
2. **No recursion depth guard**: `embedSingle()` has no `depth` parameter to detect and break recursion.
3. **No single-chunk detection**: When `smartChunk()` returns 1 chunk ≈ same size as input, `embedSingle()` should detect this as "chunking failed" and truncate instead of recursing.

## Expected Behavior

- Large text inputs should be chunked correctly (considering token/char ratio for CJK)
- If chunking cannot reduce the text, `embedSingle()` should truncate and embed (not recurse infinitely)
- A single embedding operation should not block the event loop for more than 10 seconds

## Steps to Reproduce

1. Configure plugin with Ollama `nomic-embed-text` (8192 token context)
2. Enable `autoRecall: true`
3. Send a message to the agent containing a 14KB+ Chinese text file attachment
4. Observe gateway.err.log flooding with `Document exceeded context limit` errors
5. All agents become unresponsive; only a gateway restart recovers

## Error Logs

```
2026-03-15T02:36:54+08:00 Document exceeded context limit (400 the input length exceeds the context length), attempting chunking...
2026-03-15T02:36:54+08:00 Split document into 1 chunks for embedding
2026-03-15T02:36:54+08:00 Document exceeded context limit (400 the input length exceeds the context length), attempting chunking...
2026-03-15T02:36:54+08:00 Split document into 1 chunks for embedding
... (49,979 times in 12 minutes, ~70/second)
```

## Embedding Provider

- Provider: Ollama (local)
- Model: `nomic-embed-text` (768 dimensions, 8192 token context)
- URL: `http://localhost:11434/v1`

## OS/Platform

- macOS Darwin 25.1.0 (Apple Silicon)
- Node.js 22.21.0

## Suggested Fix

We've implemented a local fix with 5 changes across 3 files:

1. **`src/embedder.ts`**: Add `depth` parameter to `embedSingle()` (max depth=3, then truncate)
2. **`src/embedder.ts`**: Detect single-chunk output (chunks.length===1 && same size) → truncate instead of recurse
3. **`src/embedder.ts`**: Add 10s timeout wrapper on `embedPassage()`/`embedQuery()`
4. **`src/chunker.ts`**: CJK-aware chunk sizing in `smartChunk()` — divide char limit by 2.5 when CJK ratio > 30%
5. **`index.ts`**: Truncate auto-recall query to 1000 chars before embedding

Will submit a PR with these changes.

RooikeCAO:
## Summary

- Fix infinite recursion in `embedSingle()` when `smartChunk()` returns a single chunk equal to input size
- Add CJK-aware chunk sizing to handle token/character ratio mismatch
- Add recursion depth limit, global timeout, and auto-recall query truncation

Closes #214

## Problem

When a large CJK text file (14KB+ Chinese `.md`) is processed by auto-recall, `embedSingle()` enters an infinite recursion loop producing ~50,000 embedding errors in 12 minutes, blocking the Node.js event loop and making all agents unresponsive.

**Root cause**: `smartChunk()` treats the token limit (8192) as a character limit, but CJK characters use 2-3 tokens each. A chunk of 5740 Chinese characters ≈ 14,000 tokens, still exceeding the 8192 token limit. Since the chunk is already ≤ `maxChunkSize` in characters, the chunker returns 1 chunk identical to the input → `embedSingle()` recurses infinitely.

### `src/embedder.ts`
- **Recursion depth limit**: Add `depth` parameter to `embedSingle()`, max depth 3, then force-truncate
- **Single-chunk detection**: When `smartChunk()` returns 1 chunk ≈ same size as input, truncate instead of recursing
- **Global timeout**: 10s `Promise.race` wrapper on `embedPassage()`/`embedQuery()`
- **`getSafeCharLimit()`**: Returns safe character count per model (2300 for nomic-embed-text considering CJK)

### `src/chunker.ts`
- **CJK-aware sizing**: Detect CJK character ratio; when >30%, divide `maxChunkSize` by 2.5
- For `nomic-embed-text`: 5734 chars → 2293 chars (≈5732 tokens, safely under 8192 limit)

### `index.ts`
- **Auto-recall truncation**: Truncate `event.prompt` to 1000 chars before embedding (only intent needed for memory search, not full attachment text)

## Test plan

- [ ] 14KB Chinese `.md` file embedding completes in <10s (was: infinite loop)
- [ ] No `Document exceeded context limit` log flooding after fix
- [ ] Gateway remains responsive during large text embedding
- [ ] Non-CJK text behavior unchanged (chunk sizes identical to before)
- [ ] Auto-recall with truncated query still returns relevant memories
- [ ] Existing test `embedder-error-hints.test.mjs` passes

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -2002,8 +2002,20 @@ const memoryLanceDBProPlugin = {
           const agentId = resolveHookAgentId(ctx?.agentId, (event as any).sessionKey);
           const accessibleScopes = scopeManager.getAccessibleScopes(agentId);
 
+          // FR-04: Truncate long prompts (e.g. file attachments) before embedding.
+          // Auto-recall only needs the user's intent, not full attachment text.
+          const MAX_RECALL_QUERY_LENGTH = 1_000;
+          let recallQuery = event.prompt;
+          if (recallQuery.length > MAX_RECALL_QUERY_LENGTH) {
+            const originalLength = recallQuery.length;
+            recallQuery = recallQuery.slice(0, MAX_RECALL_QUERY_LENGTH);
+            api.logger.info(
+              `memory-lancedb-pro: auto-recall query truncated from ${originalLength} to ${MAX_RECALL_QUERY_LENGTH} chars`
+            );
+          }
+
           const results = await retrieveWithRetry({
-            query: event.prompt,
+            query: recallQuery,
             limit: 3,
             scopeFilter: accessibleScopes,
             source: "auto-recall",
--- a/src/chunker.ts
+++ b/src/chunker.ts
@@ -163,6 +163,32 @@ function sliceTrimWithIndices(text: string, start: number, end: number): { chunk
 }
 
 // ============================================================================
+// CJK Detection
+// ============================================================================
+
+// CJK Unicode ranges: Unified Ideographs, Extension A, Compatibility,
+// Hangul Syllables, Katakana, Hiragana
+const CJK_RE =
+  /[\u3040-\u309F\u30A0-\u30FF\u3400-\u4DBF\u4E00-\u9FFF\uAC00-\uD7AF\uF900-\uFAFF]/;
+
+/** Ratio of CJK characters to total non-whitespace characters. */
+function getCjkRatio(text: string): number {
+  let cjk = 0;
+  let total = 0;
+  for (const ch of text) {
+    if (/\s/.test(ch)) continue;
+    total++;
+    if (CJK_RE.test(ch)) cjk++;
+  }
+  return total === 0 ? 0 : cjk / total;
+}
+
+// CJK chars are ~2-3 tokens each. When text is predominantly CJK, we divide
+// char limits by this factor to stay within the model's token budget.
+const CJK_CHAR_TOKEN_DIVISOR = 2.5;
+const CJK_RATIO_THRESHOLD = 0.3;
+
+// ============================================================================
 // Chunking Core
 // ============================================================================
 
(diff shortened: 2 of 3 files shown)


rwmjhb:
Thanks for chasing this. I agree the bug is real, and the overall direction makes sense. I was able to reproduce the current failure mode on `master`: with a mocked embedding provider, a ~4000-char Chinese input can trigger `smartChunk()` returning a single chunk equal to the original input, and `embedSingle()` then recurses indefinitely. In my local repro it was able to issue >1000 embedding requests in ~300ms, so this is definitely worth fixing.

That said, I don’t think we should merge this as-is yet. I see two blocking issues:

1. The new timeout wrapper in `src/embedder.ts` is not safe in its current form. `Promise.race()` + `setTimeout()` changes the surfaced result, but it does not cancel the underlying embedding request / recursion, and it does not clear the timer. In practice this means the work can continue in the background after the caller sees a timeout, and even successful calls may still be held open by the pending timer. If we want a timeout here, it should be implemented with actual cancellation (`AbortController` / `AbortSignal.timeout()` or equivalent cleanup).

2. The recursion fallback still is not guaranteed to make progress for all supported models. This PR fixes the reported `nomic-embed-text` + CJK case, but the current `depth` / `safeLimit` / chunk-size heuristics still leave failure paths for small-context models. For example, with a 512-token model like `all-MiniLM-L6-v2`, `smartChunk()` can still produce a single chunk equal to the input, and the forced truncate path is not guaranteed to shrink enough to break the recursion condition. We need a monotonic fallback: each retry must strictly reduce input length, or fail fast instead of recursing again.

Non-blocking note:
- I understand the motivation for truncating `auto-recall` queries in `index.ts`, but `slice(0, 1000)` is fairly blunt. If we keep that change, I’d like to see coverage showing it doesn’t noticeably hurt recall quality for long prompts / attachment-heavy inputs.

Before merge, I’d want regression coverage for:
- `nomic-embed-text` + long CJK input no longer loops
- a small-context model (for example `all-MiniLM-L6-v2`) also terminates correctly
- successful embeddings are not delayed by the timeout path

So my view is: good bug to fix, good direction, but this needs another revision before merge.

rwmjhb:
Closing — the underlying issue is addressed by #226 (ACL/scope filter separation) and #238 (follow-up fixes). Thanks for the contribution!

PR #217, opened 2026-03-15
feat: add extensible OAuth provider selection for memory-pro auth

BlakeFunTeis:
## Summary

- add an extensible OAuth provider registry for `memory-pro auth login`
- add an interactive provider picker in TTY mode when `--provider` is omitted
- fix the OAuth smart-extractor request path so LLM-backed extraction works end-to-end again
- preserve the existing `api-key` LLM path and add regression coverage for both auth modes

### OAuth provider selection

- add provider-aware OAuth primitives with a registry-backed `openai-codex` provider
- add `--provider` to `openclaw memory-pro auth login`
- persist `llm.oauthProvider` into plugin config and OAuth session files
- show the active OAuth provider in `auth status`
- keep non-interactive/default behavior stable by falling back to the default provider when no picker is available

### Interactive CLI flow

- add a TTY provider picker for `openclaw memory-pro auth login` when `--provider` is not passed
- support arrow-key selection + Enter confirmation before opening the browser
- keep explicit `--provider` behavior unchanged for scripts and automation

### OAuth LLM request fix

- fix the OAuth smart-extractor client to send the current streaming Responses-style payload expected by the Codex backend
- add SSE body detection fallback for responses that omit a `text/event-stream` content-type header
- stop inheriting `embedding.baseURL` in OAuth mode unless `llm.baseURL` is explicitly set

### Regression coverage and schema/docs updates

- add schema support for `llm.oauthProvider`
- add regression coverage for interactive provider selection
- add regression coverage for the `api-key` LLM path to ensure OAuth changes do not break the existing OpenAI-compatible flow
- update README / README_CN OAuth documentation to describe the new provider-aware flow without expanding unrelated embedding guidance

## Validation

- `node --test test/cli-oauth-login.test.mjs`
- `node --test test/llm-oauth-client.test.mjs`
- `node --test test/llm-api-key-client.test.mjs`
- `node test/temporal-facts.test.mjs`
- `node test/context-support-e2e.mjs`

## Local runtime verification

With `llm.auth=oauth` and `extractMinMessages=1` in the active OpenClaw runtime config:

- `memory-pro auth login` showed the interactive provider picker in TTY mode
- OAuth login completed successfully with `openai-codex`
- smart extraction reached the OAuth LLM path successfully
- a real recall/extraction run completed end-to-end and produced a persisted smart-extracted memory entry

## Notes

- current provider support is intentionally limited to `openai-codex`, but the CLI/config/runtime flow is now structured for future provider expansion
- the interactive provider picker only appears in TTY mode; scripted/non-interactive usage should continue to pass `--provider` explicitly when needed


BlakeFunTeis:
@codex review

BlakeFunTeis:
@codex review

rwmjhb:
Thanks for pushing this through. I think the PR has clear value overall: it closes a real gap in the smart-extraction flow by making OAuth usable end-to-end, and I appreciate that this is not just a partial runtime patch. The CLI flow, schema, docs, and regression tests are all here, and I ran the new targeted tests locally (`cli-oauth-login`, `llm-oauth-client`, `llm-api-key-client`) and they passed.

That said, I don’t think this is quite ready to merge yet. I see one important product/configuration issue and one related design concern.

1. `memory-pro auth login` / `logout` is not currently a reversible switch.
`login` deletes the existing `llm.apiKey` and overwrites `llm.model`, but `logout` only sets `llm.auth = "api-key"` and keeps the OAuth-era config shape. In practice that means a user who originally had a working dedicated `llm.apiKey` cannot round-trip through OAuth and get back to their previous working config. I reproduced this locally: after `login -> logout`, the config no longer had the original `llm.apiKey`, and the model remained the OAuth-selected model. For users who intentionally separate embedding credentials from LLM credentials, that is a real regression.

2. Preserving the previous `llm.baseURL` in OAuth mode looks risky unless the user explicitly opted into that.
The tests currently codify “preserve custom baseURL” as expected behavior, but the OAuth runtime then feeds that value into the OAuth endpoint builder and appends `/codex/responses`. That is reasonable for a true Codex-compatible proxy, but not for many existing OpenAI-compatible `api-key` endpoints that users may already have in config from the previous mode. So the migration behavior is currently optimistic in a way that may produce a broken post-login config.

My recommendation would be:
- treat OAuth login as either non-destructive, or store enough prior LLM config to restore it on logout
- make the `baseURL` carry-over explicit rather than automatic, or at minimum narrow it to cases we know are compatible
- add one regression test that verifies `login -> logout` restores a working pre-OAuth `llm` configuration, especially when `llm.apiKey` differed from `embedding.apiKey`

So my read is: good direction, real value, but the config migration semantics need one more pass before merge.

BlakeFunTeis:
> Thanks for pushing this through. I think the PR has clear value overall: it closes a real gap in the smart-extraction flow by making OAuth usable end-to-end, and I appreciate that this is not just a partial runtime patch. The CLI flow, schema, docs, and regression tests are all here, and I ran the new targeted tests locally (`cli-oauth-login`, `llm-oauth-client`, `llm-api-key-client`) and they passed.
> 
> That said, I don’t think this is quite ready to merge yet. I see one important product/configuration issue and one related design concern.
> 
> 1. `memory-pro auth login` / `logout` is not currently a reversible switch.
>    `login` deletes the existing `llm.apiKey` and overwrites `llm.model`, but `logout` only sets `llm.auth = "api-key"` and keeps the OAuth-era config shape. In practice that means a user who originally had a working dedicated `llm.apiKey` cannot round-trip through OAuth and get back to their previous working config. I reproduced this locally: after `login -> logout`, the config no longer had the original `llm.apiKey`, and the model remained the OAuth-selected model. For users who intentionally separate embedding credentials from LLM credentials, that is a real regression.
> 2. Preserving the previous `llm.baseURL` in OAuth mode looks risky unless the user explicitly opted into that.
>    The tests currently codify “preserve custom baseURL” as expected behavior, but the OAuth runtime then feeds that value into the OAuth endpoint builder and appends `/codex/responses`. That is reasonable for a true Codex-compatible proxy, but not for many existing OpenAI-compatible `api-key` endpoints that users may already have in config from the previous mode. So the migration behavior is currently optimistic in a way that may produce a broken post-login config.
> 
> My recommendation would be:
> 
> * treat OAuth login as either non-destructive, or store enough prior LLM config to restore it on logout
> * make the `baseURL` carry-over explicit rather than automatic, or at minimum narrow it to cases we know are compatible
> * add one regression test that verifies `login -> logout` restores a working pre-OAuth `llm` configuration, especially when `llm.apiKey` differed from `embedding.apiKey`
> 
> So my read is: good direction, real value, but the config migration semantics need one more pass before merge.



> Thanks for pushing this through. I think the PR has clear value overall: it closes a real gap in the smart-extraction flow by making OAuth usable end-to-end, and I appreciate that this is not just a partial runtime patch. The CLI flow, schema, docs, and regression tests are all here, and I ran the new targeted tests locally (`cli-oauth-login`, `llm-oauth-client`, `llm-api-key-client`) and they passed.
> 
> That said, I don’t think this is quite ready to merge yet. I see one important product/configuration issue and one related design concern.
> 
> 1. `memory-pro auth login` / `logout` is not currently a reversible switch.
>    `login` deletes the existing `llm.apiKey` and overwrites `llm.model`, but `logout` only sets `llm.auth = "api-key"` and keeps the OAuth-era config shape. In practice that means a user who originally had a working dedicated `llm.apiKey` cannot round-trip through OAuth and get back to their previous working config. I reproduced this locally: after `login -> logout`, the config no longer had the original `llm.apiKey`, and the model remained the OAuth-selected model. For users who intentionally separate embedding credentials from LLM credentials, that is a real regression.
> 2. Preserving the previous `llm.baseURL` in OAuth mode looks risky unless the user explicitly opted into that.
>    The tests currently codify “preserve custom baseURL” as expected behavior, but the OAuth runtime then feeds that value into the OAuth endpoint builder and appends `/codex/responses`. That is reasonable for a true Codex-compatible proxy, but not for many existing OpenAI-compatible `api-key` endpoints that users may already have in config from the previous mode. So the migration behavior is currently optimistic in a way that may produce a broken post-login config.
> 
> My recommendation would be:
> 
> * treat OAuth login as either non-destructive, or store enough prior LLM config to restore it on logout
> * make the `baseURL` carry-over explicit rather than automatic, or at minimum narrow it to cases we know are compatible
> * add one regression test that verifies `login -> logout` restores a working pre-OAuth `llm` configuration, especially when `llm.apiKey` differed from `embedding.apiKey`
> 
> So my read is: good direction, real value, but the config migration semantics need one more pass before merge.

Thanks for the review. I pushed a follow-up that addresses the config migration concerns, and I also merged the latest `master` into this branch. I want to call out the merge details explicitly so it is easy to verify that nothing regressed on the upstream side.

On the OAuth migration issue itself, `memory-pro auth login` is now reversible. Login snapshots the previous api-key-based `llm` config next to the OAuth file, and `memory-pro auth logout` restores that snapshot when it exists. The flow no longer drops a dedicated `llm.apiKey` on the floor, and it no longer leaves users stuck on the OAuth-selected model after logout. I also changed the fallback behavior to only preserve values that actually existed in the original config. If there is no restorable pre-OAuth `llm` config, logout now removes the plugin `llm` block instead of synthesizing an `api-key` shape from OAuth-era values.

I also removed the implicit carry-over of `llm.baseURL` from api-key mode into OAuth mode. In OAuth mode, `llm.baseURL` is only used if it is explicitly configured for an OAuth-compatible backend. This avoids the optimistic migration behavior you flagged for generic OpenAI-compatible endpoints.

On top of that, I changed the default OAuth file location from the current working directory to the plugin-scoped global path under OpenClaw home. The default is now `~/.openclaw/.memory-lancedb-pro/oauth.json` rather than `<cwd>/.memory-lancedb-pro/oauth.json`. This keeps both the OAuth token file and the restore sidecar out of repo working directories by default.

For the merge itself, I resolved conflicts by taking latest `master` as the base and then reapplying only the OAuth-related behavior needed for this PR. The resulting branch retains the recent upstream changes, including:

- `llm.timeoutMs` support in config/schema/docs
- the newer LLM diagnostics / malformed-JSON repair behavior
- `getLastError()` support on the LLM client
- workspace-boundary / USER.md-exclusive handling
- new upstream tests such as `memory-upgrader-diagnostics`
- current manifest/package updates from `master`

The merge-specific OAuth additions on top of that are:

- `llm.auth`
- `llm.oauthProvider`
- `llm.oauthPath`
- OAuth client handling in `src/llm-client.ts`
- CLI auth commands and config migration / restore behavior
- OAuth docs and regression coverage

I re-ran the full test suite after the merge and after the follow-up fixes. `npm test` passes. I also manually verified runtime behavior under OAuth: with `provider=openai-codex` and `model=gpt-5.4`, `memory-pro` successfully completed the full path of recall -> auto-capture -> smart extraction -> persistence, and I saw a successful create event in logs (`smart-extractor: created ...`).

I additionally validated the specific config edge cases from review:

- `login -> logout` restores a dedicated pre-OAuth `llm.apiKey` config
- the round-trip works when `llm.apiKey` differs from `embedding.apiKey`
- logout does not preserve OAuth-generated `llm.model` values when there was no original api-key config to restore
- default OAuth path resolution now uses the plugin-scoped path under `OPENCLAW_HOME`

If you have time for another pass, I’d especially appreciate a look at the merge-sensitive files:

- `index.ts`
- `src/llm-client.ts`
- `openclaw.plugin.json`
- `README.md`
- `README_CN.md`

Those are the places where upstream timeout/diagnostics changes and this branch’s OAuth behavior now coexist, so they are the most relevant spots to sanity-check for merge mistakes.

Addressed in:

- a8f8df3 — restore pre-OAuth LLM config on logout
- 0c134c1 — tighten oauth path and logout restore
- 4d9622a — merge upstream master

PR #216, opened 2026-03-14
fix: improve voyage embedding provider compatibility

SongJiangzhou:
## Problem

Voyage AI (and Voyage-style providers such as proxies backed by Voyage models) expose an OpenAI-compatible embeddings endpoint but **reject** standard OpenAI request fields and use **different field names** for task hints and output dimensions:

| Field | OpenAI / Generic | Voyage AI |
|-------|-----------------|-----------|
| Format hint | `encoding_format: "float"` | ❌ rejected (`400 Bad Request`) |
| Output dimensions | `dimensions` | `output_dimension` (Matryoshka models only) |
| Task hint | `task` | `input_type` (`"query"` / `"document"`) |

Without handling these differences, all embedding calls to Voyage endpoints failed with `400 Bad Request: Unknown request body key: encoding_format`, and task/dimension configuration was silently dropped.

## Root Cause

`buildPayload` unconditionally appended `encoding_format`, `dimensions`, and `task` to every request, with no mechanism to suppress or rename fields based on the provider.

Additionally, Voyage's `input_type` accepts only `"query"` and `"document"`, but the rest of the codebase uses the broader OpenAI-style values `"retrieval.query"` and `"retrieval.passage"`.

## Solution

Introduce a **provider profile + generic capabilities** layer detected once at `Embedder` construction time, replacing per-provider conditionals in `buildPayload`.

### New types

```ts
type EmbeddingProviderProfile =
  | "openai"
  | "jina"
  | "voyage-compatible"
  | "generic-openai-compatible";

interface EmbeddingCapabilities {
  /** Whether to send encoding_format: "float" */
  encoding_format: boolean;
  /** Whether to send normalized (Jina-style) */
  normalized: boolean;
  /**
   * Field name for the task hint, or null if unsupported.
   * e.g. "task" for Jina, "input_type" for Voyage, null for OpenAI/generic.
   */
  taskField: string | null;
  /** Optional value translation map (e.g. Voyage: "retrieval.query" → "query") */
  taskValueMap?: Record<string, string>;
  /**
   * Field name for the requested output dimension, or null if unsupported.
   * e.g. "dimensions" for OpenAI, "output_dimension" for Voyage.
   */
  dimensionsField: string | null;
}
```

### Detection logic (`detectEmbeddingProviderProfile`)

| Signal | Profile |
|--------|---------|
| baseURL contains `api.openai.com` | `openai` |
| baseURL contains `api.jina.ai` OR model starts with `jina-` | `jina` |
[...]

PR #219, opened 2026-03-15
fix: use input_type instead of task for NVIDIA NIM embeddings

vicjayjay:
## Summary

- NVIDIA NIM API now strictly validates request bodies and rejects unknown fields
- The `task` parameter causes a 400 error on NVIDIA endpoints
- Fix routes by provider: NVIDIA endpoints get `input_type`, others get `task`

## Root Cause

`buildPayload()` in `embedder.ts` was sending both `task` and `input_type` to all providers. NVIDIA rejects `task` as unknown.

## Test plan

- Verified with nvidia/llama-3.2-nemoretriever-1b-vlm-embed-v1
- Non-NVIDIA providers still receive task parameter as before

### DIFF
--- a/src/embedder.ts
+++ b/src/embedder.ts
@@ -453,7 +453,14 @@ export class Embedder {
       encoding_format: "float",
     };
 
-    if (task) payload.task = task;
+    if (task) {
+      // NVIDIA NIM uses input_type; other providers (Jina etc.) use task
+      if (this._baseURL && /nvidia\.com/i.test(this._baseURL)) {
+        payload.input_type = task;
+      } else {
+        payload.task = task;
+      }
+    }
     if (this._normalized !== undefined) payload.normalized = this._normalized;
 
     // Some OpenAI-compatible providers support requesting a specific vector size.


rwmjhb:
This is pointing at a real compatibility problem, but I don’t think the current patch is complete enough to merge yet.

Two gaps need to be addressed:

1. `input_type` value mapping is still wrong for existing config
   - This repo currently documents/configures `taskQuery` / `taskPassage` as `retrieval.query` / `retrieval.passage`
   - NVIDIA NIM expects `input_type` values like `query` / `passage`, not `retrieval.query` / `retrieval.passage`
   - So the current change appears to rename the field without translating the value, which likely still breaks existing NVIDIA configs

2. Provider detection is too narrow
   - The current branch only switches behavior when `baseURL` matches `nvidia.com`
   - That misses self-hosted NIM, private gateways, reverse proxies, and other non-`nvidia.com` deployments
   - If the goal is “NVIDIA NIM compatibility”, URL substring matching alone is not robust enough

Please also add automated coverage for this.
At minimum, I want tests that prove:
- NVIDIA-compatible requests send `input_type`
- non-NVIDIA providers still send `task`
- `retrieval.query` / `retrieval.passage` are mapped correctly for NVIDIA if that is the intended compatibility path

Once those pieces are in place, I’m happy to take another look.

rwmjhb:
感谢提交！NVIDIA NIM 拒绝 `task` 字段的问题是真实存在的。

不过这个 PR 基于的代码已经过时了 — #216 合并后，`buildPayload` 中的 `if (task) payload.task = task;` 已被替换为 provider profile 体系：

```ts
if (this._capabilities.taskField && task) {
  const value = cap.taskValueMap?.[task] ?? task;
  payload[cap.taskField] = value;
}
```

现在正确的做法是在 `detectEmbeddingProviderProfile` 中加一个 `"nvidia"` profile，然后在 `getEmbeddingCapabilities` 中设置 `taskField: "input_type"`，与 Voyage 的处理方式一致。这样不需要在 `buildPayload` 中硬编码 URL 判断。

如果你有兴趣，欢迎基于最新 master 重新提交。

PR #229, opened 2026-03-15
fix(auto-capture): strip OpenClaw wrapper metadata before extraction

### LINKED ISSUE #211: [Bug] autoCapture includes Feishu metadata in stored memories
## Problem Description

When using <strong>autoCapture</strong> with OpenClaw Feishu channel, the untrusted metadata blocks (Conversation info / Sender) are being captured and stored in the memory database, causing memory pollution.

## Environment

- Plugin version: 1.1.0-beta.8
- OpenClaw channel: Feishu (飞书)
- Configuration: autoCapture: true, smartExtraction: true

## Observed Behavior

Stored memory entries include content like:

## Expected Behavior

The metadata blocks should be stripped before Smart Extraction and memory storage, similar to how they are handled in normalizeQuery() for retrieval (as implemented in PR #47).

## Root Cause Analysis

The current implementation strips metadata in:
- ✓ normalizeQuery() for adaptive retrieval (PR #47)
- ✓ stripAutoCaptureInjectedPrefix() strips [UNTRUSTED DATA] blocks
- ✗ But NOT the "Conversation info (untrusted metadata)" blocks in auto-capture path

## Suggested Fix

Apply the same metadata stripping logic used in normalizeQuery() to the auto-capture pipeline, specifically in:
- stripAutoCaptureInjectedPrefix() function
- Or in the smart extraction preprocessing step

## Temporary Workaround

Disable autoCapture and use manual memory_store instead:

## References

- Similar fix for retrieval: PR #47
- Related issue with metadata handling: PR #50

---

Thank you for considering this fix. Happy to provide more details or test any patches.

chenhaonan-eth:
## Summary

This PR fixes auto-capture memory pollution caused by OpenClaw wrapper metadata being passed through to smart extraction as if it were user content.

It closes the gap reported in #211 and broadens the cleanup beyond Feishu so the same class of contamination is removed for other channel payloads that carry OpenClaw transport/context wrappers.

## Problem

In some OpenClaw sessions, the user message payload can contain transport/context wrappers such as:

- `<relevant-memories>...</relevant-memories>`
- `[UNTRUSTED DATA ...] ... [END UNTRUSTED DATA]`
- `Conversation info (untrusted metadata):` + fenced JSON
- `Sender (untrusted metadata):` + fenced JSON
- runtime event lines like `System: [timestamp] Exec completed ...`
- session reset bootstrap text injected after `/new` or `/reset`

When these wrappers were not fully removed before auto-capture/smart extraction, the plugin could store garbage memories containing metadata like `message_id`, `username`, `Conversation info`, etc.

## Root cause

`normalizeQuery()` already had cleanup behavior for retrieval, but the auto-capture path still had incomplete sanitization.

`stripAutoCaptureInjectedPrefix()` handled some wrappers, but it did **not reliably strip the OpenClaw inbound metadata blocks** and related system-event noise before smart extraction.

As a result, upstream wrapper text from OpenClaw could leak into:

1. pending ingress text
2. agent_end message normalization
3. smart extraction prompts
4. stored memory entries

### 1) Extracted cleanup into a dedicated helper

Added `src/auto-capture-cleanup.ts` to centralize auto-capture text normalization.

### 2) Strip wrapper metadata more aggressively in auto-capture

The new cleanup removes:

- `<relevant-memories>` blocks
- `[UNTRUSTED DATA ...]` blocks
- session reset bootstrap prefix
- inbound metadata sentinel blocks with fenced JSON payloads
- `System: [..] Exec completed|failed|started ...` event lines
- leading mention/addressing prefixes

It also re-runs cleanup passes and collapses extra blank lines so mixed wrapper payloads are handled robustly.

### 3) Reused the same normalization in all auto-capture entry points

Updated `index.ts` so both:

- ingress capture normalization
- `agent_end` message/block normalization

flow through the same helper and still honor `shouldSkipReflectionMessage()`.

## Regression coverage

Added a scenario to `test/smart-extractor-branches.mjs` that simulates a polluted payload shape with wrapper blocks, fenced metadata JSON, runtime event noise, and a real user sentence.

The test asserts that:

- smart extraction prompt still contains the real user statement
- extraction prompt no longer contains wrapper metadata
- stored entries contain the real memory content
- stored entries do **not** contain `Conversation info`, `Sender`, `message_id`, or `username`

## Validation

Ran:

```bash
node test/smart-extractor-branches.mjs
npm test
```

Both passed locally.

## User impact

After this patch, auto-capture should stop writing transport-level wrapper noise into memory when OpenClaw channels inject context blocks around real user messages.

That means cleaner smart extraction, cleaner stored memories, and fewer false facts caused by metadata pollution.

Fixes #211

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -38,6 +38,7 @@ import { createReflectionEventId } from "./src/reflection-event-store.js";
 import { buildReflectionMappedMetadata } from "./src/reflection-mapped-metadata.js";
 import { createMemoryCLI } from "./cli.js";
 import { isNoise } from "./src/noise-filter.js";
+import { normalizeAutoCaptureText } from "./src/auto-capture-cleanup.js";
 
 // Import smart extraction & lifecycle components
 import { SmartExtractor } from "./src/smart-extractor.js";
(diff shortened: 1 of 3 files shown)


rwmjhb:
I think this fix is worth taking.

The problem is real, the scope is appropriate, and the cleanup is being applied in the right place: before auto-capture/smart extraction turns wrapper text into candidate memory content. I also like that this is not just a one-off Feishu patch. Pulling the normalization into a dedicated helper and reusing it across the auto-capture entry points is the right direction.

I ran the targeted regression (`node test/smart-extractor-branches.mjs`) and also ran the full `npm test` suite on the PR branch in an isolated worktree. Both passed for me.

At this point I’m not seeing any additional blocking issue in the code itself. The only thing preventing merge right now is that GitHub shows the PR as `CONFLICTING` against current `master`.

So my recommendation is:
- rebase / resolve conflicts against current `master`
- rerun `node test/smart-extractor-branches.mjs` and `npm test`
- if still green, merge it

I would support merge after the branch is rebased.

AliceLJY:
Rebased onto master, resolved conflicts, fixed test async issue, and merged directly. Thank you @chenhaonan-eth for this excellent contribution! 🎉

PR #242, opened 2026-03-17
fix: add /opt/homebrew extensionAPI fallback

stanleeyY:
## Summary
- add `/opt/homebrew/lib/node_modules/openclaw/dist/extensionAPI.js` to the dynamic import fallback list
- preserve compatibility with Apple Silicon macOS installs where OpenClaw is installed via Homebrew

## Why
The plugin already tries common global install locations such as `/usr/lib/...` and `/usr/local/...`, but Homebrew on Apple Silicon uses `/opt/homebrew/...`.

Without this fallback, the plugin can fail to resolve `openclaw/dist/extensionAPI.js` in some Homebrew-based deployments even though the file exists on disk.

## Validation
- `npm test` ✅
- verified local Homebrew install path exists on macOS Apple Silicon ✅


rwmjhb (approved):
LGTM, 补全了 Apple Silicon Homebrew 安装路径。

PR #243, opened 2026-03-17
fix: prevent same-brand different-item preferences from being deduped

### LINKED ISSUE #228: [Bug] New item-level preference can be dropped when same-brand preferences already exist
## Summary

When long-term memory already contains multiple food-item preferences under the same brand/topic, adding a **new item-level preference** can be silently dropped by `memory-lancedb-pro` instead of being stored as a new atomic preference.

This makes the plugin lose real user preference facts.

A concrete example:

Existing memories already contain preferences like:
- 喜欢吃麦当劳的板烧鸡腿堡
- 喜欢吃麦当劳的麦旋风

Then the user says:
- 我喜欢吃麦当劳的麦辣鸡翅

Expected: a new preference for `麦辣鸡翅` should be stored.

Observed: the new preference may be skipped / merged away / treated as duplicate, so it does not reliably become an independent memory.

---

## Why this is a bug

From a user perspective, these are **different stable preferences**:
- brand-level preference: 喜欢麦当劳
- item-level preference: 喜欢麦辣鸡翅
- another item-level preference: 喜欢板烧鸡腿堡

These are not interchangeable facts.

If the plugin merges them too aggressively, it loses granularity and corrupts the memory model.

---

## Observed pattern

This does **not** look like a simple exact-duplicate issue.

The failure pattern appears when:
1. there are already existing item-level food preferences under the same brand/topic
2. autoCapture / smartExtraction runs
3. the new item-level preference is treated as part of an existing same-topic preference cluster
4. the new memory does not survive as an independent stored memory

This seems especially likely when Smart Extraction is enabled and the plugin decides between:
- `CREATE`
- `MERGE`
- `SKIP`

---

## Likely root cause

The dedup / smart-extraction logic appears to be using a topic granularity that is too coarse for food preferences.

It likely treats:
- “喜欢麦当劳的板烧鸡腿堡”
- “喜欢麦当劳的麦旋风”
- “喜欢麦当劳的麦辣鸡翅”

as belonging to one same semantic cluster such as “user likes McDonald's food”, then merges or skips the new item.

But for preference memory, **item-level preference should usually remain atomic**.

---

## Expected behavior

For statements like:
- 我喜欢吃 X 的 Y

if `Y` is a new concrete item and does not already exist in memory, it should default to **CREATE**, not be silently merged away.

More specifically:

### Should store separately
- 喜欢吃麦当劳的板烧鸡腿堡
- 喜欢吃麦当劳的麦旋风
- 喜欢吃麦当劳的麦辣鸡翅

### Optional additional aggregate memory is fine
A summary memory such as:
- 喜欢吃麦当劳的麦旋风、板烧鸡腿堡、麦辣鸡翅

is acceptable **only if it does not replace or suppress the atomic item-level memories**.

---

## Actual behavior

The new item-level preference can be lost.

So the memory system ends up remembering an old item list or a coarse topic, but not the newly stated preference itself.

---

## Why this matters

This is not a cosmetic issue.

It affects one of the most important memory categories: **user preferences**.

If the plugin silently compresses distinct preferences into broader themes, then over time:
- user preference recall becomes incomplete
- the assistant answers with outdated or partial preference sets
- trust in memory persistence drops

---

### Option A — keep atomic item preferences
For `preference` memories, if the new memory contains a new concrete object/item, prefer `CREATE` over `MERGE`.

### Option B — category-specific dedup policy
Use stricter anti-merge rules for:
- food preferences
- taste preferences
- brand + item preferences
- similar “X likes Y-item” statements

### Option C — allow aggregate summaries without suppressing atomics
If the plugin wants a summary memory, it should create/update the summary **in addition to** preserving the atomic item memories.

### Option D — expose merge decision diagnostics
It would help debugging if logs explicitly state when a candidate preference was:
- `CREATE`
- `MERGE`
- `SKIP`

and why.

---

## Additional note

This issue seems related in spirit to other duplicate / merge sensitivity discussions, but this report is specifically about:

> **new item-level preferences under an existing same-brand/same-topic preference cluster being dropped incorrectly**

That is more specific than generic dedup complaints.

AliceLJY:
## Summary

Fixes #228 — when long-term memory already contains item-level preferences under the same brand (e.g. "喜欢麦当劳的板烧鸡腿堡"), adding a **new** item-level preference (e.g. "喜欢麦当劳的麦辣鸡翅") can be silently dropped by the dedup pipeline.

**Root cause**: Vector similarity between same-brand item preferences is high, so they pass the similarity threshold. The LLM dedup stage then aggressively merges or skips them because they share the same topic.

## Changes

- **New file: `src/preference-slots.ts`** — Deterministic brand-item preference parser
  - `parseBrandItemPreference()` — extracts brand + item(s) from text like "喜欢吃麦当劳的麦辣鸡翅" (supports Chinese and English patterns)
  - `inferAtomicBrandItemPreferenceSlot()` — returns a typed slot for single-item preferences only (aggregate preferences like "喜欢麦旋风、鸡翅和鸡腿堡" are left to LLM)

- **Modified: `src/smart-extractor.ts`** — Added "Stage 1.5" preference-slot guard in `deduplicate()`
  - After vector pre-filter finds similar memories, but **before** LLM decision
  - If the candidate is an atomic brand-item preference, and **all** similar matches are the same brand but different items → force `create`
  - If any similar match has the same brand **and** same item → proceed to LLM dedup as normal (genuine duplicate)
  - Non-preference categories and aggregate preferences are unaffected

## Example

| Existing memory | New input | Before | After |
|---|---|---|---|
| 喜欢麦当劳的板烧鸡腿堡 | 喜欢麦当劳的麦辣鸡翅 | ❌ merged/skipped | ✅ created |
| 喜欢麦当劳的麦旋风 | 喜欢麦当劳的麦旋风 | ✅ deduped | ✅ deduped (unchanged) |
| 喜欢星巴克 | 喜欢瑞幸 | ✅ created | ✅ created (unchanged) |

## Origin

This approach is ported from [RecallNest](https://github.com/AliceLJY/recallnest) (a downstream fork focused on local-first memory), where it has been running in production since 2026-03-16.

## Test plan

- [ ] Verify atomic brand-item preferences with different items are stored independently
- [ ] Verify genuine duplicates (same brand + same item) still get deduped
- [ ] Verify non-preference categories are unaffected
- [ ] Verify aggregate preferences (multiple items in one sentence) are unaffected

### DIFF
--- a/src/preference-slots.ts
+++ b/src/preference-slots.ts
@@ -2,7 +2,7 @@ const ROLE_PREFIX_RE = /^\[(用户|助手)\]\s*/gm;
 const PREFERENCE_SPLIT_RE = /(?:、|,|，|\/|以及|及|与|和| and | & )/iu;
 const PREFERENCE_CLAUSE_STOP_RE = /(?:因为|所以|但是|不过|if |when |because |but )/iu;
 const BRAND_ITEM_PREFERENCE_PATTERNS = [
-  /(?:^|[\s，,。；;！!？?])(?:我|用户)?(?:很|更|还)?(?:喜欢|爱吃|偏爱|常吃|想吃|喜欢喝|喜欢用|喜欢买)(?:吃|喝|用|买)?(?<brand>[\p{Script=Han}A-Za-z0-9&·'\-]{1,24})的(?<items>[\p{Script=Han}A-Za-z0-9&·'\-\s、,，和及与/]{1,80})/u,
+  /(?:^|[\s，,。；;！!？?])(?:我|用户)?(?:很|更|还)?(?:喜欢|爱吃|偏爱|常吃|想吃)(?:吃|喝|用|买)?(?<brand>[\p{Script=Han}A-Za-z0-9&·'\-]{1,24})的(?<items>[\p{Script=Han}A-Za-z0-9&·'\-\s、,，和及与/]{1,80})/u,
   /\b(?:i|user)?\s*(?:really\s+|still\s+|also\s+)?(?:like|love|prefer|enjoy)\s+(?<items>[a-z0-9'&\-\s]{1,80})\s+from\s+(?<brand>[a-z0-9'&\-\s]{1,40})/iu,
 ] as const;
 
@@ -28,6 +28,7 @@ function normalizePreferenceText(value: string): string {
 export function normalizePreferenceToken(value: string): string {
   return normalizePreferenceText(value)
     .replace(/^[“"'`‘’]+|[”"'`‘’。！!？?，,；;:：]+$/gu, "")
+    .replace(/\b(?:the|a|an)\s+/giu, "")
     .replace(/\s+/g, "")
     .toLowerCase();
 }
--- a/test/preference-slots.test.mjs
+++ b/test/preference-slots.test.mjs
@@ -96,23 +96,185 @@ test("normalizePreferenceToken: strips punctuation and lowercases", () => {
 });
 
 // ---------------------------------------------------------------------------
-// category guard integration (verify "preferences" plural is used)
+// normalizePreferenceToken: English article stripping
 // ---------------------------------------------------------------------------
 
-test("smart-extractor uses 'preferences' (plural) for category check", async () => {
-  const { readFileSync } = await import("node:fs");
-  const src = readFileSync(
-    new URL("../src/smart-extractor.ts", import.meta.url),
-    "utf8",
-  );
-  // Must contain the plural form in the preference-slot guard
-  assert.ok(
-    src.includes('candidate.category === "preferences"'),
-    'Guard should check for "preferences" (plural), not "preference" (singular)',
-  );
-  // Must NOT contain the singular typo
-  assert.ok(
-    !src.includes('candidate.category === "preference"'),
-    'Should not have the singular "preference" typo',
-  );
+test("normalizePreferenceToken: strips English articles (the/a/an)", () => {
+  assert.equal(normalizePreferenceToken("the Big Mac"), "bigmac");
+  assert.equal(normalizePreferenceToken("Big Mac"), "bigmac");
+  assert.equal(normalizePreferenceToken("a Whopper"), "whopper");
+  assert.equal(normalizePreferenceToken("an Egg McMuffin"), "eggmcmuffin");
+});
+
+// ---------------------------------------------------------------------------
+// Dedup guard integration: SmartExtractor preference-slot guard behavior
+// ---------------------------------------------------------------------------
+
+const { SmartExtractor } = jiti("../src/smart-extractor.ts");
+
+function makeGuardExtractor({ vectorSearchResults, onDedupCalled }) {
+  const stored = [];
+  const store = {
+    async vectorSearch() {
+      return vectorSearchResults;
+    },
+    async store(entry) {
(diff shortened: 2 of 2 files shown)


rwmjhb:
## Review

思路很好——用确定性的品牌-单品解析在 vector pre-filter 和 LLM 之间加一层守卫，避免同品牌不同单品被误合并。守卫逻辑的保守设计（`.every()` + fallback to LLM）也合理。

### 需要修改

**1. 缺少 dedup 守卫的集成测试**

自己 self-review 里明确说要加集成测试，但目前只有 `preference-slots.ts` 的单元测试。Stage 1.5 守卫在 `deduplicate()` 里的完整路径没有测试覆盖——mock store + embedder，构造 `category: "preferences"` 的 candidate，验证同品牌不同单品走 force create、同品牌同单品走 LLM。

**2. 源码字符串匹配测试应替换为行为测试**

最后一个测试用 `readFileSync` 读源码检查 `includes('candidate.category === "preferences"')`。重构（改变量名、提取方法）就会误报。用行为测试替代：mock 一个 `category: "preferences"` 的 candidate 走 dedup 路径，验证守卫触发。这样同时解决第 1 点。

**3. Rebase**

`package.json` 加 `json5` 和 `package-lock.json` 的 `apache-arrow` peer 改动已经在 master 上了，rebase 后这些 diff 会消失。

### 建议

**4. 英文模式冠词不一致**

`normalizePreferenceToken` 去空格后，`"the Big Mac"` → `"thebigmac"` vs `"Big Mac"` → `"bigmac"`，同一单品产生不同 token。主要影响英文场景，中文不受影响。可以在 normalize 时 strip 常见冠词（the/a/an）。

**5. 中文正则有冗余分支**

```
(?:喜欢|爱吃|偏爱|常吃|想吃|喜欢喝|喜欢用|喜欢买)(?:吃|喝|用|买)?
```

`"喜欢喝"` 永远匹配不到——`"喜欢"` 在前面先命中，后面 `(?:喝)?` 消费了 `"喝"`。后三个分支可以去掉。

AliceLJY:
Thanks for the thorough review! All 5 items addressed in c90bd12:

**Must fix:**
1. ✅ Added 3 integration tests that mock SmartExtractor dedup pipeline: same-brand-different-item (guard triggers, LLM skipped), same-brand-same-item (falls through to LLM), non-preference category (guard skipped).
2. ✅ Replaced the `readFileSync` source-string test with the behavior tests above.
3. ✅ Rebased onto master. `apache-arrow` diff gone; `json5` is a genuine new dependency for this PR.

**Suggestions:**
4. ✅ `normalizePreferenceToken` now strips `the/a/an` before collapsing whitespace, so `"the Big Mac"` and `"Big Mac"` both normalize to `"bigmac"`.
5. ✅ Removed redundant `喜欢喝|喜欢用|喜欢买` branches — `喜欢` always matches first.

All 14 preference-slots tests pass. The 4 OAuth test failures in CI are pre-existing on master.

rwmjhb:
The feature looks good and tests are solid. One issue before merging:

`package.json` removes `"json5": "^2.2.3"` from dependencies — this was added in #256 (merged Mar 22) as a runtime dependency. Looks like it got dropped during your rebase conflict resolution. Please add it back and push. Thanks!

PR #268, opened 2026-03-19
fix: add NVIDIA NIM provider profile for input_type embedding field

vicjayjay:
## Summary

Supersedes #219. Implements the fix properly using the provider profile architecture from #216.

- Adds "nvidia" to `EmbeddingProviderProfile` type
- Detects NVIDIA NIM via `*.nvidia.com` base URLs, `nvidia/*` model prefixes, and `nv-embed*` model names
- Sets `taskField: "input_type"` with value mapping (`retrieval.query` → `query`, `retrieval.passage` → `passage`)
- Supports `encoding_format: "float"` and forwards `dimensions` for NVIDIA dynamic models

## Changes

- `src/embedder.ts`: Added the NVIDIA provider profile, kept `dimensions` forwarding for dynamic models, and simplified the NVIDIA baseURL detection regex to `/\.nvidia\.com/i`
- `test/nvidia-nim-provider-profile.test.mjs`: 6 automated tests covering all maintainer requirements:
  - NVIDIA requests send `input_type` (not `task`)
  - `retrieval.query`/`retrieval.passage` are mapped correctly
  - NVIDIA is detected from an `.nvidia.com` baseURL
  - Non-NVIDIA providers (Jina) still send `task`
  - Generic providers send neither `task` nor `input_type`

## Test plan

- [x] All 6 NVIDIA profile tests pass (`node --test test/nvidia-nim-provider-profile.test.mjs`)
- [x] Existing embedder tests pass (`node test/embedder-error-hints.test.mjs`)
- [ ] CI pipeline

### DIFF
--- a/src/embedder.ts
+++ b/src/embedder.ts
@@ -237,15 +237,19 @@ function detectEmbeddingProviderProfile(
   model: string,
 ): EmbeddingProviderProfile {
   const base = baseURL || "";
+  let host = "";
+  try { host = new URL(base).hostname.toLowerCase(); } catch { /* invalid URL — skip host checks */ }
 
   // Host-based detection runs first — endpoint owner semantics take precedence
   // over model-name heuristics to avoid misclassifying e.g. a jina-xxx model
   // served from .nvidia.com as Jina instead of NVIDIA.
-  if (/api\.openai\.com/i.test(base)) return "openai";
-  if (/\.openai\.azure\.com/i.test(base)) return "azure-openai";
-  if (/api\.jina\.ai/i.test(base)) return "jina";
-  if (/api\.voyageai\.com/i.test(base)) return "voyage-compatible";
-  if (/\.nvidia\.com/i.test(base)) return "nvidia";
+  // Match on parsed hostname to avoid false positives from proxy URLs that
+  // contain provider domains in their path or query string.
+  if (host.endsWith("api.openai.com")) return "openai";
+  if (host.endsWith(".openai.azure.com")) return "azure-openai";
+  if (host.endsWith("api.jina.ai")) return "jina";
+  if (host.endsWith("api.voyageai.com")) return "voyage-compatible";
+  if (host.endsWith(".nvidia.com") || host === "nvidia.com") return "nvidia";
 
   // Model-prefix fallback — only when baseURL didn't match a known host
   if (/^jina-/i.test(model)) return "jina";


vicjayjay:
@rwmjhb synced this up.

I updated the PR description so it now reflects that NVIDIA dynamic models still forward `dimensions`, simplified the detection regex to `/\.nvidia\.com/i`, and added a `.nvidia.com` baseURL detection test.

The follow-up commits are:
- `7871fa0` Add NVIDIA detection test and update imports
- `1507994` Refactor NVIDIA compatibility check regex

I also re-checked the earlier `dimensions` regression and didn't find another leftover behavior bug beyond the description drift and the two nits you called out.

vicjayjay:
Follow-up commits are pushed and the PR is ready for re-check. There are currently 3 workflow runs awaiting maintainer approval on this PR before they can execute: `CI`, `Claude Code Review`, and `Claude Code`.

AliceLJY:
看了 follow-up commits，改动没问题。

不过有个情况同步一下：我们昨天发了 v1.1.0-beta.10，main 上有比较大的变动（OpenClaw 2026.3+ hook 适配 + 若干 fix）。麻烦你 rebase 到最新 main，确认没有冲突，CI 跑过之后我们合。

vicjayjay:
Rebased onto latest master (5c477c1). One trivial conflict in `getProviderLabel()` — master added an Azure OpenAI label line, our branch added the NVIDIA NIM line. Kept both.

Commits are clean on top of v1.1.0-beta.10. Waiting for maintainer CI approval to confirm green.

AliceLJY:
Thanks for the rebase, commits look clean on top of beta.10.

Before we merge, there are two things I'd like addressed — embedding provider detection is load-bearing infrastructure and we need to be careful here.

### 1. Detection scope is too broad

Right now any `*.nvidia.com` endpoint gets the `nvidia` profile, which means `buildPayload()` will inject `input_type` into every request hitting that host. NVIDIA also serves non-retriever models on `.nvidia.com` (e.g. NV-CLIP, future multimodal endpoints) that don't expect `input_type` — this could silently break those requests or return 400s that are very hard for users to diagnose.

**Ask:** Tighten the detection. Either:
- Narrow the host match to known retriever-specific subdomains/paths, or
- Keep the broad host match but only inject `input_type` when `taskQuery`/`taskPassage` is explicitly configured (so non-retriever users who don't set those fields are unaffected)

### 2. Detection ordering — host vs model prefix precedence

The NVIDIA check currently runs *after* Jina/Voyage model-prefix checks. So a `.nvidia.com` endpoint serving a model named `voyage-xxx` or `jina-xxx` would be classified as Voyage/Jina instead of NVIDIA, sending wrong provider-specific fields (`task`/`output_dimension` instead of `input_type`/`dimensions`).

If the intent is "NVIDIA host semantics win", the `.nvidia.com` baseURL check should come before model-prefix heuristics. If model prefix should win, document why.

**Ask:** Either move the baseURL portion of the NVIDIA check above the model-prefix providers, or add a comment explaining the intended precedence.

### 3. Negative test coverage

The current 6 tests are all happy-path. Please add:
- `.nvidia.com` baseURL + conflicting model prefix (e.g. `jina-xxx`) → verify correct profile wins
- `.nvidia.com` baseURL + no `taskQuery`/`taskPassage` configured → verify no `input_type` is injected (if you go with option B in point 1)

---

None of these are hard to fix and the core implementation is solid. Just want to make sure we don't ship edge cases that are painful to debug for users. CI approval is pending — I'll approve the workflow runs so they can start in parallel while you work on the above.

vicjayjay:
@AliceLJY Addressed all three points in `e6e705d`:

**1. Detection scope** — Already safe by design. `buildPayload()` only injects `input_type` when *both* the profile defines `taskField` AND the user explicitly configures `taskQuery`/`taskPassage` (line 658: `if (this._capabilities.taskField && task)`). Users hosting NV-CLIP or other non-retriever models on `.nvidia.com` who don't set those fields are unaffected. Added a comment documenting this guard.

**2. Detection ordering** — Split detection into two phases:
- **Host-based checks first** (`.nvidia.com`, `api.jina.ai`, `api.voyageai.com`) — endpoint owner semantics win
- **Model-prefix fallback second** (`jina-`, `voyage`, `nvidia/`, `nv-embed`) — only when host didn't match

A `.nvidia.com` endpoint serving a `jina-xxx` model now correctly gets the NVIDIA profile.

**3. Negative tests** — Added 2 new tests (8 total, all passing):
- `.nvidia.com` baseURL + `jina-embeddings-v3` model → NVIDIA wins (verifies precedence)
- `nvidia/nv-clip-v1` model without `taskQuery`/`taskPassage` → no `input_type` injected (verifies safety)

Existing `embedder-error-hints` tests also pass — no regressions.

AliceLJY:
@rwmjhb 恒哥看一下这个 PR，作者两轮修完了，我和 rwmjhb 之前都看过，你最终确认下能否 merge 👀

rwmjhb:
`detectEmbeddingProviderProfile()` now treats any `.nvidia.com` host as `nvidia`, and
  `buildPayload()` will then send `input_type` whenever `taskQuery` / `taskPassage` is
  configured. NVIDIA’s OpenAI-compatible embedding docs split behavior by model family here: NV-
  Embed models use `input_type`, but GTE/GTR models do not accept that parameter.

  Because our config/docs already encourage `taskQuery` / `taskPassage`, this can break valid
  `.nvidia.com` deployments that are not NV-Embed. Can we narrow the NVIDIA-specific payload
  path to the model families that actually support `input_type`, instead of keying off host
  alone?

  Docs: https://docs.nvidia.com/nim/nemo-retriever/text-embedding/1.13.0/use-the-api-openai.html

PR #270, opened 2026-03-19
fix: use resolveToolContext in memory_forget and memory_update

### LINKED ISSUE #266: Bug: memory_forget fails with 'outside accessible scopes' even when agent should have access

AliceLJY:
## Summary
- `memory_forget` and `memory_update` used the original (stale) `context` instead of `resolveToolContext()` for scope resolution, causing "outside accessible scopes" errors when no explicit scope was passed
- Aligned both tools with the same `resolveToolContext()` pattern already used by `memory_recall`

Fixes #266

## Test plan
- [ ] `memory_store` without scope → `memory_forget(memoryId=...)` without scope → should succeed
- [ ] `memory_store` without scope → `memory_update(memoryId=..., importance=0.9)` without scope → should succeed
- [ ] Explicit `scope` parameter still works as before
- [ ] Scope access denial still works for unauthorized scopes


rwmjhb:
This looks correct. Using `resolveToolContext()` here is consistent with the existing recall path and should fix the stale-context scope issue in `memory_forget` / `memory_update`.

Only recommendation: please add a regression test covering the host flow where agent info is available at tool creation time but missing at execute time. No other blocking issue from my side.

chenjiyong:
I re-validated this PR from a clean isolated worktree and the results look good.

This change is beneficial to the memory architecture because it makes `memory_forget` and `memory_update` use the same runtime context resolution path as `memory_recall` and `memory_store`. That improves consistency at the tool layer and reduces the risk of scope drift when requests come from different agents or session keys.

Validation I ran:
- `openclaw-host-functional.mjs`
- `openclaw-agent-e2e.mjs`
- full `npm test`

All passed. I also directly exercised the updated runtime-context paths and confirmed that `memory_forget` / `memory_update` now resolve scope from runtime `toolCtx` / `sessionKey` instead of falling back to the static context agent. Looks good to merge.

PR #292, opened 2026-03-20
feat(retrieval): add manual BM25 query expansion and diagnostics

fryeggs:
## Summary

This is **PR 1/4** in a series that extends `memory-lancedb-pro` in layered steps.

This PR is intentionally the smallest and lowest-risk step. It improves **manual retrieval ergonomics** without changing auto-recall behavior.

It adds:

1. BM25 query expansion for manual / CLI retrieval only
2. Retrieval diagnostics for manual / CLI search
3. CLI debug output for retrieval diagnostics

## Why

Users often search with colloquial phrases such as `挂了`, `卡住`, or `报错`, while stored memories often contain more technical wording like `crash`, `timeout`, `error`, or `exception`.

The vector leg already helps semantically, but the BM25 leg still matters for exact-term boosting and mixed-language memory bases. This PR improves that explicit/manual lookup path while deliberately leaving auto-recall unchanged.

## Scope and safety

- no change to auto-recall query behavior
- no change to vector query text
- query expansion is limited to manual / CLI retrieval
- focused tests cover colloquial expansion, false-positive protection, gating, and debug output

## Companion PRs in this series

- PR 2/4: scored capture pipeline and ingestion safeguards
- PR 3/4: standalone runtime and MCP surface
- PR 4/4: Claude / Codex host integrations

## Validation

```bash
node --test test/query-expander.test.mjs
```

Passed locally.

### DIFF
--- a/cli.ts
+++ b/cli.ts
@@ -409,6 +409,71 @@ function formatJson(obj: any): string {
   return JSON.stringify(obj, null, 2);
 }
 
+function formatRetrievalDiagnosticsLines(diagnostics: {
+  originalQuery: string;
+  bm25Query: string | null;
+  queryExpanded: boolean;
+  vectorResultCount: number;
+  bm25ResultCount: number;
+  fusedResultCount: number;
+  finalResultCount: number;
+  stageCounts: {
+    afterMinScore: number;
+    rerankInput: number;
+    afterRerank: number;
+    afterHardMinScore: number;
+    afterNoiseFilter: number;
+    afterDiversity: number;
+  };
+  dropSummary: Array<{ stage: string; dropped: number; before: number; after: number }>;
+  failureStage?: string;
+  errorMessage?: string;
+}): string[] {
+  const topDrops =
+    diagnostics.dropSummary.length > 0
+      ? diagnostics.dropSummary
+          .slice(0, 3)
+          .map(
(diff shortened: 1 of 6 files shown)


rwmjhb:
This error path is reading diagnostics from the wrong retriever instance.

  When `context.embedder` is present, `runSearch()` executes against a fresh retriever created
  by `getSearchRetriever()`, not `context.retriever`. But the outer `catch` block still reads
  diagnostics from `context.retriever.getLastDiagnostics?.()`.

  In the normal runtime path, CLI registration does pass `embedder`, so on `memory-pro search
  --debug` / `--json --debug` failures we can lose the diagnostics payload entirely and only
  return the error.

  I verified this with a minimal repro against the PR branch. The current tests don’t catch it
  because they stub `context.retriever` directly and don’t exercise the `embedder ->
  createRetriever()` path.

  Can we thread the last-used search retriever (or its diagnostics) through the failure path so
  debug output comes from the instance that actually executed the search?

rwmjhb:
Closing this in favor of #438.

I recreated the change on a maintainer-owned branch because the original PR went stale, resolved the outstanding conflict/review issues, and trimmed the unrelated `vllm` rerank scope so the replacement stays focused on manual/CLI query expansion plus retrieval diagnostics.

Issue #272, opened 2026-03-19
Proposal: add Gemini Embedding 2 support and discuss a sibling multimodal ingest plugin


gexinggexing:
I split the work into two parts instead of turning `memory-lancedb-pro` into an all-in-one multimodal plugin.

Part 1:

- add `gemini-embedding-2-preview` model support
- update chunking/context-limit mappings
- fix CLI/service timer behavior so plugin CLI commands exit cleanly in local runs
- update docs/config hints

Part 2:

- build a separate `memory-multimodal-ingest` plugin
- support `image / audio / video / pdf` ingest
- store vectors + metadata in LanceDB
- keep the core text memory runtime unchanged

Why split it:

- the current upstream project is still optimized around text memory
- multimodal ingestion adds very different storage and pipeline concerns
- a sibling plugin keeps migration reversible and reduces regression risk

Validated locally:

- Gemini Embedding 2 API accepted image/audio/video/pdf inputs
- local OpenClaw plugin ingest succeeded for all four modalities
- LanceDB-backed search worked with text queries against stored media records

Repository with the sibling plugin MVP:

- https://github.com/gexinggexing/memory-multimodal-ingest

If this direction makes sense, I can open:

1. a small PR for `gemini-embedding-2-preview` + CLI/runtime fixes
2. a separate discussion/PR proposal for the sibling multimodal plugin path

gexinggexing:
Quick follow-up: I prepared the small PR branch locally, but after diffing against current master there is no remaining core delta on my side. The current upstream already appears to include the Part 1 changes I planned to send: gemini-embedding-2-preview support, chunk/context-limit mappings, the timer unref() fix, and the related docs/config updates.

So I won't open an empty PR for that part. The remaining work worth discussing is the sibling multimodal plugin path.

AliceLJY:
Hi @gexinggexing, thanks for the well-structured proposal and the working MVP!

The split approach makes sense — keeping multimodal ingest as a sibling plugin avoids bloating the core text memory runtime.

For Part 1 (Gemini Embedding 2 support + CLI fixes), feel free to open a PR — happy to review.

AliceLJY:
Closing as stale — no activity for ~2 months and no PR in progress. Still a welcome enhancement; please reopen or open a PR if you'd like to pursue it.

PR #296, opened 2026-03-21
docs: link experimental multimodal companion

gexinggexing:
## Summary
- add an ecosystem note for the experimental multimodal companion path
- link the upstream discussion for the sibling-plugin direction
- link the prototype repository so interested users can evaluate it without confusing it with the core plugin

## Why
Part 1 from issue #272 is already in current `master`, so I did not open an empty core PR.
This docs-only PR keeps the accepted multimodal direction discoverable without changing the core runtime boundary.

Related:
- Follow-up discussion: #275
- Original issue: #272


rwmjhb:
Closing — linking an external early-stage experimental project in the main README is premature. Once the multimodal direction reaches consensus in Discussion #275, this can be revisited. Thanks for the contribution!

PR #303, opened 2026-03-22
Add a non-invasive Markdown compatibility path for legacy memory continuity

whilyAk3sOn:
## Summary

This PR adds a **non-invasive Markdown compatibility path** for `memory-lancedb-pro` so plugin-managed durable memory remains usable through OpenClaw's original Markdown / SQLite-compatible foundation.

The goal is to make plugin adoption less destructive and more reversible:

- a user can enable `memory-lancedb-pro` mid-stream,
- use it during normal operation,
- later disable or uninstall it,
- and still retain practical continuity through legacy-compatible memory artifacts.

Instead of mixing plugin-managed memory into the original human-authored daily log path, this PR introduces a parallel compatibility subtree:

```text
memory/
  YYYY-MM-DD.md
  plugins/
    memory-lancedb-pro/
      README.md
      YYYY-MM-DD.md
```

This keeps the original daily-memory path untouched while exposing plugin-managed durable memory through a compatibility layer that legacy flows can still recognize.

> Scope note: this PR establishes the compatibility / continuity path, but it does **not** yet add a dedicated one-shot command for automatically backfilling all pre-existing workspace Markdown into LanceDB.

---

## Why

`memory-lancedb-pro` should behave like an enhancement layer, not a one-way memory migration.

Users should be able to:

- enter the plugin layer at time A,
- leave the plugin layer at time B,
- and continue from OpenClaw's original Markdown / SQLite memory foundation without being effectively locked into the plugin's internal database.

Without a compatibility path, durable memory created while the plugin is enabled can become practically trapped inside the plugin layer.

This PR addresses that by:

1. writing plugin-managed durable memory into a legacy-compatible Markdown subtree,
2. teaching upgrade scan / planner flows to explicitly recognize that subtree,
3. teaching Markdown preview/import flows to re-consume that subtree during re-entry,
4. exposing those compatibility sources in CLI/report output.

---

### 1. Added a non-invasive compatibility subtree under `memory/plugins/`

Plugin-managed compatibility artifacts now live under:

```text
memory/plugins/memory-lancedb-pro/
```

instead of modifying the original daily log path:

```text
memory/YYYY-MM-DD.md
```

A `README.md` is also created inside the subtree so later users can understand why these files exist.

### 2. Added runtime Markdown mirror support for durable writes

The plugin now writes compatibility Markdown entries for durable memory writes under:

```text
memory/plugins/memory-lancedb-pro/YYYY-MM-DD.md
```

### 3. Extended smart extraction to emit compatibility persistence events

`SmartExtractor` now supports an `onPersist` callback and forwards `persistMeta.agentId`, allowing compatibility mirror writes to land in the correct agent workspace subtree for durable smart-extraction paths, including:

- create
- merge
- supersede
- contextualize
- contradict

### 4. Treated the plugin compatibility subtree as a legacy-compatible Markdown source

Upgrade detection and planning now explicitly recognize:

```text
memory/plugins/memory-lancedb-pro/*.md
```

as a real legacy-compatible Markdown source.

This improves continuity during plugin disable / re-enable cycles and reduces duplicate-import risk by lowering same-agent SQLite import priority when compatible Markdown already exists.

### 5. Added first-class import/preview support for plugin compatibility Markdown

The Markdown preview/import layer now supports a dedicated source type:

- `plugin-compat-md`

It can:
- extract the durable text payload from compatibility lines,
- preserve `scopeHint`,
- treat these entries as durable by default.

### 6. Exposed plugin compatibility sources in upgrade-scan output

Upgrade scan reporting now explicitly surfaces plugin compatibility sources in both:

- machine-readable JSON output
- human-readable CLI output

This includes:
- plugin compatibility workspace count
- plugin compatibility file count
- explicit mention of the plugin compatibility subtree in workspace output
- clearer overlap messaging when legacy-compatible Markdown already exists

### 7. Includes supporting agent-scope groundwork needed by the continuity path

This PR also includes agent-scoped defaulting / upgrade-scan groundwork that keeps compatibility writes and legacy-source recognition aligned with per-agent workspaces instead of assuming a single flat global path.

---

## Design notes

This PR intentionally uses a parallel subtree:

```text
memory/plugins/memory-lancedb-pro/
```

rather than writing into the original:

```text
memory/YYYY-MM-DD.md
```

This keeps the solution:

- non-invasive
- reversible in practice
- explainable
- compatible with OpenClaw's existing Markdown / SQLite foundation

The compatibility subtree is not intended to replace the normal daily log. It exists to preserve continuity across plugin enable / disable / re-enable cycles.

---

## Scope boundary

This PR does **not** directly mutate or manage OpenClaw core SQLite files.

Instead, it preserves continuity by:

1. writing plugin-managed durable memory into a legacy-compatible Markdown subtree,
2. ensuring scan / planner / import flows explicitly recognize that subtree,
3. allowing OpenClaw's original Markdown / SQLite-compatible foundation to remain usable as the compatibility base.

That keeps the implementation non-invasive while still making the continuity path explicit and testable.

---

## Testing

Added or extended regression coverage for:

- `md-mirror-writer.test.mjs`
- `context-support-e2e.mjs`
- `temporal-facts.test.mjs`
- `first-run-init.test.mjs`
- `upgrade-planner.test.mjs`
- `upgrade-scan-cli.test.mjs`
- `md-import.test.mjs`
- `agent-scoped-defaults.test.mjs`

Covered scenarios include:

- compatibility files are written under `memory/plugins/memory-lancedb-pro/`
- `memory/YYYY-MM-DD.md` remains untouched
- smart extraction durable paths emit compatibility mirror events
- plugin compatibility Markdown is recognized during upgrade candidate detection
- plugin compatibility Markdown lowers same-agent SQLite import priority via overlap detection
- plugin compatibility Markdown is parsed as a dedicated import source (`plugin-compat-md`)
- upgrade-scan JSON output exposes plugin compatibility counts
- human-readable upgrade-scan output explicitly mentions plugin compatibility subtree presence
- agent-scoped defaults stay aligned with per-agent workspace continuity behavior

Full test suite verification:

```bash
npm test
```

---

## Reviewer guide

A good review order is:

1. compatibility subtree path design
2. runtime Markdown mirror writes
3. smart extraction persistence hooks
4. agent-scoped defaulting / workspace alignment
5. upgrade detection / planner continuity logic
6. Markdown import / preview continuity logic
7. upgrade-scan output visibility
8. regression coverage


whilyAk3sOn:
One clarification on scope: this PR makes the continuity path explicit, but it does **not** yet introduce a fully productized “backfill all pre-existing workspace Markdown into LanceDB” command.

What this PR *does* provide is the foundation required for that workflow to be safe and reversible:

- plugin-managed durable memory now writes into a parallel compatibility subtree under `memory/plugins/memory-lancedb-pro/`
- upgrade scan / planner explicitly recognize that subtree as legacy-compatible Markdown
- Markdown preview/import now has first-class support for plugin compatibility Markdown (`plugin-compat-md`)
- re-entry / disable / re-enable flows can reason about the compatibility subtree instead of treating it as opaque text

On my local OpenClaw deployment, I also validated the historical backfill path operationally by importing pre-existing `MEMORY.md` and dated `memory/*.md` files into the upgraded plugin using stable IDs and scoped import bundles. That local validation is intentionally **not** claimed here as a user-facing automatic feature of this PR.

So the intended framing is:

- this PR **adds the reversible compatibility layer and continuity plumbing**
- a follow-up can still productize historical Markdown backfill as a dedicated command / guided workflow if desired

I wanted to make that boundary explicit so the PR does not overclaim while still showing that the continuity model has been exercised beyond purely theoretical design.

rwmjhb:
Closing — this PR adds 5700+ lines including 6 new modules, 6 test files, and 7 phase-2 design documents, essentially planning a full architectural overhaul (worker decomposition, runtime coexistence, import boundaries). This scope goes well beyond a migration utility.

If legacy memory migration is needed, we'd suggest building it as a standalone CLI tool or separate package. That way the migration logic can evolve independently without expanding the core plugin's maintenance surface. Thanks for the contribution!

PR #311, opened 2026-03-22
feat: cross-process file lock + CLAWTEAM_MEMORY_SCOPE support

AliceLJY:
## Summary

Addresses [ClawTeam-OpenClaw#24](https://github.com/win4r/ClawTeam-OpenClaw/issues/24): ClawTeam spawns each agent as a separate process, causing concurrent LanceDB write races and memory scope isolation.

- **Cross-process file lock**: Wraps all write operations (`store`, `delete`, `update`, `bulkDelete`, `importEntry`) with `proper-lockfile` to prevent manifest conflicts when multiple agent processes write simultaneously
- **`CLAWTEAM_MEMORY_SCOPE` env var**: When set (e.g., `custom:team-myteam`), automatically extends the current agent's accessible scopes to include the shared team scope

## How it works

1. **File lock**: Acquires `{dbPath}/.memory-write.lock` with exponential backoff retries (5 retries, 100-2000ms) and 10s stale detection. The existing in-process `runSerializedUpdate()` queue is preserved; the file lock wraps it for inter-process safety.
2. **Env var**: ClawTeam sets `CLAWTEAM_MEMORY_SCOPE=custom:team-{team_name}` at spawn time → this plugin registers the scope definition and wraps `scopeManager.getAccessibleScopes()` to include the extra scopes for all agents in the process.
3. **Fully backward compatible**: Without the env var, behavior is unchanged. Without `proper-lockfile` being exercised (single process), overhead is minimal (one lock/unlock per write).

## Changes

| File | Change |
|------|--------|
| `package.json` | Add `proper-lockfile` dependency |
| `src/store.ts` | Add `runWithFileLock()`, wrap `store`/`delete`/`update`/`bulkDelete`/`importEntry` |
| `index.ts` | Read `CLAWTEAM_MEMORY_SCOPE`, register scope definitions, wrap `getAccessibleScopes` |

## Test plan

- [ ] Existing test suite passes
- [ ] Concurrent writes from multiple processes don't corrupt LanceDB
- [ ] `CLAWTEAM_MEMORY_SCOPE=custom:team-demo` extends agent scope access
- [ ] Without env var, no behavior change

Companion PR: [ClawTeam-OpenClaw#26](https://github.com/win4r/ClawTeam-OpenClaw/pull/26) (env var propagation + prompt injection)

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -86,6 +86,9 @@ interface PluginConfig {
   autoRecall?: boolean;
   autoRecallMinLength?: number;
   autoRecallMinRepeated?: number;
+  autoRecallMaxItems?: number;
+  autoRecallMaxChars?: number;
+  autoRecallPerItemMaxChars?: number;
   captureAssistant?: boolean;
   retrieval?: {
     mode?: "hybrid" | "vector";
@@ -215,6 +218,14 @@ function resolveEnvVars(value: string): string {
   });
 }
 
+function resolveFirstApiKey(apiKey: string | string[]): string {
+  const key = Array.isArray(apiKey) ? apiKey[0] : apiKey;
+  if (!key) {
+    throw new Error("embedding.apiKey is empty");
+  }
+  return resolveEnvVars(key);
+}
+
 function resolveOptionalPathWithEnv(
   api: Pick<OpenClawPluginApi, "resolvePath">,
   value: string | undefined,
@@ -236,6 +247,11 @@ function parsePositiveInt(value: unknown): number | undefined {
     if (Number.isFinite(n) && n > 0) return Math.floor(n);
   }
   return undefined;
+}
+
+function clampInt(value: number, min: number, max: number): number {
+  if (!Number.isFinite(value)) return min;
+  return Math.min(max, Math.max(min, Math.floor(value)));
 }
 
 function resolveLlmTimeoutMs(config: PluginConfig): number {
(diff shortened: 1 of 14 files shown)


chenjiyong:
After looking at ClawTeam’s model more closely, I think this PR is more valuable than it first appeared.

ClawTeam is built around multiple agents with separate identities, worktrees, and inbox/task-based coordination, not a single shared agent context. In that model, `CLAWTEAM_MEMORY_SCOPE` makes sense as a shared memory visibility layer: it lets different team agents read from a broader team scope while still keeping their own default identity and workspace isolation.

That said, the current implementation mainly expands accessible/read scopes. It does not change the default write target for auto-capture, reflection, or session-memory flows. So for ClawTeam, this looks like a solid “shared recall/access” improvement, but not yet a full “team memory write-back” solution.

I still think the direction is worth merging for ClawTeam use cases, especially because concurrent writers are realistic in a swarm setup and the cross-process file lock helps there as well. My main remaining concern is maintainability: the ClawTeam behavior is currently injected by patching `getAccessibleScopes()` at plugin init time, and there are no dedicated tests for either `CLAWTEAM_MEMORY_SCOPE` behavior or the new cross-process locking path.

AliceLJY:
Hi @chenjiyong, thanks for the thorough review! Addressed your maintainability concerns in commit `a88a382`:

**Changes:**
- Extracted the inline `CLAWTEAM_MEMORY_SCOPE` logic from `index.ts` into a dedicated `src/clawteam-scope.ts` module with two testable functions: `parseClawteamScopes()` and `applyClawteamScopes()`
- Added `test/clawteam-scope.test.mjs` — 14 test cases covering env var parsing, scope registration, scope extension, and backward compatibility
- Added `test/cross-process-lock.test.mjs` — 5 test cases covering sequential writes, concurrent writes (no data loss), concurrent updates (no corruption), and lock release
- Type-narrowed `applyClawteamScopes()` to `MemoryScopeManager` specifically, and fixed a potential array mutation bug (now copies before pushing)
- Updated JSDoc to clarify that extending `getAccessibleScopes()` also affects `isAccessible()` checks (not read-only as previously documented)

**Not in scope (follow-up):**
- Write-back to team scope (as you noted, `getDefaultScope` is unchanged)
- Cross-process fork-based testing (current tests verify same-process concurrency; true multi-process safety is left for integration tests)

Ready for re-review when you have a chance!

PR #313, opened 2026-03-23
feat: recallMode with adaptive intent routing and tiered injection

AliceLJY:
## Summary

- Adds `recallMode` config option (`full` | `summary` | `adaptive` | `off`) for granular control over auto-recall behavior
- **Adaptive mode** (new): automatically analyzes query intent to determine category routing and injection depth — inspired by [OpenViking](https://github.com/volcengine/OpenViking)'s hierarchical retrieval intent routing
- Intent detection is pure pattern matching (no LLM calls) for zero extra latency
- Supports both English and Chinese query patterns

### How adaptive mode works

| Query type | Example | Category filter | Injection depth |
|---|---|---|---|
| Preference | "What's my coding style?" | `preference` | L0 (one-line) |
| Decision | "Why did we choose Postgres?" | `decision` | L1 (medium) |
| Entity | "Who is the auth team lead?" | `entity` | L1 (medium) |
| Event | "What happened last deploy?" | `event` | Full text |
| Fact | "How does the API work?" | `fact` | L1 (medium) |
| Ambiguous | "Write a sort function" | No filter | L0 (compact) |

Automatic fallback: if category filtering yields zero results, retries without filter.

### New files
- `src/intent-analyzer.ts` — `analyzeIntent()`, `applyCategoryBoost()`, `formatAtDepth()`
- `test/intent-analyzer.test.mjs` — 20 tests (all passing)

### Config example
```json
{
  "recallMode": "adaptive"
}
```

## Test plan
- [x] 20 unit tests for intent analysis, category boost, and tiered formatting
- [ ] Manual test with OpenClaw agent: verify adaptive recall injects correct depth
- [ ] Verify fallback works when category filter yields zero results
- [ ] Verify backward compat: existing `recallMode: "full"` unchanged

## References
- OpenViking hierarchical retrieval: `openviking/retrieve/hierarchical_retriever.py`
- [memory-lancedb-pro-enhancements](https://github.com/CortexReach/memory-lancedb-pro-enhancements) roadmap

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #254, opened 2026-03-18
Harden auto-recall with timeout and circuit breaker

shardbigwhite-dotcom:
## Summary
- bound auto-recall on `before_agent_start` with a small total timeout
- add a lightweight circuit breaker after repeated recall failures
- fast-fail the circuit immediately for known fatal retrieval/storage/config errors
- skip recall safely instead of risking agent-start / Control UI stalls

## Why
This addresses the failure mode described in #253: when retrieval is unhealthy, `autoRecall` should degrade to "no memory recall" instead of making the request appear hung.

## Notes
- startup self-checks were already fire-and-forget and time-bounded
- this patch hardens the actual user-facing auto-recall path
- package-lock changes were intentionally excluded from this PR

Issue #218, opened 2026-03-15
[Bug] retriever.test() times out on startup even with working embedding API


Stancool:
### Plugin Version

1.1.0-beta.8

### OpenClaw Version

2026.3.12

## Environment
- OS: Windows 11 + WSL2 (Ubuntu)
- - Node.js: v22.22.1
- - - openclaw: 2026.3.12
- - - - memory-lancedb-pro: 1.1.0-beta.8
## Embedding Configuration
```json
{
  "embedding": {
    "apiKey": "ollama",
    "model": "mxbai-embed-large",
    "baseURL": "http://172.23.144.1:11434/v1",
    "dimensions": 1024
  }
}
```

## Problem

Gateway startup fails with:

```
memory-lancedb-pro: startup checks failed: Error: retriever.test() timed out after 8000ms
```

## Investigation

Embedding API works fine - Direct curl test returns in <1 second:

```
curl -s http://172.23.144.1:11434/v1/embeddings \
  -H 'Content-Type: application/json' \
  -d '{"model": "mxbai-embed-large", "input": "test"}'
# Returns valid embedding in ~1s
```

Network is fast - WSL to Windows ping: 0.4ms

Tried fixes that didn't work:
- Deleted and recreated empty lancedb-pro database
- - Pre-warmed Ollama models before gateway start
- - - Multiple gateway restarts
## Suspected Cause

`retriever.test()` calls `retrieve()` which may have issues with:
- Empty database initialization
- - Table/index creation on first query
- - - Some internal operation not covered by the 8s timeout
## Logs

```
13:14:46 [plugins] memory-lancedb-pro@1.1.0-beta.8: plugin registered
13:15:26 [gateway] memory-lancedb-pro: startup checks failed: Error: retriever.test() timed out after 8000ms
```

Note: 40 seconds elapsed between registration and timeout error (expected 8s).

### Expected Behavior

Startup check should complete successfully when embedding API is responsive.

### Steps to Reproduce

1. Set up embedding config with Ollama (mxbai-embed-large) on WSL2
2. 2. Start OpenClaw gateway with memory-lancedb-pro plugin
3. 3. Observe: plugin registers successfully at 13:14:46
4. 4. See error: `memory-lancedb-pro: startup checks failed: Error: retriever.test() timed out after 8000ms` at 13:15:26 (40 seconds later)

### Error Logs / Screenshots

```shell
[...]

Issue #60, opened 2026-03-05
Startup checks: retriever.test() frequently times out after 8000ms (OpenClaw 2026.3.2, WSL2)


i0ivi0i:
### Summary
In OpenClaw (WSL2, OpenClaw 2026.3.2), `memory-lancedb-pro` frequently logs a startup warning:

```
memory-lancedb-pro: startup checks failed: Error: retriever.test() timed out after 8000ms
```

Gateway keeps running (checks are fire-and-forget), but this suggests the retrieval test can hang/slow and may correlate with slow searches.

### Environment
- OpenClaw: 2026.3.2 (85377a2)
- OS: WSL2 Ubuntu (kernel 6.6.87.2)
- Plugin: memory-lancedb-pro 1.0.31
- Embeddings: Jina OpenAI-compatible
  - model: `jina-embeddings-v3`
  - baseURL: `https://api.jina.ai/v1`
- Retrieval: `hybrid` with `rerank=cross-encoder` (`jina-reranker-v3`)

### Observations
- Repeated occurrences in systemd journal:
  - 01:28:05 `retriever.test() timed out after 8000ms`
  - 01:34:09 `retriever.test() timed out after 8000ms`
  - 01:51:41 `retriever.test() timed out after 8000ms`
- Manual query works but can be slow (example: `openclaw memory-pro search "ALPHA-18789"` took ~25s on one run).

### Expected
Startup checks should be reliable or provide more actionable diagnostics (which stage timed out: FTS probe? rerank call? embedding? LanceDB query?).

### Suggestions
- Add per-stage timing logs inside `retriever.test()` (and maybe a smaller internal timeout for external rerank calls).
- If rerank is enabled, ensure test does not block/hang on reranker endpoint; degrade gracefully.
- Consider making the timeout configurable via plugin config.

Thanks!

win4r:
Thanks for the detailed report — this warning is coming from the 8s *startup check wrapper* around `retriever.test()` (fire-and-forget), so the gateway will keep running even when the check times out.

A couple likely reasons, especially with Jina + WSL2:

1) **External Jina latency / connectivity**: `retriever.test()` runs a real `retrieve({ query: "test query", limit: 1 })`, which includes **a query embedding call** (`embedder.embedQuery`). Unlike the rerank call (which has an internal 5s abort), the embedding path currently relies on the provider/SDK defaults and can be slow/hang on bad network/DNS/proxy, so the *overall* 8s wrapper may trip.

[...]

PR #316, opened 2026-03-23
feat: add configurable auto-recall timeout with safer default

### LINKED ISSUE #314: [Enhancement] auto-recall 3000ms timeout is too aggressive and should be configurable
## Summary

`memory-lancedb-pro` currently hard-codes auto-recall startup timeout to **3000ms**, and that default appears too aggressive for real OpenClaw deployments using remote embedding APIs.

In our environment, auto-recall does not fail consistently because retrieval is broken; instead it **sometimes succeeds and sometimes times out** under normal load, which means the default timeout is sitting too close to the real retrieval latency budget.

This makes long-term memory injection flaky even when the system is otherwise healthy.

## Current behavior

On `memory-lancedb-pro@1.1.0-beta.9`, the plugin logs warnings like:

```text
memory-lancedb-pro: auto-recall timed out after 3000ms; skipping memory injection to avoid stalling agent startup
```

At the same time, in adjacent runs / adjacent agents, we also see successful logs like:

```text
memory-lancedb-pro: injecting 3 memories into context for agent qa
```

So this is not a simple “retrieval is down” case. It looks more like the default timeout is too tight and gets tripped by normal network jitter / gateway concurrency / retrieval variance.

## Environment

- OpenClaw: `2026.3.13`
- Plugin: `memory-lancedb-pro@1.1.0-beta.9`
- autoRecall: `true`
- autoCapture: `true`
- smartExtraction: `false`
- retrieval mode: `hybrid`
- embedding model: `text-embedding-3-small`
- memory store size: ~1091 memories

## Why this seems too aggressive

In the plugin source, the timeout is currently hard-coded:

```ts
const AUTO_RECALL_TIMEOUT_MS = 3_000;
```

In our environment, even a cold manual recall path is already close to that threshold. A simple manual search took about **3369ms** end-to-end on one run, which means the `3000ms` startup budget is too close to the real-world latency envelope.

That creates a bad UX pattern:

- memory injection is flaky rather than predictably on/off
- startup sometimes includes useful memory context, sometimes silently skips it
- operators have no config knob to tune behavior for remote embeddings or slower networks

## Expected behavior

One of these would be much safer:

1. **Increase the default auto-recall timeout** to something like `8000ms`
2. **Make auto-recall timeout configurable** in plugin config
3. Ideally do both: higher default + config override

## Suggested fix

Suggested direction:

- add config like `autoRecallTimeoutMs`
- default it to something less brittle than `3000ms` (for example `8000ms`)
- keep the current “skip injection instead of stalling forever” behavior, which is good

That would preserve startup safety while making recall more reliable for hosted / remote embedding setups.

## Related issues

This feels related but not identical to:

- #137 Heartbeat / NO_REPLY turns may still trigger autoRecall injection and cause compaction-related latency
- #60 Startup checks: `retriever.test()` frequently times out after 8000ms
- #218 `retriever.test()` times out on startup even with working embedding API

Those issues point in the same direction: retrieval latency is real and environment-dependent, so hard-coded tight budgets are fragile.

## Extra note

There is also a framework-level OpenClaw issue about `before_agent_start` hooks needing stronger timeout guardrails (`openclaw/openclaw#48534`), but this report is specifically about the **plugin-side auto-recall timeout default being too low / not configurable**.

AliceLJY:
## Summary

- Add `autoRecallTimeoutMs` config option to control auto-recall startup timeout (default `8000ms`)
- Wrap `orchestrateDynamicRecall` in `Promise.race` with the configurable timeout
- On timeout, log a warning and skip memory injection gracefully (no agent startup stall)

## Problem

After the `orchestrateDynamicRecall` refactor, the `before_agent_start` hook had **no timeout guard** — slow embedding API calls could stall agent startup indefinitely. The previous beta.9 had a hard-coded 3000ms timeout which was too aggressive for remote embedding setups; the current main lost the timeout entirely.

## Usage

```json
{
  "autoRecallTimeoutMs": 12000
}
```

No change needed for existing users — default is `8000ms`, up from the old `3000ms`.

## Changes

**1 file changed, 17 insertions(+), 1 deletion(-)**

- `index.ts`: Add `autoRecallTimeoutMs` to `PluginConfig`, add `DEFAULT_AUTO_RECALL_TIMEOUT_MS` constant, wrap recall pipeline in `Promise.race` timeout, parse config value in config resolver

## Related

- #313 — `recallMode` config (complementary: controls recall granularity)
- #254 — auto-recall circuit breaker (complementary: controls failure handling)
- #60, #218 — `retriever.test()` startup timeout issues (same latency theme)

Closes #314 — thanks @chinaboard for the detailed report.

## Test plan

- [x] All 37 existing tests pass
- [ ] Verify default 8000ms timeout applies when `autoRecallTimeoutMs` is not set
- [ ] Verify custom timeout (e.g. `12000`) is respected
- [ ] Verify timeout triggers graceful skip with warning log, not a crash

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -75,6 +75,7 @@ interface PluginConfig {
   autoRecallExcludeReflection?: boolean;
   autoRecallMaxAgeDays?: number;
   autoRecallMaxEntriesPerKey?: number;
+  autoRecallTimeoutMs?: number;
   captureAssistant?: boolean;
   retrieval?: {
     mode?: "hybrid" | "vector";
@@ -260,6 +261,7 @@ const DEFAULT_AUTO_RECALL_TOP_K = 3;
 const DEFAULT_AUTO_RECALL_EXCLUDE_REFLECTION = true;
 const DEFAULT_AUTO_RECALL_MAX_AGE_DAYS = 30;
 const DEFAULT_AUTO_RECALL_MAX_ENTRIES_PER_KEY = 10;
+const DEFAULT_AUTO_RECALL_TIMEOUT_MS = 8_000;
 const DEFAULT_AUTO_RECALL_CATEGORIES: MemoryCategory[] = ["preference", "fact", "decision", "entity", "other"];
 const DEFAULT_REFLECTION_RECALL_MODE: ReflectionRecallMode = "fixed";
 const DEFAULT_REFLECTION_RECALL_TOP_K = 6;
@@ -1662,6 +1664,7 @@ const memoryLanceDBProPlugin = {
     // Auto-Recall: inject relevant memories before agent starts.
     // Default is OFF to prevent the model from accidentally echoing injected context.
     if (config.autoRecall === true) {
+      const autoRecallTimeoutMs = config.autoRecallTimeoutMs ?? DEFAULT_AUTO_RECALL_TIMEOUT_MS;
       api.on("before_agent_start", async (event, ctx) => {
         try {
           const agentId = ctx?.agentId || "main";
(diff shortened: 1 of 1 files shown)


AliceLJY:
已在 main 合并，关闭。

PR #343, opened 2026-03-25
feat: memory compaction — progressive summarization for stored memories

AliceLJY:
## Summary

- **`src/memory-compactor.ts`** (new): Pure, dependency-free compaction module implementing the *progressive summarization* pattern. Clusters semantically similar memories using greedy cosine-similarity expansion (seeded by highest-importance entry), then merges each cluster into a single refined entry — deduplicating lines, taking max importance, plurality-voting on category.
- **`store.ts`**: Adds `fetchForCompaction(maxTimestamp, scopeFilter, limit)` — fetches old entries *with* raw vectors (intentionally omitted from `list()` for performance; compaction needs them for similarity computation).
- **`index.ts`**: Wires in `memory_compact` MCP tool (gated by `enableManagementTools`) with `dry_run` / `min_age_days` / `similarity_threshold` / `scopes` parameters. Also registers an optional `gateway_start` auto-compaction hook with configurable cooldown.
- **`openclaw.plugin.json`**: `memoryCompaction` config schema + uiHints.
- **`test/memory-compactor.test.mjs`**: 23 tests covering `cosineSimilarity`, `buildClusters`, `buildMergedEntry`, and `runCompaction` (including dry-run, empty store, orthogonal entries, scan limit).

## Motivation

Without compaction, the LanceDB index accumulates fragments of the same idea across many sessions. Related memories like *"prefers dark mode"*, *"always uses dark theme"*, *"dark mode on all apps"* remain as three separate entries — each returned at retrieval time, wasting context budget and diluting precision. Progressive summarization collapses these into one high-quality entry, so the index gets *more* refined over time rather than noisier.

## Config example

```json
"memoryCompaction": {
  "enabled": true,
  "minAgeDays": 7,
  "similarityThreshold": 0.88,
  "cooldownHours": 24
}
```

## Manual use (requires `enableManagementTools: true`)

```
// Preview without writing
memory_compact { dry_run: true }

// Run for specific scopes
memory_compact { scopes: ["project:myapp"], similarity_threshold: 0.92 }
```

## Test plan

- [x] `node --test test/memory-compactor.test.mjs` — 23/23 pass
- [ ] Manual: enable `memoryCompaction.enabled: true` in a live OpenClaw instance; confirm `gateway_start` log shows compaction stats
- [ ] Manual: call `memory_compact` with `dry_run: true`, verify cluster count makes sense for your LanceDB
- [ ] Verify existing auto-capture / auto-recall behavior is unchanged when `memoryCompaction` is absent from config

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -24,6 +24,12 @@ import { appendSelfImprovementEntry, ensureSelfImprovementLearningFiles } from "
 import type { MdMirrorWriter } from "./src/tools.js";
 import { shouldSkipRetrieval } from "./src/adaptive-retrieval.js";
 import { parseClawteamScopes, applyClawteamScopes } from "./src/clawteam-scope.js";
+import {
+  runCompaction,
+  shouldRunCompaction,
+  recordCompactionRun,
+  type CompactionConfig,
+} from "./src/memory-compactor.js";
 import { runWithReflectionTransientRetryOnce } from "./src/reflection-retry.js";
 import { resolveReflectionSessionSearchDirs, stripResetSuffix } from "./src/session-recovery.js";
 import {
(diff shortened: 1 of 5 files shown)


NW15D:
Mar 26 18:03:18 OpenClaw node[1962]: 2026-03-26T18:03:18.841+00:00 [plugins] plugin tool name conflict (memory-lancedb-pro): memory_compact
it's ok ? it conflict with core_tools openclaw.
i use github version of plugin

PR #371, opened 2026-03-26
feat: add configurable rerank timeout (rerankTimeoutMs)

ggzeng:
## Summary

The cross-encoder rerank request timeout was hardcoded at 5 seconds, which is too aggressive for self-hosted rerank services (e.g. HuggingFace TEI via Infinity) that may need 6-7 seconds to respond.

## Changes

- Add `rerankTimeoutMs` optional field to `RetrievalConfig` interface
- Replace hardcoded `5000` with `config.rerankTimeoutMs ?? 10_000`
- Default raised from 5s to 10s for better compatibility with self-hosted endpoints

## Usage

```json
{
  "retrieval": {
    "rerank": "cross-encoder",
    "rerankTimeoutMs": 20000
  }
}
```

## Backwards Compatibility

Fully backwards compatible. Existing configs without `rerankTimeoutMs` will get the new 10s default.

### DIFF
--- a/src/retriever.ts
+++ b/src/retriever.ts
@@ -45,6 +45,8 @@ export interface RetrievalConfig {
   rerankModel?: string;
   /** Reranker API endpoint (default: https://api.jina.ai/v1/rerank). */
   rerankEndpoint?: string;
+  /** Reranker request timeout in milliseconds (default: 10000). */
+  rerankTimeoutMs?: number;
   /** Reranker provider format. Determines request/response shape and auth header.
    *  - "jina" (default): Authorization: Bearer, string[] documents, results[].relevance_score
    *  - "siliconflow": same format as jina (alias, for clarity)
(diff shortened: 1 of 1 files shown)


AliceLJY:
Hey @ggzeng, thanks for the contribution! Just a heads-up — PR #356 by @jlin53882 was submitted about 11 hours earlier and implements the same `rerankTimeoutMs` feature. It has already gone through a round of review feedback and includes additional changes (schema in `openclaw.plugin.json`, `PluginConfig` type in `index.ts`, explicit default in `DEFAULT_RETRIEVAL_CONFIG`, and timeout cleanup in the `finally` block).

One notable difference: your PR raises the default timeout from 5s to 10s, while #356 keeps the existing 5s default. The 10s default is arguably a better choice for self-hosted endpoints, but that's a design decision worth discussing in one place.

Could you coordinate with #356? If the higher default is important to you, it might be simplest to suggest that change on #356 rather than maintaining two parallel PRs for the same feature. If your approach differs in a meaningful way I'm missing, happy to hear more!

Thanks!

AliceLJY:
Hi @ggzeng — thank you for taking the time to submit this PR! Making `rerankTimeoutMs` configurable is a real need, and we appreciate you jumping in to address it.

After reviewing both contributions, we're going to close this in favor of #356, which was submitted a bit earlier and covers the same feature more comprehensively — it adds the JSON Schema definition in `openclaw.plugin.json`, updates the `PluginConfig` type in `index.ts`, wires up the config in `retriever.ts`, and includes timeout cleanup in a `finally` block.

That said, your PR raises a valid point that #356 doesn't address: **the default timeout value**. A 10s default may be more appropriate for self-hosted reranker setups where latency is higher. We'd love it if you could leave a comment on #356 suggesting they consider raising the default — that feedback would be genuinely useful there.

You can follow along here: https://github.com/CortexReach/memory-lancedb-pro/pull/356

Thanks again for the contribution, and we hope to see you around the project!

PR #349, opened 2026-03-25
feat: Add Bitwarden secrets and Anthropic distill support

Coke1120:
## Summary
- Add Bitwarden secret resolver (`src/secret-resolver.ts`) for secure API key management via Bitwarden CLI or env vars
- Add Anthropic LLM client (`src/llm-client.ts`) with distill/extraction support
- Add CLI entry point (`cli.ts`) exposing distill functionality
- Integrate secret resolution into lesson extract worker
- Add tests for secret resolver, LLM API key client, and plugin manifest regression
- Update plugin manifest and package dependencies

## Test plan
- [ ] Run `test/secret-resolver.test.mjs` — verify Bitwarden and env-var resolution paths
- [ ] Run `test/llm-api-key-client.test.mjs` — verify Anthropic client key handling
- [ ] Run `test/plugin-manifest-regression.mjs` — verify manifest schema unchanged
- [ ] Smoke-test distill via `examples/new-session-distill/` with a real session

Closes #348

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #398, opened 2026-03-29
feat: Bitwarden secret resolver and async register

Coke1120:
## Summary
- Add `src/secret-resolver.ts` supporting `${ENV_VAR}` and `bws://<secret-id>` Bitwarden Secrets Manager refs for embedding, rerank, and LLM API keys
- `register()` stays synchronous; async secret resolution runs in `initPromise`. Hooks await it before use; `register()` returns `initPromise` so callers that do await (tests, future host support) can wait for full init
- `service.start()` logs a clear warn if secret resolution failed, so users know why memory is unavailable
- bws access token passed via `BWS_ACCESS_TOKEN` env var instead of `--access-token` CLI arg to avoid process listing exposure
- Skip double `resolveEnvVars` in embedder for already-resolved keys
- Update `openclaw.plugin.json` docs to advertise `bws://` support on all API key fields
- Fix all tests to `await plugin.register()` and add `selfImprovement: { enabled: false }` to test configs that assert `command:new === undefined`

## Test plan
- [x] `node --test test/secret-resolver.test.mjs` — verify Bitwarden and env-var resolution paths
- [x] `npm test` — full test suite passes

Split from #349.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,14 +13,14 @@ A LanceDB-backed OpenClaw memory plugin that stores preferences, decisions, and 
 [![npm version](https://img.shields.io/npm/v/memory-lancedb-pro)](https://www.npmjs.com/package/memory-lancedb-pro)
 [![LanceDB](https://img.shields.io/badge/LanceDB-Vectorstore-orange)](https://lancedb.com)
 [![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE)
-
 <h2>⚡ <a href="https://github.com/CortexReach/memory-lancedb-pro/releases/tag/v1.1.0-beta.10">v1.1.0-beta.10 — OpenClaw 2026.3+ Hook Adaptation</a></h2>
 
 <p>
-  ✅ Fully adapted for OpenClaw 2026.3+ new plugin architecture<br>
-  🔄 Uses <code>before_prompt_build</code> hooks (replacing deprecated <code>before_agent_start</code>)<br>
-  🩺 Run <code>openclaw doctor --fix</code> after upgrading
+ ✅ Fully adapted for OpenClaw 2026.3+ new plugin architecture<br>
+ 🔄 Uses <code>before_prompt_build</code> hooks (replacing deprecated <code>before_agent_start</code>)<br>
+ 🩺 Run <code>openclaw doctor --fix</code> after upgrading
 </p>
+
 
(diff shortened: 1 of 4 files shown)


rwmjhb:
### Review: REQUEST-CHANGES

The Bitwarden secret resolver addresses a real operational gap (issue #349). Two issues need fixing before merge:

**Must fix:**

1. **Async `register()` breaks synchronous callers** — The OpenClaw loader does not await the returned promise, but CLI/hook registration now happens only after `await resolveSecretValues()`. Hooks and tools are `undefined` until the promise settles. Test files at `test/session-summary-before-reset.test.mjs:104` and `:144` also have unawaited `plugin.register(api)` calls that observe partially initialized state.

2. **Test suite fails** — `selfImprovement: { enabled: false }` was added to only one of three test config blocks in `plugin-manifest-regression.mjs`. The `sessionDefaultApi` and `sessionEnabledApi` fixtures still omit it, so `command:new` gets registered by the default-on selfImprovement feature and breaks the hook-absence assertions.

- Resolved Bitwarden secrets are passed through `resolveEnvVars()` again in `src/embedder.ts:422-424`. If a real API key contains the literal substring `${...}`, it will be re-interpreted as an env var template and throw. Bitwarden-resolved values should be treated as opaque.
- The Bitwarden access token is passed as `--access-token` CLI argument to `bws secret get`, making it visible in process listings. Consider using an env var or stdin pipe instead.
- Branch is behind `main` (`stale_base=true`) — please rebase.

rwmjhb:
## Review: Bitwarden secret resolver and async register

**Verdict: request-changes** | Confidence: 0.90 | Value: 45%

Bitwarden Secrets Manager integration is a good addition — hardcoded API keys are a real pain point. A few issues need fixing before merge:

### Must Fix

**1. Test suite fails — `selfImprovement` disable incomplete (HIGH)**

Commit `9b01329` adds `selfImprovement: { enabled: false }` to the first `createMockApi` config, but `sessionDefaultApi` (line ~165) and `sessionEnabledApi` (line ~185) don't have it. The `appendSelfImprovementNote` hook still registers → assertion at line 176 fails: `hooks['command:new'] === undefined`.

Fix: add `selfImprovement: { enabled: false }` to all three config blocks in `plugin-manifest-regression.mjs`.

**2. Async `register()` is a breaking change (HIGH)**

Making `register()` async means all existing synchronous call sites silently become fire-and-forget — hooks and services may not be ready when callers assume they are. This needs to be called out as a breaking change with migration guidance, or wrap the async secret resolution internally so the public API stays synchronous.

### Nice to Have

- **Access token in process list**: `secret-resolver.ts:97` passes `--access-token` as a CLI arg, visible via `ps aux`. Consider using `BWS_ACCESS_TOKEN` env var in the `execFile` env option instead.
- **Duplicate key resolution**: LLM API key is resolved twice via `bws` CLI in a single `register()` call (index.ts:1692 and 2192). Resolve once early and reuse.
- **Template re-expansion**: resolved Bitwarden secrets containing `${...}` patterns get re-interpreted as env var templates by the embedder. Consider marking resolved values as final.

---

Issue #260, opened 2026-03-18
autoCapture + smartExtraction agent_end hook causes Telegram messages to be silently dropped


qingshuicn:
## Bug Description

When `autoCapture: true` and `smartExtraction: true`, the `agent_end` hook locks the session while making LLM calls (for smart extraction) and embedding calls (for dedup/capture). During this time, incoming Telegram messages are silently dropped — no error is logged, and the user receives no response.

## Environment

- **OpenClaw**: v2026.3.13
- **memory-lancedb-pro**: v1.1.0-beta.9
- **Channel**: Telegram (long-polling, `streaming: "partial"`)
- **Embedding**: DashScope `text-embedding-v4`
- **LLM**: DashScope `qwen-plus`

## Steps to Reproduce

1. Enable `autoCapture: true` and `smartExtraction: true` in plugin config
2. Connect via Telegram channel
3. Send a message → bot replies normally (first message)
4. Send follow-up messages → bot processes them (tools.profile loads, memory plugin runs) but **no `sendMessage` is ever logged** — responses are silently lost

## Observed Behavior

- Gateway log shows agent sessions starting (tools.profile loaded) for each incoming message
- `memory-lancedb-pro: regex fallback found 0 capturable texts` appears after each session
- **Zero `sendMessage ok` entries** after the first successful response
- No errors logged anywhere — completely silent failure

## Expected Behavior

All Telegram messages should receive responses regardless of plugin hook activity.

## Diagnosis

The `agent_end` hook calls the LLM for 6-category extraction and embedding API for dedup. With a 30s default timeout, this can lock the session long enough to trigger OpenClaw's session store lock timeout, causing subsequent messages to be silently dropped.

Timeline from logs:
```
19:55:11-19:55:18 — 8x sendMessage ok (first conversation works)
19:55:57 — tools.profile loaded (new message received)
19:56:04 — tools.profile loaded (another message)
19:56:22 — tools.profile loaded (another message)
...
(no sendMessage for any of these — all silently dropped)
```

## Workaround

Setting `autoCapture: false` and `smartExtraction: false` resolves the issue. The plugin still works for manual `memory_recall`/`memory_capture` tool calls.

## Additional Config Issue Found

[...]

PR #391, opened 2026-03-28
fix: default selfImprovement.enabled to true when config block omitted

AliceLJY:
## Summary

- When `selfImprovement` config block is entirely omitted from plugin config, `enabled` defaults to `false`, which prevents all three `self_improvement_*` tools from registering
- This is inconsistent: when the block IS provided but `enabled` is not explicitly set, it correctly defaults to `true` via the `!== false` check
- Fix: change the else-branch default from `enabled: false` to `enabled: true` so the behavior is consistent regardless of whether the config block is present

## Details

In `index.ts` around line 3830, the config parsing for `selfImprovement` has two branches:

1. When `selfImprovement` is an object: `enabled` defaults to `true` (via `!== false`)
2. When `selfImprovement` is omitted: `enabled` was hardcoded to `false` (the bug)

Verified by testing on a live OpenClaw instance: adding `selfImprovement: { enabled: true }` to the config makes all 3 self-improvement tools appear, confirming the tools work — they just weren't registering due to the default.

Fixes #375

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #406, opened 2026-03-29
feat(i18n): German language support + fix fire-and-forget race in agent_end

### LINKED ISSUE #393: Auto-capture silently fails: missing German/i18n triggers + extractMinMessages too high for short conversations
### Plugin Version

1.1.0-beta.10

### OpenClaw Version

2026.3.24

### Bug Description

Auto-capture via the `agent_end` hook silently fails to persist memories for short conversations (1-2 turns) in non-English languages — specifically German. Two interacting root causes:

**1. `MEMORY_TRIGGERS` has no German patterns** (index.ts ~L1242)
"Merke dir: X" (German for "Remember: X") matches no trigger regex, so `shouldCapture()` returns false and the regex fallback drops the text entirely. Same gap in `AUTO_CAPTURE_EXPLICIT_REMEMBER_RE` (only Chinese) and `FORCE_RETRIEVE_PATTERNS` in adaptive-retrieval.ts (only English + Chinese).

**2. `extractMinMessages` default of 4 is too high for CLI/single-turn interactions**
Smart extraction never fires for conversations with <4 eligible texts (e.g. a single `openclaw agent --message="..."` call produces only 1 eligible text). Falls through to regex fallback which then fails due to (1).

**3. `void backgroundRun;` fire-and-forget pattern** (index.ts ~L2888)
While the gateway process is persistent so this works in practice, making the hook async with `await Promise.race([backgroundRun, timeout])` is safe because `runAgentEnd()` is called fire-and-forget from core (`.catch()` only) — session locks are released in the `finally` block regardless of hook completion. This would also make the plugin work correctly in `--local` embedded mode.

### Expected Behavior

"Merke dir: X" should trigger memory capture (equivalent to "remember: X" or "记住: X"). Short conversations (1-2 turns) should also be captured via smart extraction or regex fallback.

### Steps to Reproduce

1. Default config, German-speaking user
2. `openclaw agent --message="Merke dir: Testtoken ist Cobalt Tiger 7742"`
3. Agent responds "Gemerkt."
4. `openclaw memory-pro search "Cobalt Tiger"` → **No matches**
5. Gateway log confirms: `smart-extractor: no memories extracted` → `regex fallback found 0 capturable texts`

**Verified fix:** After adding German patterns to `MEMORY_TRIGGERS` and setting `extractMinMessages: 1`:
- Gateway log: `regex fallback found 1 capturable text(s)` → `auto-captured 1 memories for agent main`
- `openclaw memory-pro search "Emerald Bear"` → Found with 75% score

**Suggested patches (all tested and verified):**

**MEMORY_TRIGGERS** (index.ts ~L1260):
```typescript
// German triggers
/merk dir|merke dir|erinner dich|vergiss nicht|nicht vergessen/i,
/ich bevorzuge|ich mag|ich hasse|ich will|ich brauche/i,
/wir haben entschieden|ab jetzt|ab sofort|in zukunft/i,
/mein\s+\w+\s+ist|heißt|wohne|arbeite/i,
/immer|niemals|wichtig/i,
```

**AUTO_CAPTURE_EXPLICIT_REMEMBER_RE** (index.ts ~L741):
```typescript
// Add English + German alongside existing Chinese
/^(?:请|請)?(?:remember(?:\s+this)?|merke?\s+dir|vergiss\s+(?:das\s+)?nicht|记住|記住|记一下|記一下|别忘了|別忘了)[。.!?？!]*$/iu;
```

**FORCE_RETRIEVE_PATTERNS** (adaptive-retrieval.ts ~L35):
```typescript
// German retrieval triggers
/\b(erinnerst du dich|weißt du noch|was (war|ist) mein|letzte[sn]?\s+mal|vorher|gestern|neulich)\b/i,
/\b(merk dir|merke dir|habe ich dir gesagt|habe ich erw[äa]hnt)\b/i,
```

**agent_end hook** (index.ts ~L2556+2888): Change from sync fire-and-forget to async with timeout:
```typescript
// Type: (event, ctx) => Promise<void> | void  (was: void)
const agentEndAutoCaptureHook: AgentEndAutoCaptureHook = async (event, ctx) => {
  // ... existing logic ...
  const backgroundRun = (async () => { /* ... */ })();
  agentEndAutoCaptureHook.__lastRun = backgroundRun;
  // Safe: runAgentEnd() is fire-and-forget from core, session lock freed in finally
  try {
    await Promise.race([backgroundRun, new Promise<void>(r => setTimeout(r, 15_000))]);
  } catch { /* errors logged inside backgroundRun */ }
};
```

**Config recommendation:** Consider lowering `extractMinMessages` default from 4 to 2.

### Error Logs / Screenshots

```shell
Gateway log after store attempt (default config, German user):

smart-extractor: no memories extracted
memory-lancedb-pro: smart extraction produced no persisted memories for agent main (created=0, merged=0, skipped=0); falling back to regex capture
memory-lancedb-pro: regex fallback found 0 capturable texts for agent main


After applying German MEMORY_TRIGGERS patch + extractMinMessages=1:

memory-lancedb-pro: regex fallback found 1 capturable text(s) for agent main
memory-lancedb-pro: auto-captured 1 memories for agent main in scope agent:main
```

### Embedding Provider

None

### OS / Platform

Debian 13 (LXC on Proxmox), Node.js 22

Banger455:
## Summary

Fixes #393 — Memory store reports `STORE-OK` but recall in a new session returns `RECALLED: NONE` for German-speaking users.

### Root cause

Three separate bugs compound into a silent data-loss scenario:

1. **Missing German trigger patterns** — `MEMORY_TRIGGERS`, `AUTO_CAPTURE_EXPLICIT_REMEMBER_RE`, and `FORCE_RETRIEVE_PATTERNS` cover English, Czech, and Chinese but have no German entries. Phrases like *"Merke dir: X"* fall through `shouldCapture()` → nothing is stored.
2. **Fire-and-forget `void backgroundRun`** in the `agent_end` hook — works in gateway mode (persistent process) but races against `process.exit()` in `--local` / embedded CLI mode. The capture Promise gets killed before it flushes to LanceDB.

#### Commit 1 — `feat(i18n): add German language triggers`

| Array | File | Added patterns |
|-------|------|---------------|
| `MEMORY_TRIGGERS` | `index.ts` | 5 regexes: "merke dir", "vergiss nicht", "ich bevorzuge", "wir haben entschieden", "ab jetzt/sofort/in Zukunft", personal info |
| `AUTO_CAPTURE_EXPLICIT_REMEMBER_RE` | `index.ts` | German + English explicit remember commands |
| `FORCE_RETRIEVE_PATTERNS` | `src/adaptive-retrieval.ts` | 2 regexes: "erinnerst du dich", "weißt du noch", "gestern", "habe ich erwähnt", etc. |

#### Commit 2 — `fix(auto-capture): await backgroundRun with timeout`

- Hook becomes `async`; `void backgroundRun` replaced by `await Promise.race([backgroundRun, 15s timeout])`
- `AgentEndAutoCaptureHook` type widened to `Promise<void> | void`
- **Session-lock / channel-delivery safety (Issue #260):** OpenClaw core calls `runAgentEnd()` fire-and-forget with `.catch()`, so the new `await` does **not** block session locks or downstream deliveries in gateway mode. The timeout is a safety net for hung API calls only.

### Note on `extractMinMessages`

The default value of `extractMinMessages` (currently 4) means smart extraction never fires for CLI one-shot commands (only 1 eligible text). We worked around this with a config override (`extractMinMessages: 1`), but the default might be worth revisiting upstream — leaving that decision to the maintainers.

### Verification (on live OpenClaw instance)

| Test | Result |
|------|--------|
| Gateway log shows auto-capture | ✅ `regex fallback found 1 capturable text(s)` → `auto-captured 1 memories` |
| `openclaw memory-pro search "Emerald Bear"` | ✅ Found with 75% score |
| Recall in fresh session: "Welche Testtokens kennst du?" | ✅ All 6 test tokens correctly recalled |
| Memory stats: 0 garbage | ✅ 32 clean memories |

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -744,7 +744,7 @@ function shouldSkipReflectionMessage(role: string, text: string): boolean {
 
 const AUTO_CAPTURE_MAP_MAX_ENTRIES = 2000;
 const AUTO_CAPTURE_EXPLICIT_REMEMBER_RE =
-  /^(?:请|請)?(?:remember(?:\s+this)?|merke?\s+dir|vergiss\s+(?:das\s+)?nicht|记住|記住|记一下|記一下|别忘了|別忘了)[。.!?？!]*$/iu;
+  /^(?:请|請)?(?:remember(?:\s+this)?|merk(?:e?\s+dir|\s+es\s+dir)|vergiss\s+(?:das\s+)?nicht|nicht\s+vergessen|记住|記住|记一下|記一下|别忘了|別忘了)[。.!?？!]*$/iu;
 
 /**
  * Prune a Map to stay within the given maximum number of entries.
@@ -1265,11 +1265,11 @@ const MEMORY_TRIGGERS = [
   /重要|關鍵|关键|注意|千萬別|千万别/,
   /幫我|筆記|存檔|存起來|存一下|重點|原則|底線/,
   // German triggers
-  /merk dir|merke dir|merk es dir|erinner dich|vergiss nicht|nicht vergessen/i,
+  /merk(?:e?\s+dir|\s+es\s+dir)|erinner(?:e)?\s+dich|vergiss\s+(?:das\s+)?nicht|nicht\s+vergessen/i,
   /ich bevorzuge|ich mag|ich hasse|ich will|ich brauche/i,
   /wir haben entschieden|wir nutzen|ab jetzt|ab sofort|in zukunft/i,
-  /mein\s+\w+\s+ist|heißt|wohne|arbeite/i,
-  /immer|niemals|wichtig/i,
+  /mein\s+\w+\s+(?:ist|heißt)|ich\s+(?:wohne|arbeite)\b/i,
+  /\b(immer|niemals|wichtig)\b/i,
 ];
 
 const CAPTURE_EXCLUDE_PATTERNS = [
@@ -2904,13 +2904,18 @@ const memoryLanceDBProPlugin = {
         // fire-and-forget (.catch() only), so this await does NOT block
         // session locks or channel deliveries (see Issue #260).
         // The 15s timeout is a safety net for hung API calls.
+        let safetyTimer: ReturnType<typeof setTimeout> | undefined;
         try {
           await Promise.race([
             backgroundRun,
-            new Promise<void>(resolve => setTimeout(resolve, 15_000)),
+            new Promise<void>(resolve => {
+              safetyTimer = setTimeout(resolve, 15_000);
+            }),
           ]);
         } catch {
           // Errors already logged inside backgroundRun
+        } finally {
+          if (safetyTimer !== undefined) clearTimeout(safetyTimer);
         }
       };
 
--- a/test/agent-end-async-capture.test.mjs
+++ b/test/agent-end-async-capture.test.mjs
@@ -1,0 +1,186 @@
+import { describe, it } from "node:test";
+import assert from "node:assert/strict";
+import { fileURLToPath } from "node:url";
+import path from "node:path";
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import Module from "node:module";
+import jitiFactory from "jiti";
+
+// Ensure openclaw module resolution works in CI environments
+process.env.NODE_PATH = [
+  process.env.NODE_PATH,
+  "/opt/homebrew/lib/node_modules/openclaw/node_modules",
+  "/opt/homebrew/lib/node_modules",
+].filter(Boolean).join(":");
+Module._initPaths();
+
+const jiti = jitiFactory(import.meta.url, { interopDefault: true });
+const plugin = jiti("../index.ts");
+
+function createMockApi(pluginConfig) {
+  return {
+    pluginConfig,
+    hooks: {},
+    toolFactories: {},
+    logger: {
+      info() {},
+      warn() {},
+      error() {},
+      debug() {},
+    },
+    resolvePath(value) { return value; },
+    registerTool(toolOrFactory, meta) {
+      this.toolFactories[meta.name] =
+        typeof toolOrFactory === "function" ? toolOrFactory : () => toolOrFactory;
+    },
+    registerCli() {},
+    registerService() {},
+    on(name, handler) { this.hooks[name] = handler; },
+    registerHook(name, handler) { this.hooks[name] = handler; },
+  };
+}
+
+const workDir = mkdtempSync(path.join(tmpdir(), "agent-end-async-test-"));
+
+try {
+  describe("agent_end hook — async capture behavior", () => {
+
+    // -----------------------------------------------------------------------
+    // Test 1: hook returns a Promise (is async)
+    // -----------------------------------------------------------------------
+    it("agent_end hook returns a Promise (async function)", () => {
+      const api = createMockApi({
+        dbPath: path.join(workDir, "db-async-check"),
+        autoRecall: false,
+        embedding: {
+          provider: "openai-compatible",
+          apiKey: "dummy",
+          model: "text-embedding-3-small",
+          baseURL: "http://127.0.0.1:9/v1",
+          dimensions: 1536,
+        },
+      });
+      plugin.register(api);
+
+      const hook = api.hooks.agent_end;
+      assert.ok(hook, "agent_end hook should be registered");
+
+      // Call with a non-successful event to hit early return path
+      const result = hook({ success: false, messages: [] }, {});
+      // Even early return from an async function yields a Promise
+      assert.ok(
+        result === undefined || result instanceof Promise,
+        "hook should return a Promise or undefined (async function signature)",
+      );
+    });
+
+    // -----------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


Banger455:
Thanks for the thorough review @rwmjhb! All points addressed in `e773153`:

### Regex fixes (MUST FIX) ✅

1. **Word boundaries** — `/immer|niemals|wichtig/` now uses `\b(immer|niemals|wichtig)\b` to prevent false positives on `Zimmermann`, `Schwimmerin`, `Flimmern`, `Wichtigkeit`, etc.

2. **Ungrouped alternation** — rewritten to `mein\s+\w+\s+(?:ist|heißt)|ich\s+(?:wohne|arbeite)\b` so `heißt`/`wohne`/`arbeite` require proper context. "Die Stadt heißt Berlin" no longer triggers.

3. **Explicit-remember consistency** — Both `AUTO_CAPTURE_EXPLICIT_REMEMBER_RE` and `MEMORY_TRIGGERS` now cover the same German variants: `merk dir`, `merke dir`, `merk es dir`, `vergiss nicht`, `vergiss das nicht`, `nicht vergessen`, `erinnere dich`, `erinner dich`.

### Timer cleanup (NICE TO HAVE) ✅

`setTimeout` handle is now stored and `clearTimeout`'d in a `finally` block so the 15s safety timer doesn't keep Node alive after `backgroundRun` resolves.

### Test coverage (NICE TO HAVE) ✅

Two new test files, **58 test cases** total:

- **`test/german-i18n-triggers.test.mjs`** (54 cases)
  - `shouldCapture` positive: 24 cases (remember commands, preferences, decisions, personal info, emphasis)
  - `shouldCapture` false-positive prevention: 9 cases (compound words, missing context prefix, length bounds)
  - `shouldSkipRetrieval` force-retrieve: 15 cases (German retrieval triggers + short-command skip verification)
  - Explicit remember consistency: 6 cases (all German variants padded above length threshold)

- **`test/agent-end-async-capture.test.mjs`** (4 cases)
  - Hook returns Promise (async signature verification)
  - Error swallowing (unreachable endpoint doesn't throw)
  - `__lastRun` stores a Promise for downstream consumers
  - Early return for empty events (no 15s timer leak)

### Build failure note

The `plugin-manifest-regression.mjs` L155 failure (`command:new === undefined`) is a pre-existing upstream mismatch: `26abb04` changed `selfImprovement.enabled` to default `true`, which registers on `command:new`, but the test wasn't updated. This is unrelated to our diff — happy to fix it in a separate PR if helpful.

### Rebase

Branch rebased onto latest `master` (`7fe2ae0`).

app3apps:

Thanks for the follow-up and the additional test coverage. At this point I don’t want to keep this open indefinitely, especially since it combines two fairly different changesets: German trigger expansion and the `agent_end` async/timeout behavior change.

If you want to revive this later, please rebase onto current `master` and open a fresh PR (or reopen if the branch is still valid), ideally with the language-trigger changes and the lifecycle-hook behavior change split if possible. That will make the re-review much easier.

Banger455:
Hey @AliceLJY, thanks for the detailed review — sorry for the delayed response!

### 1. async hook safety

Confirmed safe. The OpenClaw core `runAgentEnd()` call path already wraps hook execution in fire-and-forget with `.catch()` (see Issue #260 discussion):

- **Gateway mode**: Core calls `runAgentEnd()` with `.catch()`, so the returned Promise resolves in the background — no change in behavior. Session locks and channel delivery are unaffected.
- **`--local` / embedded CLI mode**: This is where the `await` actually matters. Without it, `process.exit()` kills the capture Promise before LanceDB flush completes. The 15 s timeout is a safety net for hung API calls only.
- **Unhandled rejections**: Our hook wraps `backgroundRun` in try/catch and swallows errors (verified in `agent-end-async-capture.test.mjs` — the "error swallowing" test case). Core's `.catch()` is a second safety layer. Even if the hook throws, nothing propagates as an unhandled rejection.

### 2. `immer` trigger precision

You're right — standalone `immer` is too noisy. I'll scope it to explicit memory-intent contexts in the follow-up PRs:

```js
/\b(immer\s+(?:wenn|daran|denken|merken|beachten)|niemals|wichtig)\b/i
```

This way "Das ist immer so" won't trigger, but "Denk immer daran" or "Immer wenn du X machst" will. The primary German triggers (`merke dir`, `vergiss nicht`) remain the workhorses.

---

### Regarding the PR split (@app3apps, @rwmjhb)

Understood and agreed — combining two different changesets made this harder to review. I'll rebase onto current `master` and open **two fresh PRs**:

1. **PR A — `feat(i18n): German language triggers`**: `MEMORY_TRIGGERS`, `AUTO_CAPTURE_EXPLICIT_REMEMBER_RE`, `FORCE_RETRIEVE_PATTERNS` + test coverage (`german-i18n-triggers.test.mjs`) — with the `immer` precision fix from above baked in.
2. **PR B — `fix(auto-capture): await backgroundRun with timeout`**: async hook, `Promise.race`, `clearTimeout` cleanup + test coverage (`agent-end-async-capture.test.mjs`)

Both will be rebased on current `master` (which now includes the `command:new` test fix from #391 — thanks @AliceLJY!). Will open them shortly.

PR #414, opened 2026-03-30
fix: resolve env vars in retrieval rerank config

ChaoYang78:
## Summary
- resolve env placeholders in `retrieval.rerankApiKey` before building rerank requests
- also resolve `retrieval.rerankEndpoint`, `retrieval.rerankModel`, and `retrieval.rerankProvider` for consistency with env-based config
- add a regression test covering env expansion in retrieval rerank config

## Why
`embedding.apiKey` already supports `${ENV_VAR}` expansion, but `retrieval.rerankApiKey` was passed through without explicit resolution in `parsePluginConfig`. In env-based OpenClaw configs this can lead to rerank auth failures or literal `${JINA_API_KEY}` values being propagated.

## Validation
- patched running OpenClaw instance and verified memory pipeline remained healthy
- `node --test test/resolve-env-vars-array.test.mjs`

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -3808,7 +3808,25 @@ export function parsePluginConfig(value: unknown): PluginConfig {
     autoRecallPerItemMaxChars: parsePositiveInt(cfg.autoRecallPerItemMaxChars) ?? 180,
     maxRecallPerTurn: parsePositiveInt(cfg.maxRecallPerTurn) ?? 10,
     captureAssistant: cfg.captureAssistant === true,
-    retrieval: typeof cfg.retrieval === "object" && cfg.retrieval !== null ? cfg.retrieval as any : undefined,
+    retrieval:
+      typeof cfg.retrieval === "object" && cfg.retrieval !== null
+        ? (() => {
+          const retrieval = { ...(cfg.retrieval as Record<string, unknown>) } as Record<string, unknown>;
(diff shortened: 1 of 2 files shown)


rwmjhb:
### Review: APPROVE (rebase needed)

The fix is correct — `parsePluginConfig` was passing retrieval config fields through without resolving `${...}` env placeholders, causing rerank auth failures in env-based deployments. The dedicated regression test confirms the fix works.

**Before merge, please rebase onto `main`** — the `plugin-manifest-regression` test failure (`sessionMemory should stay disabled by default`) is unrelated to your change and comes from base branch divergence.

**Two things worth considering** (not blocking):

1. **Eager resolution when reranking is disabled** — the new code resolves `rerankApiKey`, `rerankEndpoint`, etc. unconditionally. If `rerank: "none"` but config still has `rerankApiKey: "${RERANK_KEY}"` and that env var is unset, `resolveEnvVars` will throw at parse time even though reranking is off. Consider gating resolution on `retrieval.rerank === "cross-encoder"`.

2. **No validation for `rerankProvider`** — a typo in the env var (e.g. `RERANK_PROVIDER=jiina`) silently falls through to Jina semantics. Pre-existing issue, but now slightly more reachable via env-var indirection.

Issue #405, opened 2026-03-29
[BUG] selfImprovement hooks registered without config block


jlin53882:
### Environment

- **Plugin version**: `memory-lancedb-pro@1.1.0-beta.10` (master branch)
- **Platform**: CI / GitHub Actions

### Error (from `plugin-manifest-regression.test.mjs`)

```
AssertionError [ERR_ASSERTION]: sessionMemory should stay disabled by default
+ actual - expected
+ [AsyncFunction: appendSelfImprovementNote]
- undefined
```

### Root Cause

When no `selfImprovement` config block is provided, the `command:new` and `command:reset` hooks are registered unconditionally.

In `index.ts`, the guard for `agent:bootstrap` hook correctly checks:

```typescript
if (config.selfImprovement?.enabled !== false) {
  api.registerHook("agent:bootstrap", ...);
}
```

But the `command:new`/`command:reset` hooks only check `beforeResetNote`:

```typescript
if (config.selfImprovement?.beforeResetNote !== false) {  // ← missing enabled check!
  const appendSelfImprovementNote = async (event) => { ... };
  api.registerHook("command:new", appendSelfImprovementNote, {...});
  api.registerHook("command:reset", appendSelfImprovementNote, {...});
}
```

When `config.selfImprovement` is `undefined`:
- `undefined !== false` → `true`
- So hooks are registered even though `selfImprovement` is not configured at all

### Impact

All plugin instances without explicit `selfImprovement` config block have `command:new`/`command:reset` hooks incorrectly registered.

### Related

- Related PR (fix for a different bug): #404
- This is a separate regression bug in the `master` branch

AliceLJY:
Thanks for the report. This looks like a default-behavior regression — `selfImprovement` hooks should only register when explicitly enabled in config.

Could you confirm: is this reproducible on a clean install with no `selfImprovement` config block, or does it only happen in CI with a specific test fixture? The test name (`plugin-manifest-regression.test.mjs`) suggests this might be a test environment issue.

If you can share the minimal config that triggers the unwanted registration, that would help narrow it down.

jlin53882:
## Confirmed: upstream regression

Verified by inspecting the local `extensions/memory-lancedb-pro/index.ts` (installed as `memory-lancedb-pro@1.1.0-beta.10`). The guard on line 2825:

```ts
if (config.selfImprovement?.beforeResetNote !== false) {
  api.registerHook("command:new", ...);
[...]

PR #418, opened 2026-03-30
fix: selfImprovement hooks only register when explicitly enabled

jlin53882:
### Root Cause

When `config.selfImprovement` is absent (undefined), the `command:new` and `command:reset` hooks are registered unconditionally because the guard only checks `beforeResetNote !== false`:

```ts
// BEFORE (buggy)
if (config.selfImprovement?.beforeResetNote !== false) {
  api.registerHook("command:new", ...);
  api.registerHook("command:reset", ...);
}
```

The `agent:bootstrap` hook correctly guards with `enabled !== false`, but the other two hooks do not. This causes unwanted hook registration in all plugin instances that don't explicitly configure `selfImprovement`.

### Fix

Add `enabled !== false` to the guard, consistent with the `agent:bootstrap` pattern:

```ts
// AFTER (fixed)
if (config.selfImprovement?.enabled !== false && config.selfImprovement?.beforeResetNote !== false) {
  api.registerHook("command:new", ...);
  api.registerHook("command:reset", ...);
}
```

### Verification

Confirmed locally by inspecting `index.ts` in the extensions install (`memory-lancedb-pro@1.1.0-beta.10`). Both `extensions/memory-lancedb-pro/index.ts` and the upstream fork have the same guard — the regression is upstream, not local.

### References

- Fixes [CortexReach/memory-lancedb-pro#405](https://github.com/CortexReach/memory-lancedb-pro/issues/405)

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -2938,7 +2938,7 @@ const memoryLanceDBProPlugin = {
         description: "Inject self-improvement reminder on agent bootstrap",
       });
 
-      if (config.selfImprovement?.beforeResetNote !== false) {
+      if (config.selfImprovement?.enabled !== false && config.selfImprovement?.beforeResetNote !== false) {
         const appendSelfImprovementNote = async (event: any) => {
           try {
             const action = String(event?.action || "unknown");


rwmjhb:
### Review: CLOSE

The bug you're describing is real in principle, but this PR's code change is a no-op on the current base.

`parseConfig` at `index.ts:3822-3834` already normalizes a missing `selfImprovement` block to `{ enabled: true, beforeResetNote: true, ... }`. By the time execution reaches your new guard at `index.ts:2941`, `config.selfImprovement.enabled` is always `true` — the check never changes behavior.

This was intentional: commit `26abb04` (`fix: default selfImprovement.enabled to true when config block omitted (#391)`) explicitly shipped default-on semantics, and `openclaw.plugin.json:548-556` declares `selfImprovement.enabled` with default `true`.

The `plugin-manifest-regression.mjs:155` test failure (`command:new` should be `undefined`) is actually testing that **sessionMemory** stays disabled by default, but `command:new` is registered by the **selfImprovement** block — these are separate features. The regression test itself needs re-examination.

If the intent is to make selfImprovement opt-in instead of opt-out, that requires changing `parseConfig`'s default from `enabled: true` to `enabled: false`, which would revert #391 and is a product decision, not a bug fix.

rwmjhb:
Closing: the code change is behaviorally redundant after parseConfig normalization. See review comment above for details.

PR #433, opened 2026-03-31
fix: remove duplicate memory_compact tool registration

AliceLJY:
## Summary

- Remove orphaned inline `memory_compact` tool registration in `index.ts` (leftover prototype from before PR #343)
- The canonical implementation in `src/tools.ts` via `registerMemoryCompactTool()` is kept as-is
- Auto-compaction at `gateway_start` hook (which calls `runCompaction` directly) is unaffected

**Root cause**: `memory_compact` was registered twice — once in `tools.ts` (proper version) and once in `index.ts` (old prototype). OpenClaw's tool-name conflict detection logged ERRORs on every gateway start/reload.

**Fix**: Delete the 88-line prototype block in `index.ts`. One registration, zero conflicts.

Fixes #431

## Test plan

- [ ] `openclaw status` no longer shows `plugin tool name conflict (memory-lancedb-pro): memory_compact` ERROR
- [ ] `memory_compact` tool still works via the `tools.ts` implementation
- [ ] Auto-compaction at gateway start still triggers when `memoryCompaction.enabled: true`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #441, opened 2026-04-01
fix: rename deduplication tool to memory_deduplicate to avoid collision

### LINKED ISSUE #431: plugin tool name conflict (memory-lancedb-pro): memory_compact — 工具重复注册导致 ERROR
### Plugin Version

1.1.0-beta.10

### OpenClaw Version

2026.3.24

### Bug Description

每次网关启动后，日志中持续输出 ERROR：
`plugin tool name conflict (memory-lancedb-pro): memory_compact`

3月30日出现 5 次，3月31日出现 54 次（配置重载后重新注册工具时触发）。
根因：memory_compact 工具被注册了两次：
registerAllMemoryTools() 内部调用 registerMemoryCompactTool() 注册了一次
index.ts 第 2012 行又通过 api.registerTool({ name: "memory_compact", ... }) 再次注册
registerAllMemoryTools 位于 src/tools.ts 第 1907 行，调用后在第 1989 行；紧接着第 2012 行又注册了一次同名工具，触发 OpenClaw 的工具名冲突检测。

### Expected Behavior

memory_compact 只注册一次，不产生 ERROR 日志。

### Steps to Reproduce

1.安装 memory-lancedb-pro v1.1.0-beta.10
2.在 openclaw.json 中启用插件（plugins.entries.memory-lancedb-pro.enabled: true）
3.启动或重载网关
4.查看日志，发现 plugin tool name conflict ERROR

### Error Logs / Screenshots

```shell
plugin tool name conflict (memory-lancedb-pro): memory_compact
来源文件：src/tools.ts 中 registerAllMemoryTools 已注册 memory_compact，index.ts 第 2012 行重复注册
```

### Embedding Provider

None

### OS / Platform

Windows 11 + WSL2（Ubuntu24.04）

Displayer226:
## Summary
This PR resolves the naming conflict for the `memory_compact` tool by renaming the metadata-based deduplication version to `memory_deduplicate`.

## Motivation
Fixes #431. 
This is offered as an **alternative** to PR #433. 

While #433 proposes removing one of the implementations, this PR argues that both are essential for a complete memory lifecycle:
1. **Deduplication (`memory_deduplicate`)**: A lightweight, metadata-based tool in `src/tools.ts` used for structural hygiene (archiving identical or redundant facts).
2. **Summarization (`memory_compact`)**: A heavyweight, LLM-based tool in `index.ts` used for increasing knowledge density by merging related memories into refined entries.

By renaming instead of removing, we preserve both functionalities for the user.

## Key Changes
- Renamed `registerMemoryCompactTool` to `registerMemoryDeduplicateTool` in `src/tools.ts`.
- Changed the exposed tool name from `memory_compact` to `memory_deduplicate`.
- Updated `registerAllMemoryTools` to reflect the new name.

## Testing
- Verified that both tools can now be registered simultaneously without gateway startup errors.
- Manual test of `memory_deduplicate` confirmed it correctly identifies and archives duplicate metadata entries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 
-## 1.1.0-beta.2 (Smart Memory Beta + Access Reinforcement)
+## 1.1.0-beta.11 (Tool Naming & Parameter Consistency)
+
+- **Breaking Change**: Renamed the original `memory_compact` tool (deduplication) to `memory_deduplicate` to distinguish it from progressive summarization.
+- **Breaking Change**: Standardized `memory_compact` (progressive summarization) and `memory_deduplicate` parameters to use **camelCase** (`dryRun`, `minAgeDays`, `similarityThreshold`).
+- **Feat**: Both `memory_compact` and `memory_deduplicate` now default to `dryRun: true` (preview mode) for safety.
+- **Fix**: Updated governance tests to reflect the rename and parameter changes.
+
+## 1.1.0-beta.10 (Smart Memory Beta + Access Reinforcement)
 
 This is a **beta** release published under the npm dist-tag **`beta`** (it does not affect the stable `latest` channel).
 
--- a/index.ts
+++ b/index.ts
@@ -2088,19 +2088,19 @@ const memoryLanceDBProPlugin = {
         description:
           "Consolidate semantically similar old memories into refined single entries " +
           "(progressive summarization). Reduces noise and improves retrieval quality over time. " +
-          "Use dry_run:true first to preview the compaction plan without making changes.",
+          "Use dryRun:true (default) first to preview the compaction plan without making changes.",
         inputSchema: {
           type: "object" as const,
           properties: {
-            dry_run: {
+            dryRun: {
               type: "boolean",
-              description: "Preview clusters without writing changes. Default: false.",
+              description: "Preview clusters without writing changes. Default: true.",
             },
-            min_age_days: {
+            minAgeDays: {
               type: "number",
               description: "Only compact memories at least this many days old. Default: 7.",
             },
-            similarity_threshold: {
+            similarityThreshold: {
               type: "number",
               description: "Cosine similarity threshold for clustering [0-1]. Default: 0.88.",
             },
@@ -2116,16 +2116,16 @@ const memoryLanceDBProPlugin = {
           const compactionCfg: CompactionConfig = {
             enabled: true,
             minAgeDays:
-              typeof args.min_age_days === "number"
-                ? args.min_age_days
+              typeof args.minAgeDays === "number"
+                ? args.minAgeDays
                 : (config.memoryCompaction?.minAgeDays ?? 7),
             similarityThreshold:
-              typeof args.similarity_threshold === "number"
-                ? Math.max(0, Math.min(1, args.similarity_threshold))
+              typeof args.similarityThreshold === "number"
+                ? Math.max(0, Math.min(1, args.similarityThreshold))
                 : (config.memoryCompaction?.similarityThreshold ?? 0.88),
             minClusterSize: config.memoryCompaction?.minClusterSize ?? 2,
             maxMemoriesToScan: config.memoryCompaction?.maxMemoriesToScan ?? 200,
-            dryRun: args.dry_run === true,
+            dryRun: args.dryRun !== false,
             cooldownHours: config.memoryCompaction?.cooldownHours ?? 24,
           };
           const scopes =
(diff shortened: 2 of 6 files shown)


rwmjhb:
### Review: REQUEST-CHANGES

The rename approach is the right call — both tools serve different purposes (dedup vs progressive compaction), so they should coexist under distinct names. Closing #433 in favor of this direction.

**Must fix:**

1. **Governance test is broken** — `test/memory-governance-tools.test.mjs:157` still asserts `memory_compact` is registered. After the rename, this test fails with `AssertionError: tool memory_compact should be registered`. This file is also not in the default `npm test` script, so CI won't catch it.

2. **Parameter style mismatch** — the renamed `memory_deduplicate` uses camelCase `dryRun` (default: preview), while the surviving `memory_compact` uses snake_case `dry_run` (default: write). A caller using `{ dryRun: true }` against `memory_compact` silently gets write behavior instead of preview.

**Worth considering:**

- The rename is a breaking change for any existing user or LLM agent calling `memory_compact` expecting dedup behavior — consider documenting the migration in the PR description or a changelog note.
- Branch is behind `main` — please rebase.

Displayer226:
Hello !

Thanks for the review! I've updated the PR to address all the 'Must fix' items and rebased on the latest master:

   - Parameter Standardization: Standardized memory_compact parameters to camelCase (dryRun, minAgeDays, similarityThreshold) and
     set dryRun: true as the default for safety, matching the memory_deduplicate style.
   - Fixed Governance Test: Updated test/memory-governance-tools.test.mjs to target the correct tool name.
   - Changelog: Added a note in CHANGELOG.md regarding the breaking rename and parameter changes for version 1.1.0-beta.11.
   - Rebase & CI Cleanup: Rebased on master and fixed two existing test regressions in plugin-manifest-regression and
     session-summary-before-reset caused by recent changes to selfImprovement hook defaults.
   - Extra Verification: Added a new check in test/plugin-manifest-regression.mjs to ensure both tools remain correctly registered with their new camelCase interface and safe defaults going forward.

The full test suite is now passing locally (npm test).

AliceLJY (approved):
LGTM — clean rename + parameter standardization, addresses all prior review feedback.

**Verified:**
- `memory_compact` (progressive summarization) and `memory_deduplicate` (metadata dedup) now coexist with distinct names
- Parameters consistently use camelCase across both tools
- `dryRun` defaults to `true` (safe preview mode) for both tools
- Governance test updated to target `memory_deduplicate`
- Plugin manifest regression test extended with assertions for both tools' parameter schemas
- CHANGELOG clearly documents the breaking changes
- `selfImprovement: { enabled: false }` additions fix unrelated test regressions from upstream

**Minor note (non-blocking):** In `plugin-manifest-regression.mjs`, the updated `registerTool` mock silently drops calls where `toolOrFactory` is a function but `meta` is undefined (neither `if` branch matches). Original code would have thrown on `meta.name`. Since this is test-only, low risk, but worth noting in case a future tool registration path hits that gap.

Assigning to @rwmjhb for final merge decision.

app3apps:
Current status from my side: I still think this PR is logically fine, but it is not mergeable as-is because `master` has moved and there is now a real conflict in `test/plugin-manifest-regression.mjs`.

I checked the merge locally. This looks like a test-only conflict, not a new product/code regression.

The resolution should keep both parts of intent:
- keep the new `memory_compact` / `memory_deduplicate` registration + camelCase schema assertions from this PR
- keep this PR's explicit `selfImprovement: { enabled: false }` setup in the affected test harnesses

With that setup, the `command:new` assertions in this file should stay aligned with `selfImprovement` being disabled in the test config, rather than copying over the newer default-on assertion from `master` unchanged.

I also verified locally that after resolving the conflict that way, `node test/plugin-manifest-regression.mjs` passes.

So my recommendation is: please rebase, resolve `test/plugin-manifest-regression.mjs` with that shape, and then this should be ready to merge.

rwmjhb:
## Review: rename deduplication tool to memory_deduplicate to avoid collision

Tool name collision causing ERROR on every gateway start is a real problem worth fixing.

### Must Fix

**1. No test validates the rename or collision fix**

The PR changes the tool name but adds no test asserting the new name works or that the collision is resolved. A minimal test: register both tools, verify no error.

**2. Rebase needed — build is red (stale base)**

`plugin-manifest-regression.mjs` failure is pre-existing.

### Worth discussing

- **Breaking change**: renaming `memory_compact` to `memory_deduplicate` breaks any existing consumer (LLM agents, scripts) that references the old name. Consider: is the alternative approach in PR #433 (remove the duplicate registration instead of renaming) a better fix? That preserves the existing tool name.

PR #442, opened 2026-04-01
feat: expose autoRecallTimeoutMs in plugin configuration schema

Displayer226:
## Summary
Exposes `autoRecallTimeoutMs` in the plugin configuration schema.

## Motivation
Users running local embedding or reranking services (e.g., Jina, TEI, or local Ollama instances) often encounter "auto-recall timed out" warnings during agent startup when their hardware has higher latency. Currently, this value is hardcoded or difficult to adjust via the standard config schema.

This change allows users to increase the timeout to match their specific infrastructure needs, ensuring stable memory injection.

## Key Changes
- Added `autoRecallTimeoutMs` to the properties in `openclaw.plugin.json`.
- Set a safe minimum of `1ms` and a default of `5000ms` to maintain backward compatibility while allowing flexibility.

## Testing
- Verified `openclaw config validate` passes with the new key.
- Confirmed that setting `autoRecallTimeoutMs: 15000` in a live environment successfully prevents timeouts on slower local inference setups.

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -118,7 +118,7 @@
       },
       "autoRecallTimeoutMs": {
         "type": "integer",
-        "minimum": 1,
+        "minimum": 500,
         "maximum": 60000,
         "default": 5000,
         "description": "Timeout for the entire auto-recall pipeline (embedding + search + rerank) in milliseconds."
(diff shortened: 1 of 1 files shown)


Displayer226:
Hello ! 

Thanks for the review, just rebased and updated it to 500ms !

PR #444, opened 2026-04-01
fix: filter subagent runtime wrappers from auto-capture

### LINKED ISSUE #443: auto-capture should strip subagent runtime wrappers before smart extraction
## Summary

`auto-capture` can store subagent runtime wrappers such as `[Subagent Context]` and `[Subagent Task]` as long-term memories instead of filtering them out before smart extraction.

## Impact

This pollutes both LanceDB and the md mirror with execution scaffolding that should never be treated as user memory.

Observed side effects:
- noisy recall results
- lower-quality smart extraction artifacts
- wrapper text persisted as `source=auto-capture`
- lightweight extraction models are more likely to store the wrapper verbatim

## Reproduction

A subagent was given a wrapper-prefixed task similar to:

```text
[Subagent Context] You are running as a subagent (depth 1/1)...
[Subagent Task] Reply with a brief acknowledgment only. Facts for automatic memory extraction quality test: ...
```

After auto-capture ran, one of the resulting memories was effectively the wrapper text itself, including `[Subagent Context]` / `[Subagent Task]`, instead of only storing the actual user facts.

## Expected behavior

Runtime orchestration wrappers should be treated as execution metadata and removed before smart extraction. They should never be persisted as memories.

## Root cause

There are two filtering gaps:

1. `normalizeAutoCaptureText()` / `stripAutoCaptureInjectedPrefix()` do not strip subagent runtime wrappers.
2. `stripEnvelopeMetadata()` strips channel envelope metadata, but does not strip subagent/task runtime wrapper lines.

So wrapper text can survive all the way into the extraction prompt.

## Proposed fix

Add a double guard:

1. **Auto-capture ingress filtering**
   - drop messages that are runtime wrapper payloads such as `[Subagent Context]...` / `[Subagent Task]...`

2. **Extractor-side fallback filtering**
   - strip runtime wrapper lines in `stripEnvelopeMetadata()` as a second defense

3. **Prompt hardening**
   - explicitly tell the extraction model not to store runtime scaffolding / orchestration wrappers as memories

4. **Regression coverage**
   - add tests proving wrapper lines are removed from extractor input
   - add an auto-capture regression showing wrapper text does not reach the extracted memory

## Notes

This looks like a preprocessing bug first, not primarily a model-selection problem. A stronger model may fail less often, but the root issue is that orchestration metadata is being passed downstream as if it were conversation content.

slj130:
## Summary

This PR prevents smart extraction from storing subagent runtime scaffolding such as `[Subagent Context]` and `[Subagent Task]` as long-term memories.

Closes #443.

## What changed

- filter wrapper-only subagent payloads during auto-capture ingress
- strip runtime wrapper lines in `stripEnvelopeMetadata()` as a second defense
- harden the extraction prompt so runtime scaffolding is explicitly excluded
- add regression coverage for both envelope stripping and the auto-capture path

## Why this approach

This is primarily a preprocessing bug, not just a model-quality issue.
A stronger model might fail less often, but wrapper metadata should never reach the extraction stage in the first place.

## Tests

- `node --test test/strip-envelope-metadata.test.mjs`
- `node test/smart-extractor-branches.mjs`

### DIFF
--- a/src/auto-capture-cleanup.ts
+++ b/src/auto-capture-cleanup.ts
@@ -11,6 +11,10 @@ const AUTO_CAPTURE_SESSION_RESET_PREFIX =
   "A new session was started via /new or /reset. Execute your Session Startup sequence now";
 const AUTO_CAPTURE_ADDRESSING_PREFIX_RE = /^(?:<@!?[0-9]+>|@[A-Za-z0-9_.-]+)\s*/;
 const AUTO_CAPTURE_SYSTEM_EVENT_LINE_RE = /^System:\s*\[[^\n]*?\]\s*Exec\s+(?:completed|failed|started)\b.*$/gim;
+const AUTO_CAPTURE_RUNTIME_WRAPPER_LINE_RE = /^\[(?:Subagent Context|Subagent Task)\]\s*/i;
+const AUTO_CAPTURE_RUNTIME_WRAPPER_PREFIX_RE = /^\[(?:Subagent Context|Subagent Task)\]/i;
+const AUTO_CAPTURE_RUNTIME_WRAPPER_BOILERPLATE_RE =
+  /(?:You are running as a subagent\b.*?(?:$|(?<=\.)\s+)|Results auto-announce to your requester\.?\s*|do not busy-poll for status\.?\s*|Reply with a brief acknowledgment only\.?\s*|Do not use any memory tools\.?\s*)/gi;
 
 function escapeRegExp(value: string): string {
   return value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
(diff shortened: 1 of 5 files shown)


rwmjhb:
### Review: APPROVE (rebase needed)

Good fix — filtering subagent runtime wrappers from auto-capture directly improves recall quality. The three-layer defense (ingress regex, `stripEnvelopeMetadata`, prompt guidance) is solid.

**Before merge:** rebase onto `main` — the `reflection-bypass-hook` test failure is pre-existing, unrelated to your changes.

**Two things to be aware of** (not blocking):

1. **Ingress filter drops entire messages** — `stripAutoCaptureRuntimeWrappers` returns empty string for the whole message when the first line matches `[Subagent Context]` or `[Subagent Task]`. If a message mixes wrapper lines with real user content (e.g., wrapper header + factual statements), the facts are silently lost. Consider stripping only the matching lines instead of the entire message.

2. **`stripEnvelopeMetadata` only strips first line of multiline wrappers** — the regex is line-scoped, so continuation lines from a multi-line `[Subagent Context]` block survive into the extraction prompt.

PR #452, opened 2026-04-02
feat: auto-supersede similar memories on store (isLatest rule-based)

AliceLJY:
## Summary

- When storing a memory with vector similarity > 0.95 and same category as an existing active memory, automatically mark the old one as superseded (set `invalidated_at` + `superseded_by`)
- New memory gets `supersedes` link back to the old one, preserving the version chain
- Eligible categories: preference, fact, entity, other (NOT decision/reflection — these are append-only by nature)
- Remove unused `filterActiveResults` dead code from retriever (store layer already handles this via `excludeInactive: true`)

## Motivation

memory-lancedb-pro 当前在 LongMemEval benchmark 上约 44%。竞品（Supermemory、Mem0）通过版本链确保检索只返回最新事实。本 PR 用规则化方法（向量相似度 > 0.95 + 同 category）实现轻量 isLatest 过滤，不依赖 LLM 调用。

## How it works

```
memory_store("I like Python", category: "preference")
  → stored as mem-A

memory_store("I like Rust", category: "preference")
  → similarity(A, B) = 0.96 > 0.95, same category "preference"
  → auto-supersede: mem-A.invalidated_at = now, mem-A.superseded_by = mem-B
  → mem-B.supersedes = mem-A
  → retrieval returns only mem-B
```

## Test plan

- [x] 8 test cases in `test/is-latest-auto-supersede.test.mjs`
- [x] Similar preference memories trigger auto-supersede
- [x] Old memory metadata has `invalidated_at` and `superseded_by`
- [x] New memory metadata has `supersedes`
- [x] Retrieval only returns new (active) memory
- [x] Decision/reflection NOT auto-superseded
- [x] Cross-category similarities do NOT trigger supersede
- [x] All existing tests pass

### DIFF
--- a/src/tools.ts
+++ b/src/tools.ts
@@ -772,8 +772,11 @@ export function registerMemoryStoreTool(
           // (bypasses importance/recency weighting).
           // Fail-open by design: dedup must never block a legitimate memory write.
           // excludeInactive: superseded historical records must not block new writes.
+          // Align with TEMPORAL_VERSIONED_CATEGORIES: only preference and entity
+          // are semantically version-controlled. "fact"/"other" can reverse-map
+          // to unrelated semantic categories, risking cross-supersede.
           const SUPERSEDE_ELIGIBLE: ReadonlySet<string> = new Set([
-            "preference", "fact", "entity", "other",
+            "preference", "entity",
           ]);
           let existing: Awaited<ReturnType<MemoryStore["vectorSearch"]>> = [];
           try {
@@ -822,13 +825,16 @@ export function registerMemoryStoreTool(
             const factKey =
               oldMeta.fact_key ?? deriveFactKey(oldMeta.memory_category, text);
 
-            // Store new memory with supersedes link
+            // Store new memory with supersedes link, preserving canonical fields
+            // from the old entry (aligns with memory_update supersede path).
             const newMeta = buildSmartMetadata(
               { text, category: category as any, importance: safeImportance },
               {
                 l0_abstract: text,
-                l1_overview: `- ${text}`,
+                l1_overview: oldMeta.l1_overview || `- ${text}`,
                 l2_content: text,
+                memory_category: oldMeta.memory_category,
+                tier: oldMeta.tier,
                 source: "manual",
                 state: "confirmed",
                 memory_layer: deriveManualMemoryLayer(category as string),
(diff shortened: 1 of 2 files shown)


rwmjhb:
## Review: auto-supersede similar memories on store

**Verdict: request-changes** | Confidence: 0.90 | Value: 55%

Good direction — the stale-memory problem is real and worth solving. The rule-based supersede approach (similarity > 0.95 + same category) is a clean design. A few things need attention before merge:

### Must Fix

**1. Category scope conflicts with canonical model (HIGH)**

`SUPERSEDE_ELIGIBLE` includes `fact` and `other`, but the canonical model (`TEMPORAL_VERSIONED_CATEGORIES`) only version-controls `preferences` and `entities`. `reverseMapLegacyCategory()` maps legacy `fact` to either `profile` or `cases` depending on content — so two semantically different `fact` entries could cross-supersede each other if embeddings happen to land in the 0.95-0.98 window.

Suggestion: restrict `SUPERSEDE_ELIGIBLE` to `preference` and `entity` to align with the existing versioning model, or add a `memory_category` check after candidate selection to prevent cross-category superseding.

**2. Tests validate a reimplementation, not the production code (HIGH)**

`simulateMemoryStore()` (lines 79-174) manually reimplements the entire auto-supersede algorithm. It doesn't import or invoke the actual `registerMemoryStoreTool` handler. If a bug is introduced in `tools.ts`, these tests would still pass. The repo already has a pattern for testing through the real handler — see `createTool(registerMemoryStoreTool, ...)` in `recall-text-cleanup.test.mjs`.

Suggestion: either (a) test through the actual tool handler, or (b) extract the supersede logic into a shared pure function that both `tools.ts` and the test import.

**3. Rebase required — build is red (CRITICAL, stale base)**

`plugin-manifest-regression.mjs:155` fails with `AssertionError: sessionMemory should stay disabled by default`. This is a pre-existing failure (`stale_base=true`), but the PR needs a rebase to confirm its changes don't interact with the failing code path.

### Nice to Have

- **Metadata flattening**: superseding via `memory_store` rebuilds metadata from scratch instead of preserving canonical fields (`memory_category`, `tier`, `source_session`, structured overview) from the old entry. The existing `memory_update` supersede path explicitly preserves these — consider aligning.
- **Test `memory_layer` divergence**: `simulateMemoryStore` hardcodes `memory_layer: "durable"`, but production uses `deriveManualMemoryLayer(category)` which returns `"working"` for `entity`/`other`.
- **2 new `as any` casts** in `src/tools.ts` — consistent with existing codebase pattern but widens the type hole.

---

PR #471, opened 2026-04-02
fix: register memory runtime stub to satisfy openclaw doctor check

### LINKED ISSUE #434: [Bug] openclaw doctor shows 'No active memory plugin is registered' despite plugin being loaded and functional
## Bug Description

When running `openclaw doctor`, the "Memory search" diagnostic shows:

```
No active memory plugin is registered for the current config.
```

However, the `memory-lancedb-pro` plugin is actually loaded and fully functional:
- `openclaw doctor` shows: `memory-lancedb-pro@1.1.0-beta.10: plugin registered`
- `memory_stats` tool returns valid data confirming memory is working
- All memory features (recall, store, etc.) work correctly

## Root Cause

The `openclaw doctor` "Memory search" diagnostic checks for memory plugin registration via `api.registerMemoryRuntime()` (from OpenClaw core's memory-state module).

The built-in `memory-core` plugin correctly calls:
```javascript
api.registerMemoryRuntime(memoryRuntime);
api.registerMemoryPromptSection(buildPromptSection);
api.registerMemoryFlushPlan(buildMemoryFlushPlan);
```

However, `memory-lancedb-pro` does NOT call `api.registerMemoryRuntime()`, only:
- `api.registerTool(...)`
- `api.registerHook(...)`
- `api.registerCli(...)`
- `api.on(...)`

This causes the doctor diagnostic to fail, even though the plugin's memory functionality is completely working.

## Expected Behavior

The doctor diagnostic should either:
1. **Detect that memory-lancedb-pro is loaded and working** (preferred), OR
2. memory-lancedb-pro should call `api.registerMemoryRuntime()` to register itself properly with OpenClaw's memory system

## Environment

- OpenClaw version: (recent)
- memory-lancedb-pro version: 1.1.0-beta.10
- macOS
- Plugin path: `~/.openclaw/extensions/memory-lancedb-pro`

## Workaround

None needed - memory works fine, only the doctor diagnostic warning is misleading.

---

*Reported via OpenClaw AI assistant*

king6731253:
## Summary

Registers a minimal `MemoryPluginRuntime` stub so `openclaw doctor` no longer shows the 'No active memory plugin is registered' warning.

Note: memory-lancedb-pro still uses its tool-based architecture for actual memory operations; this stub only satisfies the registration check.

## Changes

- Added `registerMemoryRuntime()` call with minimal stub implementation in `index.ts`
- Stub implements `getMemorySearchManager()` and `resolveMemoryBackendConfig()` with no-op responses
- Includes guard check `if (typeof api.registerMemoryRuntime === 'function')` for forward compatibility

## Fixes

Fixes #434

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -2005,6 +2005,11 @@ const memoryLanceDBProPlugin = {
       `  - Use memory_store or auto-capture for recallable memories.\n`
     );
 
+    // Health status for memory runtime stub (reflects actual plugin health)
+    // Updated by runStartupChecks after testing embedder and retriever
+    let embedHealth: { ok: boolean; error?: string } = { ok: false, error: "startup not complete" };
+    let retrievalHealth: boolean = false;
+
     // ========================================================================
     // Stub Memory Runtime (satisfies openclaw doctor memory plugin check)
     // memory-lancedb-pro uses a tool-based architecture, not the built-in memory-core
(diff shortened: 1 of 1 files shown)


AliceLJY:
This fixes the registration check, but as written it hard-codes `probeEmbeddingAvailability()` to `{ ok: true }` and `probeVectorAvailability()` to `true`.

That means `openclaw doctor` will report memory healthy even when `memory-lancedb-pro` itself cannot actually embed/search (bad API key, dead DB, startup checks failing, etc.), so the warning disappears by masking real failures.

Before this is mergeable, the runtime stub needs to reflect the plugin's real health signals instead of unconditional success.

PR #489, opened 2026-04-03
feat(i18n): add German memory trigger patterns and retrieval triggers

Banger455:
## Summary

Split from #406 – this PR focuses **only** on German i18n support.

### Changes

**`index.ts`**
- Expand `AUTO_CAPTURE_EXPLICIT_REMEMBER_RE` with German (`merk dir`, `vergiss nicht`, `nicht vergessen`) and English (`remember this`) patterns
- - Add 5 German trigger regexes to `MEMORY_TRIGGERS`: explicit-remember, preferences, decisions, personal facts, temporal markers
- - All patterns use `\b` word boundaries to prevent false positives on compound words (e.g. Zimmermann, Schwimmerin)
**`src/adaptive-retrieval.ts`**
- Add 2 German retrieval trigger patterns to `FORCE_RETRIEVE_PATTERNS`: conversational recall (`erinnerst du dich`, `weißt du noch`) and temporal cues (`gestern`, `neulich`, `kürzlich`)
**`test/german-i18n-triggers.test.mjs`** (new)
- 54 test cases covering: 27 positive capture triggers, 10 compound-word false-positive guards, 12+ retrieval triggers, explicit-remember consistency checks
### Test plan
- [x] `node --test test/german-i18n-triggers.test.mjs` passes all 54 cases
- [ ] - [ ] Existing English/CJK tests unaffected (`node --test test/`)
Closes the i18n portion of #406.
### Incidental Changes (noted for reviewers)

**`src/adaptive-retrieval.ts` – `normalizeQuery()`**
- Removed stale blank lines and reordered trailing comment; no logic change.

**`src/adaptive-retrieval.ts` – `SKIP_PATTERNS` regex flag `/i` → `/iu`**
- Required for correct Unicode case-folding on the new German patterns (e.g. `ß`, `ü`, `ä`).
- Existing CJK patterns are unaffected — none rely on case-insensitive matching.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -750,7 +750,7 @@ function shouldSkipReflectionMessage(role: string, text: string): boolean {
 
 const AUTO_CAPTURE_MAP_MAX_ENTRIES = 2000;
 const AUTO_CAPTURE_EXPLICIT_REMEMBER_RE =
-  /^(?:请|請)?(?:记住|記住|记一下|記一下|别忘了|別忘了)[。.!?？!]*$/u;
+  /^(?:请|請)?(?:remember(?:\s+this)?|merk(?:e?\s+dir|\s+es\s+dir)|vergiss\s+(?:das\s+)?nicht|nicht\s+vergessen|记住|記住|记一下|記一下|别忘了|別忘了)[。.!?？!]*$/iu;
 
 /**
  * Prune a Map to stay within the given maximum number of entries.
@@ -1270,6 +1270,12 @@ const MEMORY_TRIGGERS = [
   /老是|講不聽|總是|总是|從不|从不|一直|每次都/,
   /重要|關鍵|关键|注意|千萬別|千万别/,
   /幫我|筆記|存檔|存起來|存一下|重點|原則|底線/,
+  // German triggers
+  /merk(?:e?\s+dir|\s+es\s+dir)|erinner(?:e)?\s+dich|vergiss\s+(?:das\s+)?nicht|nicht\s+vergessen/i,
+  /ich bevorzuge|ich mag|ich hasse|ich will|ich brauche/i,
+  /wir haben entschieden|wir nutzen|ab jetzt|ab sofort|in zukunft/i,
+  /mein\s+\w+\s+(?:ist|heißt)|ich\s+(?:wohne|arbeite)\b/i,
+  /\b(immer\s+(?:wenn|daran|denken|merken|beachten)|niemals|wichtig)\b/i,
 ];
 
 const CAPTURE_EXCLUDE_PATTERNS = [
--- a/index.ts
+++ b/index.ts
@@ -1272,7 +1272,7 @@ const MEMORY_TRIGGERS = [
   /幫我|筆記|存檔|存起來|存一下|重點|原則|底線/,
   // German triggers
   /merk(?:e?\s+dir|\s+es\s+dir)|erinner(?:e)?\s+dich|vergiss\s+(?:das\s+)?nicht|nicht\s+vergessen/i,
-  /ich bevorzuge|ich mag|ich hasse|ich will|ich brauche/i,
+  /ich bevorzuge|ich mag\b|ich hasse|ich will\b|ich brauche/i,
   /wir haben entschieden|wir nutzen|ab jetzt|ab sofort|in zukunft/i,
   /mein\s+\w+\s+(?:ist|heißt)|ich\s+(?:wohne|arbeite)\b/i,
   /\b(immer\s+(?:wenn|daran|denken|merken|beachten)|niemals|wichtig)\b/i,
@@ -1331,14 +1331,14 @@ export function detectCategory(
 ): "preference" | "fact" | "decision" | "entity" | "other" {
   const lower = text.toLowerCase();
   if (
-    /prefer|radši|like|love|hate|want|偏好|喜歡|喜欢|討厭|讨厌|不喜歡|不喜欢|愛用|爱用|習慣|习惯|bevorzuge|ich mag|ich hasse|ich will|ich brauche|lieber|am liebsten/i.test(
+    /prefer|radši|like|love|hate|want|偏好|喜歡|喜欢|討厭|讨厌|不喜歡|不喜欢|愛用|爱用|習慣|习惯|ich bevorzuge|ich mag\b|ich hasse|ich will\b|ich brauche|lieber|am liebsten/i.test(
       lower,
     )
   ) {
     return "preference";
   }
   if (
-    /rozhodli|decided|we decided|will use|we will use|we'?ll use|switch(ed)? to|migrate(d)? to|going forward|from now on|budeme|決定|决定|選擇了|选择了|改用|換成|换成|以後用|以后用|規則|流程|SOP|entschieden|wir nutzen|ab jetzt|ab sofort|in zukunft|wechseln zu|umsteigen/i.test(
+    /rozhodli|decided|we decided|will use|we will use|we'?ll use|switch(ed)? to|migrate(d)? to|going forward|from now on|budeme|決定|决定|選擇了|选择了|改用|換成|换成|以後用|以后用|規則|流程|SOP|\bentschieden|wir nutzen|ab jetzt|ab sofort|in zukunft|wechseln zu|umsteigen/i.test(
       lower,
     )
   ) {
--- a/index.ts
+++ b/index.ts
@@ -1345,7 +1345,7 @@ export function detectCategory(
     return "decision";
   }
   if (
-    /\+\d{10,}|@[\w.-]+\.\w+|is called|jmenuje se|我的\S+是|叫我|稱呼|称呼|heiße|heißt|mein name/i.test(
+    /\+\d{10,}|@[\w.-]+\.\w+|is called|jmenuje se|我的\S+是|叫我|稱呼|称呼|heiße|heißt|mein name/iu.test(
       lower,
     )
   ) {
--- a/index.ts
+++ b/index.ts
@@ -1274,7 +1274,7 @@ const MEMORY_TRIGGERS = [
   /merk(?:e?\s+dir|\s+es\s+dir)|erinner(?:e)?\s+dich|vergiss\s+(?:das\s+)?nicht|nicht\s+vergessen/i,
   /ich bevorzuge|ich mag\b|ich hasse|ich will\b|ich brauche/i,
   /wir haben entschieden|wir nutzen|ab jetzt|ab sofort|in zukunft/i,
-  /mein\s+\w+\s+(?:ist|heißt)|ich\s+(?:wohne|arbeite)\b/i,
+  /mein\s+\w+\s+(?:ist|heißt)|ich\s+wohne\b|ich\s+arbeite\s+bei\b/i,
   /\b(immer\s+(?:wenn|daran|denken|merken|beachten)|niemals|wichtig)\b/i,
 ];
 
--- a/index.ts
+++ b/index.ts
@@ -1331,28 +1331,28 @@ export function detectCategory(
 ): "preference" | "fact" | "decision" | "entity" | "other" {
   const lower = text.toLowerCase();
   if (
-    /prefer|radši|like|love|hate|want|偏好|喜歡|喜欢|討厭|讨厌|不喜歡|不喜欢|愛用|爱用|習慣|习惯/i.test(
+    /prefer|radši|like|love|hate|want|偏好|喜歡|喜欢|討厭|讨厌|不喜歡|不喜欢|愛用|爱用|習慣|习惯|bevorzuge|ich mag|ich hasse|ich will|ich brauche|lieber|am liebsten/i.test(
       lower,
     )
   ) {
     return "preference";
   }
   if (
-    /rozhodli|decided|we decided|will use|we will use|we'?ll use|switch(ed)? to|migrate(d)? to|going forward|from now on|budeme|決定|决定|選擇了|选择了|改用|換成|换成|以後用|以后用|規則|流程|SOP/i.test(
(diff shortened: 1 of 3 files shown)


AliceLJY (changes requested):
The German trigger additions in `index.ts` are well-crafted (good use of `\b` word boundaries, solid false-positive guards in tests). However, `src/adaptive-retrieval.ts` has a **critical encoding corruption** that breaks existing functionality.

## Blocking: UTF-8 mojibake in adaptive-retrieval.ts

The existing CJK characters and emoji in `SKIP_PATTERNS` and `FORCE_RETRIEVE_PATTERNS` have been corrupted. The diff shows:

**SKIP_PATTERNS — emoji corrupted:**
```
-  /^(...|👍|👎|✅|❌)\s*[.!]?$/i,
+  /^(...|ð|ð|â|â)\s*[.!]?$/i,
```

**SKIP_PATTERNS — CJK corrupted:**
```
-  /^(...|实施|實施|开始|開始|继续|繼續|好的|可以|行)\s*[.!]?$/i,
+  /^(...|å®æ½|å¯¦æ½|å¼å§|éå§|ç»§ç»­|ç¹¼çº|å¥½ç|å¯ä»¥|è¡)\s*[.!]?$/i,
```

**FORCE_RETRIEVE_PATTERNS — CJK corrupted:**
```
-  /(你记得|[你妳]記得|之前|上次|以前|还记得|還記得|提到过|提到過|说过|說過)/i,
+  /(ä½ è®°å¾|[ä½ å¦³]è¨å¾|ä¹å|ä¸æ¬¡|ä»¥å|è¿è®°å¾|éè¨å¾|æå°è¿|æå°é|è¯´è¿|èªªé)/i,
```

**Also corrupted:** The comment on L77 (`"你记得吗"` → `"ä½ è®°å¾å"`), the `？` full-width question mark on L72/L81 (`？` → `ï¼`), and German umlauts in the new patterns (`weißt` → `weiÃt`, `früher` → `frÃ¼her`, `kürzlich` → `kÃ¼rzlich`).

This is classic UTF-8-as-Latin-1 mojibake. The file was likely saved or committed with wrong encoding. **This will break all existing Chinese retrieval triggers and emoji skip patterns at runtime.**

## What needs to happen

1. Re-save `src/adaptive-retrieval.ts` with correct UTF-8 encoding. Make sure your editor/git config preserves UTF-8.
2. Verify the German additions (`weißt du noch`, `früher`, `kürzlich`) also come through as proper UTF-8.
3. The test file has the same mojibake in test descriptions (the `Ã` artifacts), though those are cosmetic since assertions test against the imported functions, not source text.

## Non-blocking notes on index.ts (looks good)

- The German trigger regexes in `MEMORY_TRIGGERS` are well-structured
- `\b` word boundaries correctly prevent `Zimmermann`/`Schwimmerin` false positives
- The scoped `immer` pattern (`immer\s+(?:wenn|daran|denken|merken|beachten)`) is a smart improvement over the bare `immer` suggested in #393
- Test coverage is thorough (27 positive, 10 negative, 12 retrieval, explicit-remember consistency)

Please fix the encoding issue and force-push. Happy to re-review after.

Banger455:
Thanks for the thorough review @AliceLJY! Fixed in the latest commit — `src/adaptive-retrieval.ts` has been rewritten with correct UTF-8 encoding. All CJK characters, emoji (👍👎✅❌), and German umlauts (weißt, früher, kürzlich) are now properly encoded. The German retrieval trigger patterns are in place as intended.

AliceLJY (commented):
Thanks for fixing the encoding in the source files (`index.ts`, `adaptive-retrieval.ts`) — those are clean now ✅

However, **the test file `test/german-i18n-triggers.test.mjs` still has mojibake**:

| Line | Corrupted | Should be |
|------|-----------|-----------|
| 33 | `fÃ¼r spÃ¤ter` | `für später` |
| 41 | `Ã¼ber CI` | `über CI` |
| 45 | `fÃ¼r dev` | `für dev` |
| 193 | `WeiÃt du noch` | `Weißt du noch` |
| 229 | `erwÃ¤hnt` | `erwähnt` |

**Critical**: Line 193 tests `shouldSkipRetrieval("WeiÃt du noch...")` — this passes **by coincidence** (returns `false` because it doesn't match ANY pattern, not because the German `weißt du noch` trigger fired). The test gives false confidence without actually validating the German retrieval trigger.

Please re-save the test file with proper UTF-8 encoding. The fix for the source files worked perfectly — just need to apply the same treatment to the test file.

rwmjhb:
## Review: feat(i18n): add German memory trigger patterns and retrieval triggers

Good direction — German users currently get zero capture/retrieval support. The implementation code (`index.ts`, `adaptive-retrieval.ts`) looks correct for proper UTF-8 input.

### Must Fix

**1. Test file has UTF-8 mojibake — tests pass by coincidence**

`test/german-i18n-triggers.test.mjs` contains corrupted strings: `fÃ¼r` instead of `für`, `WeiÃt` instead of `Weißt`, `heiÃt` instead of `heißt`. The `shouldCapture("Mein Projekt heiÃt OpenClaw")` test actually fails when run directly. Other mojibake tests pass by coincidence (the corrupted suffix doesn't affect the trigger match).

AliceLJY flagged this in their second review — still unfixed.

**2. German memories classified as `other` instead of proper categories**

`detectCategory()` has no German branches. German preferences, facts, and decisions all return `"other"` → mapped to `patterns` / `working` layer. English equivalents get `preference` / `durable`. This means German memories have weaker retention and category-aware recall.

### Nice to Have

- `intent-analyzer.ts` also has no German rules — German queries miss category boost in auto-recall ranking
- Undocumented `/iu` flag change on `SKIP_PATTERNS` regex

Banger455:
Both issues addressed in latest commits:

**@AliceLJY** — `test/german-i18n-triggers.test.mjs` encoding fixed (9e56f31). All umlauts/ß/em-dashes now correct UTF-8. Line 193 `Weißt du noch` now properly validates the German retrieval trigger.

**@rwmjhb** — `detectCategory()` now has German branches in all four categories (1ca6c2b):
- **preference**: bevorzuge, ich mag, ich hasse, ich will, ich brauche, lieber, am liebsten
- **decision**: entschieden, wir nutzen, ab jetzt, ab sofort, in zukunft, wechseln zu, umsteigen
- **entity**: heiße, heißt, mein name
- **fact**: ist, sind, hat, haben (inside existing \\b group)

Ready for re-review.

rwmjhb (changes requested):
## Review: feat(i18n): add German memory trigger patterns and retrieval triggers

解决德语用户 memory trigger 静默丢失的方向是对的。但测试文件有 UTF-8 编码问题需要先修：

### Must Fix

1. **Test file mojibake**: `test/german-i18n-triggers.test.mjs` 中多处德语字符损坏——`fÃ¼r`（应为 `für`）、`WeiÃt`（应为 `Weißt`）、`erwÃ¤hnt`（应为 `erwähnt`）。这些测试之所以通过，是因为恰好不匹配任何 pattern，而非真正验证了德语 trigger。AliceLJY 在 4/4 的 review 中已指出此问题。

2. **Category classification**: 德语 auto-captured memories 仍被分类为 `other`，会被降级为 generic working-memory，削弱了这个 feature 的实际效果。

### Scope Drift

- `src/adaptive-retrieval.ts` 中 `normalizeQuery()` 的空行删除和注释重排与德语 i18n 无关。
- `SKIP_PATTERNS` 的 regex flag 从 `/i` 改为 `/iu` 是功能性变更，可能影响现有 CJK pattern 匹配，但 PR description 未提及。

请修复 test file 编码问题，确保德语 trigger 被真正测试到。

Banger455:
Three follow-up commits addressing all review feedback:

**9e56f31** `fix(i18n): restore correct UTF-8 encoding in test file`
All 27 mojibake instances in `test/german-i18n-triggers.test.mjs` replaced (ü/ä/ö/ß/—). Tests now actually validate German patterns instead of passing by coincidence.

**1ca6c2b** `feat(i18n): add German patterns to detectCategory()`
German auto-captured memories are now classified as preference/decision/entity/fact instead of falling through to `other`. Four pattern groups added to existing branches.

**cbf9f2b** `fix(ci): sync stripEnvelopeMetadata with master — two-step subagent stripping`
The branch was forked before a master fix reached `stripEnvelopeMetadata()`. Synced to the two-step approach (strip prefix, then strip boilerplate lines separately). This unblocks the `cli-smoke` CI job.

Also updated the PR description with an **Incidental Changes** section documenting the `normalizeQuery()` whitespace cleanup and `/i` → `/iu` flag rationale (Unicode case-folding for ß/ü/ä; no impact on existing CJK patterns).

Banger455:
Follow-up fix after deep review:

fix(ci): restore full stripEnvelopeMetadata - Previous commit used a brace-counter that matched } inside a regex literal as the function closing brace, truncating steps 1-4 (System timestamps, metadata sections, JSON block stripping, blank-line collapse). This restores the complete function with the two-step subagent fix plus all original metadata stripping steps.

fix(i18n): use /iu flag on detectCategory entity branch - The entity branch regex contains heiße/heißt which needs Unicode case-folding (/iu) to match all-caps input. Changed /i to /iu on that branch.

Banger455:
Hey, rebased onto current master and cleaned things up a bit:

- Dropped the `smart-extractor.ts` changes entirely — master already has a much better solution with the extracted helper functions, so our two-step regex approach wasn't needed anymore. That also addresses the scope drift concern from @rwmjhb.
- The `detectCategory()` German branches, retrieval triggers, and test file are all still there and working fine.
- All 57 tests pass locally after rebase.

So the diff is now just 3 files (index.ts, adaptive-retrieval.ts, test file) — pure i18n, nothing else.

CI will probably need an "Approve and run" from a maintainer since this is a fork PR. Would appreciate a re-review when you get a chance!

app3apps:
Re-checked this PR and I still see one merge-blocking behavior issue.

The new German personal-info trigger is currently too broad: `ich arbeite ...` now matches ordinary work-status messages, not just stable personal info. In practice, inputs like “Ich arbeite gerade an PR 489” or “Ich arbeite noch am Fix für den Test” now pass `shouldCapture`, which means they can flow into the regex fallback capture path as long-term memory.

I verified that against the current head and compared it with base behavior, where those same inputs were not captured.

I’d hold merge until that trigger is narrowed, for example toward `ich arbeite bei ...`, and covered with a negative regression test. The CI wiring for the new German test still looks incomplete too, but I’d treat that as follow-up rather than the blocker here.

Banger455:
@app3apps Good catch — fixed in 44b9ebc.

**Change:** `ich\s+(?:wohne|arbeite)\b` → `ich\s+wohne\b|ich\s+arbeite\s+bei\b`

Now only employer statements like "Ich arbeite bei Anthropic" trigger capture. Transient work status ("Ich arbeite gerade an PR 489", "Ich arbeite heute von zuhause") is no longer matched.

Added 3 negative regression tests. 72 tests pass.

Banger455:
Note: the `cli-smoke` CI failure is **not caused by this PR**. The same assertion (`test/cli-smoke.mjs:316 — recallResult.details.count` is `undefined`) is also failing on master since the April 12 run ([#24304669192](https://github.com/CortexReach/memory-lancedb-pro/actions/runs/24304669192)). Our branch has zero diff on that file vs master.

rwmjhb (changes requested):
The problem is real and the pattern coverage is solid. Two test encoding issues need fixing before this can be trusted to work.

**Must fix**

**F2/F4 — UTF-8 mojibake in test file causes core assertions to pass by coincidence**
`test/german-i18n-triggers.test.mjs` contains corrupted bytes: `WeiÃt` (U+00C3 U+0054) instead of `Weißt` (U+00DF), `heiÃt` instead of `heißt`, `fÃ¼r` instead of `für`, etc.

The tests pass because the corrupted strings don't match any pattern — so `shouldSkipRetrieval` defaults `false` for long queries, and `shouldCapture` falls through to a different trigger. The German patterns themselves are not being exercised. AliceLJY flagged this in their 2026-04-04 review and it's still present.

Fix: re-save the file with correct UTF-8. Verify with hex: `ß` should be bytes `C3 9F`, not `C3 83 54`.

**MR1 — German auto-captured memories are classified as `other`**
The category assignment is unchanged, so German-triggered memories land in `other` and are treated as generic working-memory patterns rather than domain knowledge. Consider mapping common German capture patterns to the appropriate category in the same pass.

---

**Non-blocking**

- **F3** — Most retrieval tests also use corrupted strings, so they pass coincidentally. Fix encoding across the whole file, not just F2/F4 lines.
- **F5** — `SKIP_PATTERNS` affirmation regex has `/i` changed to `/iu`. Functional change not mentioned in the PR description — could affect CJK pattern matching. Please document the intent.
- **F6** — Blank-line removal and comment reformatting in `normalizeQuery()` is unrelated to German i18n. Minor scope drift but not blocking.

rwmjhb (changes requested):
German trigger patterns are a real gap, and adding them is the right call. A few issues need to be resolved before merge.

**Must fix before merge:**

- **MR1: German auto-captured memories are classified as `other`.** The `other` category demotes memories to generic working-memory patterns, losing the semantic categorization benefit. German patterns should map to the appropriate semantic categories (preference, procedure, etc.) the same way English patterns do.

- **F2 / F4: Mojibake in test file causes core patterns to go untested.** The test strings for `'weißt du noch'` and `'heißt'` are corrupted due to encoding issues in the test file. These tests currently pass vacuously rather than validating the actual patterns — encoding must be fixed so the tests exercise real inputs.

- F5: The `/iu` flag change on `SKIP_PATTERNS` (adding unicode flag) is undocumented — is this intentional? It could affect existing CJK pattern matching.
- MR2: Adaptive recall intent routing still does not understand German preference or decision queries — consider a follow-up.
- F3: Most retrieval tests pass by coincidence rather than validating German-specific trigger behavior.

Fix MR1 and the encoding issue (F2/F4), then this is ready.

PR #495, opened 2026-04-03
fix: strip subagent wrapper continuation lines

AliceLJY:
## Summary
- strip `[Subagent Context]` / `[Subagent Task]` wrapper lines before memory extraction
- treat the follow-on subagent runtime continuation lines as orchestration metadata, not conversation content
- restore the existing `strip-envelope-metadata` regression expectations on current `master`

## Testing
- `node --test test/strip-envelope-metadata.test.mjs`

This is split out as a standalone baseline fix so unrelated PRs do not stay red because of the current subagent-wrapper regression on `master`.

### DIFF
--- a/src/smart-extractor.ts
+++ b/src/smart-extractor.ts
@@ -70,14 +70,14 @@ import { batchDedup } from "./batch-dedup.js";
  * - Standalone JSON blocks containing message_id/sender_id fields
  */
 export function stripEnvelopeMetadata(text: string): string {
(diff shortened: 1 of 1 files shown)


AliceLJY:
Closing — master already has a more robust implementation via `stripLeadingRuntimeWrappers()` (line 71 of smart-extractor.ts) that covers all the patterns this PR targeted. No remaining unique changes after rebase.

PR #493, opened 2026-04-03
feat: Proposal A v3 - Dynamic Importance Feedback Signals (Phase 1)

jlin53882:
## Summary

Proposal A Phase 1: Dynamic importance feedback signals — importance scores are dynamically adjusted based on whether recalled memories were actually used in the agent's response.

## Phase 1（已完成 ✅）

• agent_end 捕捉回應文字
• isRecallUsed() 判斷是否使用
• +0.05 確認使用，-0.03 連續 2 次未使用（threshold 已從 3 改為 2）
• bad_recall_count counter 遞增

---

### Phase 1 Feedback Signals（已實作 ✅）

| Signal | 條件 | 調整 |
|--------|------|------|
| 用戶明確確認 | 使用者回答「正確」/「是」/「對」/「right」 | importance += 0.15 上限 1.0 |
| 用戶明確標記錯誤 | 使用者表示「不」/「錯誤」/「not right」 | importance -= 0.10 下限 0.1 |

---

## Phase 2 預計處理

• min_recall_count 閾值（AliceLJY 建議）
• 調整幅度變成可配置（importanceBoostOnUse、importancePenaltyOnMiss 等）

---

## 已修復的 Bugs（Phase 1 內）

| 等級 | 問題 | 修復方式 |
|------|------|---------|
| 🔴 P1 | isRecallUsed() 只找 stock phrases 和 raw IDs，不找實際注入的 summary | 新增 injectedSummaries verbatim match |
| 🔴 P1 | 用戶確認/錯誤關鍵字應從下一輪 user prompt 讀取 | 從 event.messages 萃取 role=user 的 content |
| 🔴 P1 | usedRecall 套用到所有 recallIds，多筆記憶集體 boost/penalty | 每個 recallId 個別迴圈評分 |
| 🔴 P1 | pendingRecall 只用 sessionKey，跨 agent 覆寫 | 改用 sessionKey:agentId 複合鍵 |
| 🟡 P2 | parsePluginConfig() 未解析 feedback 區塊 | 新增 feedback 解析邏輯 |
| 🟡 P2 | event.prompt 是 string 非 array | 改用 event.messages 萃取 |
| 🟡 P2 | Promise.race timeout 殘留 pendingRecall | finally { pendingRecall.delete() } |
| 🟡 P2 | feedback hooks 包在 autoCapture 條件內 | 鉤子移到 if block 外部 |

## Files changed

| File | Change |
|------|--------|
[...]

PR #499, opened 2026-04-04
fix: strip boilerplate continuation lines in stripLeadingRuntimeWrappers

jlin53882:
## Summary

Strip known boilerplate continuation lines (e.g. `Results auto-announce to your requester.`, `Do not use any memory tools.`) that appear right after the `[Subagent Context]` / `[Subagent Task]` wrapper prefix but do not match the wrapper prefix regex.

## What was fixed

**Problem**: `stripLeadingRuntimeWrappers()` correctly stripped wrapper prefix lines and their first-line content, but left behind boilerplate continuation lines that appear on separate lines after the prefix. These orphan boilerplate lines would leak into auto-captured memories as noise.

**Fix**: In the `strippingLeadIn` loop, check each continuation line against `AUTO_CAPTURE_RUNTIME_WRAPPER_BOILERPLATE_RE` and `continue` if matched. This prevents boilerplate continuation lines from being preserved.

## Files changed
- `src/auto-capture-cleanup.ts`: +11 lines (new constants + logic)

## Testing
- `[Subagent Context] You are running as a subagent... Results auto-announce to your requester.` -> fully stripped
- Real content after wrapper block is preserved
- Wrapper-only payloads collapse to empty (handled by existing null-return guard)

## Related
- Split out from PR #493 (Proposal A v3) per AliceLJY review feedback
- Follow-up to fix/preserve-content-after-subagent-wrappers (#444)

### DIFF
--- a/src/auto-capture-cleanup.ts
+++ b/src/auto-capture-cleanup.ts
@@ -113,6 +113,17 @@ function stripLeadingRuntimeWrappers(text: string): string {
       continue;
     }
 
+    // Bug fix: also strip known boilerplate continuation lines (e.g.
+    // "Results auto-announce to your requester.", "Do not use any memory tools.")
+    // that appear right after the wrapper prefix. These lines do NOT match the
+    // wrapper prefix regex but are part of the wrapper boilerplate.
+    if (strippingLeadIn) {
+      AUTO_CAPTURE_RUNTIME_WRAPPER_BOILERPLATE_RE.lastIndex = 0;
+      if (AUTO_CAPTURE_RUNTIME_WRAPPER_BOILERPLATE_RE.test(current)) {
+        continue;
+      }
+    }
+
     strippingLeadIn = false;
     cleanedLines.push(line);
   }


rwmjhb:
## Review: fix: strip boilerplate continuation lines in stripLeadingRuntimeWrappers

Good problem — wrapper boilerplate leaks into auto-captured memories.

### Must Fix

**1. Only fixes one of two wrapper-stripping paths**

`auto-capture-cleanup.ts` is fixed, but `smart-extractor.ts` has its own independent regex at line 72-80 that still fails. Codex found the root cause: the regex at `smart-extractor.ts:76` contains a literal backspace byte (`0x08`) instead of `\b` word boundary — the pattern never matches.

**2. Rebase needed — 2 test failures in strip-envelope-metadata.test.mjs**

AliceLJY:
Closing -- superseded by #508 which includes the same boilerplate-skip logic plus the critical sawWrapperPrefix guard that this PR lacks. Merging this PR alone would introduce a false-positive stripping bug.

Issue #417, opened 2026-03-30
Bug: extractMinMessages=2 + autoCaptureSeenTextCount 累积逻辑失效 → 所有单轮对话都掉入 regex fallback，污染全库为脏数据


lintianyuan666:
### Plugin Version

1.1.0

### OpenClaw Version

2026.3.28

### Bug Description

开启 smartExtraction: true 后，正常配置 extractMinMessages: 2，但 auto-capture 几乎所有对话都落入 regex fallback 写入 raw text，导致库内记忆全部是脏数据（l0_abstract == text，无 LLM 蒸馏）。
代码有两条累积路径试图凑够 extractMinMessages，两条都失效：

───

路径A：autoCaptureSeenTextCount diffing（失效）

// index.ts line 2169-2176
const previousSeenCount = autoCaptureSeenTextCount.get(sessionKey) ?? 0;
let newTexts = eligibleTexts;  // ← 每次 agent_end 的 eligibleTexts 是"当前事件的消息数"，不是"历史累积量"
if (pendingIngressTexts.length > 0) {
  newTexts = pendingIngressTexts;
} else if (previousSeenCount > 0 && eligibleTexts.length > previousSeenCount) {
  newTexts = eligibleTexts.slice(previousSeenCount);  // ← 永远不会触发，因为 eligibleTexts.length === previousSeenCount === 1
}
autoCaptureSeenTextCount.set(sessionKey, eligibleTexts.length);  // ← 每次覆盖成"1"，diffing 失效

在单轮 DM 场景：

• 事件1：eligibleTexts=1, previousSeenCount=0 → newTexts=1 → smart extraction 跳过（需要≥2）
• 事件2：eligibleTexts=1, previousSeenCount=1 → 1 > 1 为 false → newTexts=1 → 同样跳过

日志佐证：

08:44:28 smart-extractor: extracted 3 candidates  ← 历史累积生效过一次（跨会话或特定模式）
08:46:41 regex fallback found 1 capturable text(s)  ← 后续全走 regex

───

路径B：pendingIngressTexts 跨消息累积（冷启动失效）

// message_received hook — 累积入口
const conversationKey = buildAutoCaptureConversationKeyFromIngress(channelId, conversationId);
queue.push(normalized);  // ← 来自用户发送的 ingress 消息

// agent_end hook — 消费出口
[...]

Issue #8, opened 2026-02-27
[BUG] Ollama nomic-embed-text dimension mismatch (expected 768, got 192)


MrAnderson1989:
## Bug Description

When using `nomic-embed-text` with Ollama, the plugin fails with dimension mismatch error.

## Expected Behavior
- Config: `dimensions: 192`
- Plugin should use 192 dimensions as configured

## Actual Behavior
- Plugin's `EMBEDDING_DIMENSIONS` hardcodes `nomic-embed-text` to 768 dimensions
- Ollama's `nomic-embed-text` model actually outputs 192 dimensions
- Error: dimension mismatch (expected 768, got 192)

## Root Cause
In `src/embedder.ts`, the `EMBEDDING_DIMENSIONS` map has:
```typescript
"nomic-embed-text": 768,
```
But Ollama's version of this model outputs 192-dimensional embeddings.

## Suggested Fix
1. Update the hardcoded dimension for `nomic-embed-text` from 768 to 192, OR
2. Make the dimension config actually override the hardcoded values (currently not working)

## Environment
- Plugin: memory-lancedb-pro
- Ollama version: (latest)
- Model: nomic-embed-text

Issue #4, opened 2026-02-25
CLI Commands 無效


linax777:
README  上面寫的 openclaw memory  命令無效   系統還是使用自帶的 memory 命令

其實介紹 youtube 影片 06:41 的對話就有說到 註冊 openclaw memory 命令異常

win4r:
Thanks for reporting this! Two bugs were causing CLI commands to fail:

**Bug 1: Command name collision**
OpenClaw has a built-in `memory` command (`index`/`search`/`status`). When the plugin tried to register its own `memory` command, OpenClaw's CLI loader detected the overlap and skipped the plugin entirely.

**Bug 2: Callback signature mismatch**
OpenClaw passes a context object `{ program, config, ... }` to CLI registrars, but the plugin expected `program` directly — so even without the name collision, registration would have thrown an error.

**Fix (fb99ea4):**
1. Fixed the callback signature: `(program)` → `({ program })`
2. Renamed the CLI command from `memory` to `memory-pro` to avoid collision

After updating, CLI commands work as:
```bash
openclaw memory-pro stats
openclaw memory-pro search "query"
openclaw memory-pro list
openclaw memory-pro export
# etc.
```

To update: `git pull` your plugin directory and restart the gateway (`rm -rf /tmp/jiti/ && openclaw gateway restart`).

PR #518, opened 2026-04-04
fix: auto-capture cumulative turn counting for smart extraction (issue #417)

jlin53882:
## Summary

Fixes the cumulative turn counting for smart extraction in auto-capture. Resolves issue #417.

## Problem

With `extractMinMessages: 2` + `smartExtraction: true`, single-turn DM conversations (1 user message per `agent_end`) always fall through to regex fallback, writing dirty data (`l0_abstract == text`, no LLM distillation).

Root cause: `autoCaptureSeenTextCount` was being overwritten per-event (always set to current event's message count = 1), never accumulating. Two fix paths identified:

- **Path A**: `autoCaptureSeenTextCount` diffing (fix: accumulate instead of overwrite)
- **Path B**: `buildAutoCaptureConversationKeyFromIngress` returned `null` for DM (no `conversationId`), so `pendingIngressTexts` was never written (fix: fallback to `channel` for DM)

### Fix 1: `buildAutoCaptureConversationKeyFromIngress` (Path B)
DM: `conversationId` is `undefined` → returns `channel` instead of `null`. Now matches the key extracted by regex in `agent_end`.

### Fix 2: `autoCaptureSeenTextCount` cumulative counting (Path A)
Changed from `set(sessionKey, eligibleTexts.length)` (overwrite) to `set(sessionKey, currentCumulativeCount)` (accumulate).

### Fix 3: Smart extraction threshold uses cumulative count
`currentCumulativeCount >= minMessages` instead of `cleanTexts.length >= minMessages`.

### Fix 4: `isExplicitRememberCommand` guard
Preserves explicit remember command behavior in DM context.

### Fix 5: `extractMinMessages` cap
`Math.min(config.extractMinMessages ?? 4, 100)` — prevents misconfiguration.

### Fix 6: `MAX_MESSAGE_LENGTH = 5000` guard
Prevents OOM from super-long messages in `pendingIngressTexts`.

## Semantic Change (Needs Discussion)

The fix changes `extractMinMessages` semantics from "per-event message count" to "cumulative conversation turns". This aligns with the design intent (AliceLJY confirmed this is the correct direction for beta.11), but existing users who configured `extractMinMessages` based on the old semantics may see different behavior.

### Q1: `extractMinMessages` semantic change acceptable?
**Before**: `extractMinMessages=N` = each `agent_end` event needs N messages
**After**: `extractMinMessages=N` = smart extraction triggers at turn N

AliceLJY confirmed this is the correct direction. Is this semantic change acceptable as a bug fix, or should it be a separate config option?

### Q2: `isExplicitRememberCommand` behavior after Fix 3
When `pendingIngressTexts.length > 0` (multi-turn), `texts.length` is typically > 1, so `texts.length === 1 && isExplicitRememberCommand(...)` guard rarely triggers. The fix preserves this by checking `lastPending` fallback. Is this the intended behavior?

### Q3: Window size for cumulative counting
The rolling window for `pendingIngressTexts` is `-6` messages. For DM with `extractMinMessages=2`, this is sufficient. But if `extractMinMessages` is set to a very large value (e.g., 50), the window may not hold enough history. Should there be a dynamic adjustment?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,26 @@
 # Changelog
 
+## Unreleased
+
+### Fix: cumulative turn counting for auto-capture smart extraction (#417, PR #518)
+
+**Bug**: With `extractMinMessages: 2` + `smartExtraction: true`, single-turn DM conversations always fell through to regex fallback, writing dirty data (`l0_abstract == text`, no LLM distillation).
+
+**Root causes**:
+- `autoCaptureSeenTextCount` was overwritten per-event (always 1 for DM), never accumulating
+- `buildAutoCaptureConversationKeyFromIngress` returned `null` for DM (no `conversationId`), so `pendingIngressTexts` was never written
+
+**Changes**:
+- **Cumulative counting**: `autoCaptureSeenTextCount` now accumulates across events instead of overwriting per-event
+- **DM key fallback**: `buildAutoCaptureConversationKeyFromIngress` falls back to `channelId` for DM, fixing `pendingIngressTexts` writes
+- **Smart extraction threshold**: now uses cumulative turn count (`currentCumulativeCount`) instead of per-event message count
+- **`extractMinMessages` cap**: `Math.min(config.extractMinMessages ?? 4, 100)` prevents misconfiguration
+- **MAX_MESSAGE_LENGTH guard**: 5000 char limit in `pendingIngressTexts` prevents OOM
+
+**Note**: `extractMinMessages` semantics changed from "per-event message count" to "cumulative conversation turns". This is a bug fix since the old semantics were broken for DM; a changelog note is included for beta.11 users.
+
+---
+
 ## 1.1.0-beta.2 (Smart Memory Beta + Access Reinforcement)
 
 This is a **beta** release published under the npm dist-tag **`beta`** (it does not affect the stable `latest` channel).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,12 +12,13 @@
 
 **Changes**:
 - **Cumulative counting**: `autoCaptureSeenTextCount` now accumulates across events instead of overwriting per-event
-- **DM key fallback**: `buildAutoCaptureConversationKeyFromIngress` falls back to `channelId` for DM, fixing `pendingIngressTexts` writes
+- **DM key fallback**: `buildAutoCaptureConversationKeyFromIngress` falls back to `channelId` when `conversationId` is falsy, so DM sessions now correctly write to `pendingIngressTexts` and match the key extracted by `buildAutoCaptureConversationKeyFromSessionKey`
 - **Smart extraction threshold**: now uses cumulative turn count (`currentCumulativeCount`) instead of per-event message count
-- **`extractMinMessages` cap**: `Math.min(config.extractMinMessages ?? 4, 100)` prevents misconfiguration
-- **MAX_MESSAGE_LENGTH guard**: 5000 char limit in `pendingIngressTexts` prevents OOM
+- **`extractMinMessages` cap**: `Math.min(config.extractMinMessages ?? 4, 100)` prevents misconfiguration (e.g., setting 999999 would permanently disable smart extraction)
+- **MAX_MESSAGE_LENGTH guard**: 5000 char limit per message in `pendingIngressTexts` rolling window prevents OOM from malformed input
+- **Test**: added `runCumulativeTurnCountingScenario` in `test/smart-extractor-branches.mjs` verifying turn-1 skip and turn-2 trigger with `extractMinMessages=2`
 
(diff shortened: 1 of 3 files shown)


AliceLJY (approved):
Core fix is correct -- cumulative counting resolves the single-turn DM regex-fallback pollution described in #417. The 6 sub-fixes work together logically.

Minor questions (non-blocking):
1. Was removing `autoCapturePendingIngressTexts.delete(conversationKey)` intentional? The old code cleared pending texts after consumption.
2. Consider adding a test simulating 2 sequential `agent_end` events with `extractMinMessages=2` to lock in the fix.
3. The `extractMinMessages` semantic change (per-event to cumulative) is the right call since the old semantics were broken, but worth a changelog note for beta.11.

CI `cli-smoke` failure is pre-existing on master, unrelated. Approved.

rwmjhb (changes requested):
## Review: fix: auto-capture cumulative turn counting for smart extraction (#417)

问题定位准确——单轮 DM 因为 `autoCaptureSeenTextCount` 被每次覆写而永远触发 regex fallback，导致脏数据。但实现有几个需要关注的点：

### Must Fix

1. **语义混淆**: `autoCaptureSeenTextCount` 现在混合了 transcript length 和 cumulative-turn 两种语义，`currentCumulativeCount` 在整个 session 生命周期内单调递增无 reset——超过 turn N 之后每轮都会触发 smart extraction。

2. **Build 失败**: TypeScript 编译不过，需要先修。

### Questions

- `autoCapturePendingIngressTexts.delete(conversationKey)` 被移除但 PR description 没说明——是有意为之吗？如果是，什么机制防止 stale pending texts 被重复消费？
- `isExplicitRememberCommand` guard 在 multi-turn DM（这个 PR 的目标场景）下不可达——是设计意图还是 bug？
- 新测试的 `createMockApi` 传了第 5 个参数但函数签名不接受。

jlin53882:
## Phase 2: Design Questions for Maintainer Input

Phase 1 cleanup (PR #534) is ready — fixes `createMockApi` interface and removes dead code.

Before Phase 2 (semantic fixes), we need maintainer input on these 5 questions:

---

### Q1: Smart Extraction Re-trigger Policy

**Issue**: `currentCumulativeCount` is monotonic (no reset). Once it crosses `extractMinMessages=N`, every subsequent turn triggers smart extraction.

**Current behavior (with Phase 1)**:
- Turn 1: cumulative=1, skip
- Turn 2: cumulative=2, trigger ✓
- Turn 3: cumulative=3, trigger again ⚠️
- Turn 4: cumulative=4, trigger again ⚠️

**Is this the intended behavior? Should there be a session-level once-only flag?**

---

### Q2: REPLACE vs APPEND Strategy

**PR #518 chose REPLACE** (`newTexts = pendingIngressTexts`) to avoid deduplication issues when the same text appears in both `pendingIngressTexts` and `eligibleTexts`.

**Alternative**: Keep APPEND but deduplicate before joining.

**Which strategy is preferred?**

---

### Q3: Stale `pendingIngressTexts` Prevention

**Issue**: `autoCapturePendingIngressTexts.delete(conversationKey)` was removed in PR #518, relying on REPLACE + `slice(-6)` to prevent stale data.

**Concern**: If `agent_end` fires without a preceding `message_received` (e.g., direct trigger, retry/replay), stale texts could be consumed.

**Is there a mechanism to ensure `message_received` always precedes `agent_end` for a given conversationKey?**

---

### Q4: `extractMinMessages` Semantic Change

PR #518 changes semantics from "per-event message count" to "cumulative conversation turns".

This is a breaking change for users who configured `extractMinMessages` based on the old semantics.

**Is this semantic change acceptable as a bug fix, or should it be a separate config option?**

---

### Q5: `isExplicitRememberCommand` Guard (Dead Code)

Phase 1 removes this guard as unreachable under REPLACE, but I want to confirm this is correct:

**Original intent**: Handle single "請記住..." messages without enough context, by appending `priorRecentTexts`.

**Why unreachable now**: Under REPLACE, `texts = pendingIngressTexts` (rolling window of up to 6 messages), so `texts.length === 1` can never be true in multi-turn scenarios.

**Should this behavior be preserved? If so, where should the context-enrichment logic live?**

---

PR #534 (Phase 1): https://github.com/CortexReach/memory-lancedb-pro/pull/534
PR #518 (main PR): https://github.com/CortexReach/memory-lancedb-pro/pull/518

jlin53882:
This PR has been superseded by PR #549, which resolves all blocking concerns from the PR #534 review (Must Fix #1, #2, #3, and Minor item #4) and additionally corrects a placement bug in Fix #1 identified during adversarial review.

Summary of fixes in #549:
- **Fix #8**: `pendingIngressTexts.delete()` — clears consumed pending texts
- **Fix #9**: `currentCumulativeCount` reset moved inside success block — resets only on successful extraction (not on failure), preventing re-trigger loops
- **Fix #4**: `pluginConfigOverrides` comment explaining embedding override behavior

Please review #549 instead.

Issue #12, opened 2026-02-28
根据文档安装上插件后gateway无法正常启动


LinYKen:
尝试重启gateway会出现以下超时问题

Restarted systemd service: openclaw-gateway.service
Timed out after 60s waiting for gateway port 18789 to become healthy.
Gateway process is running but port 18789 is still free (startup hang/crash loop or very slow VM startup).
Service runtime: status=running, state=active, pid=145586, lastExit=0
Gateway port 18789 status: free.
Gateway restart timed out after 60s waiting for health checks.

stop后重启也无法访问

LinYKen:
OpenClaw 2026.2.24

win4r:
已在 v1.0.10 修复：启动阶段的 embedder/retriever 自检不再阻塞 gateway 启动（改为后台执行并加超时），避免外部 embedding/rerank 服务慢/不可用导致 18789 端口一直不监听从而 restart 超时。\n\n请升级：\n- npm: memory-lancedb-pro@1.0.10\n- Release: https://github.com/win4r/memory-lancedb-pro/releases/tag/v1.0.10\n\n验证点：重启后应在 60s 内看到 gateway 18789 端口监听；即使 embedding key 无效也只会输出 warn，不会卡死。\n

win4r:
先关闭 issue；如果升级到 v1.0.10 后仍复现（端口不监听/重启超时），请重新打开并附上 /tmp/openclaw/openclaw-*.log 相关片段。

PR #534, opened 2026-04-05
fix: Phase 1 review fixes for PR #518 (issue #417)

### LINKED ISSUE #3: rerankEndpoint属性没有加到openclaw.plugin.json里面
如题，会导致gateway启动的时候报错
“lugins.entries.memory-lancedb-pro.config: invalid config: retrieval: must NOT have additional properties”

jlin53882:
## Summary

Phase 1 fixes for PR #518 review feedback. These are pure cleanup fixes with no semantic changes:

1. `createMockApi` now accepts an optional 5th `pluginConfigOverrides` parameter
2. Removed unreachable dead code block (`isExplicitRememberCommand` guard)

See original PR #518 for the main feature.

### Fix 1: `createMockApi` accepts pluginConfigOverrides (test/smart-extractor-branches.mjs)

The test was calling `createMockApi(..., { extractMinMessages: 2, smartExtraction: true })` as a 5th parameter, but the function only accepted 4 parameters. The 5th argument was silently ignored.

```javascript
// Before
function createMockApi(dbPath, embeddingBaseURL, llmBaseURL, logs) {

// After
function createMockApi(dbPath, embeddingBaseURL, llmBaseURL, logs, pluginConfigOverrides = {}) {
```

`pluginConfigOverrides` is spread into `pluginConfig`, allowing tests to override defaults.

### Fix 2: Removed dead `isExplicitRememberCommand` guard (index.ts)

The guard:
```typescript
if (
  texts.length === 1 &&
  lastPending !== null &&
  isExplicitRememberCommand(lastPending) &&
  priorRecentTexts.length > 0
) {
  texts = [...pendingIngressTexts, ...priorRecentTexts.slice(-1), ...eligibleTexts];
}
```

This is unreachable under the REPLACE strategy (Fix #3 of PR #518):
- `texts = newTexts = pendingIngressTexts` (REPLACE mode)
- `pendingIngressTexts` is a rolling window of up to 6 messages
- `texts.length === 1` can only be true when `pendingIngressTexts.length === 1`
- In normal multi-turn DM, `pendingIngressTexts.length` grows from 1→2→...→6
- The guard was designed for the old APPEND strategy and is obsolete

## Testing

- `smart-extractor-branches.mjs`: passes ✅
- `npm test`: all pre-existing tests pass ✅

## Known Issues (Out of Scope for Phase 1)

PR #518 has 2 pre-existing test failures in `strip-envelope-metadata.test.mjs` (tests #12 and #13) caused by other changes in the PR, unrelated to these Phase 1 fixes.

## Phase 2 Discussion Points (pending maintainer input)

The following design questions need maintainer input before Phase 2 can proceed:

1. **Smart extraction re-trigger policy**: After cumulative count crosses threshold, should it trigger once per session or on every subsequent turn?
2. **Stale pending texts mechanism**: What prevents stale `pendingIngressTexts` from being consumed twice if `agent_end` fires without a preceding `message_received`?
3. **REPLACE vs APPEND strategy**: REPLACE was chosen to avoid deduplication issues. Is APPEND feasible with proper deduplication?
4. **`extractMinMessages` semantic change**: This PR changes semantics from "per-event message count" to "cumulative turns". Is this acceptable?
5. **`isExplicitRememberCommand` guard removal**: What was the original intent of this guard? Is removing it correct?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,26 @@
 # Changelog
 
+## Unreleased
+
+### Fix: cumulative turn counting for auto-capture smart extraction (#417, PR #518)
+
+**Bug**: With `extractMinMessages: 2` + `smartExtraction: true`, single-turn DM conversations always fell through to regex fallback, writing dirty data (`l0_abstract == text`, no LLM distillation).
+
+**Root causes**:
+- `autoCaptureSeenTextCount` was overwritten per-event (always 1 for DM), never accumulating
+- `buildAutoCaptureConversationKeyFromIngress` returned `null` for DM (no `conversationId`), so `pendingIngressTexts` was never written
+
+**Changes**:
+- **Cumulative counting**: `autoCaptureSeenTextCount` now accumulates across events instead of overwriting per-event
+- **DM key fallback**: `buildAutoCaptureConversationKeyFromIngress` falls back to `channelId` for DM, fixing `pendingIngressTexts` writes
+- **Smart extraction threshold**: now uses cumulative turn count (`currentCumulativeCount`) instead of per-event message count
+- **`extractMinMessages` cap**: `Math.min(config.extractMinMessages ?? 4, 100)` prevents misconfiguration
+- **MAX_MESSAGE_LENGTH guard**: 5000 char limit in `pendingIngressTexts` prevents OOM
+
+**Note**: `extractMinMessages` semantics changed from "per-event message count" to "cumulative conversation turns". This is a bug fix since the old semantics were broken for DM; a changelog note is included for beta.11 users.
+
+---
+
 ## 1.1.0-beta.2 (Smart Memory Beta + Access Reinforcement)
 
 This is a **beta** release published under the npm dist-tag **`beta`** (it does not affect the stable `latest` channel).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,12 +12,13 @@
 
 **Changes**:
 - **Cumulative counting**: `autoCaptureSeenTextCount` now accumulates across events instead of overwriting per-event
-- **DM key fallback**: `buildAutoCaptureConversationKeyFromIngress` falls back to `channelId` for DM, fixing `pendingIngressTexts` writes
+- **DM key fallback**: `buildAutoCaptureConversationKeyFromIngress` falls back to `channelId` when `conversationId` is falsy, so DM sessions now correctly write to `pendingIngressTexts` and match the key extracted by `buildAutoCaptureConversationKeyFromSessionKey`
 - **Smart extraction threshold**: now uses cumulative turn count (`currentCumulativeCount`) instead of per-event message count
-- **`extractMinMessages` cap**: `Math.min(config.extractMinMessages ?? 4, 100)` prevents misconfiguration
-- **MAX_MESSAGE_LENGTH guard**: 5000 char limit in `pendingIngressTexts` prevents OOM
+- **`extractMinMessages` cap**: `Math.min(config.extractMinMessages ?? 4, 100)` prevents misconfiguration (e.g., setting 999999 would permanently disable smart extraction)
+- **MAX_MESSAGE_LENGTH guard**: 5000 char limit per message in `pendingIngressTexts` rolling window prevents OOM from malformed input
+- **Test**: added `runCumulativeTurnCountingScenario` in `test/smart-extractor-branches.mjs` verifying turn-1 skip and turn-2 trigger with `extractMinMessages=2`
 
-**Note**: `extractMinMessages` semantics changed from "per-event message count" to "cumulative conversation turns". This is a bug fix since the old semantics were broken for DM; a changelog note is included for beta.11 users.
+**⚠️ Breaking change**: `extractMinMessages` semantics changed from "per-event message count" to "cumulative conversation turns". Before: each `agent_end` needed ≥N messages. After: smart extraction triggers at conversation turn N. This is a bug fix since the old semantics were structurally broken for DM; users relying on the old behavior may need to adjust their `extractMinMessages` values.
 
 ---
 
--- a/index.ts
+++ b/index.ts
@@ -2746,18 +2746,10 @@ const memoryLanceDBProPlugin = {
 
           const priorRecentTexts = autoCaptureRecentTexts.get(sessionKey) || [];
           let texts = newTexts;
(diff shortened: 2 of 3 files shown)


win4r:
@claude

rwmjhb:
## Review Summary

**Verdict: REQUEST-CHANGES** | Value: 54% | Size: MEDIUM

### What This PR Does

Fixes single-turn DM conversations falling through to regex fallback instead of smart-extracted memories, by making `autoCaptureSeenTextCount` cumulative and adding a channel-based fallback key when `conversationId` is null.

### Must Fix

1. **F3 — DM test doesn't test the fix**: The new DM test uses a non-null `conversationId`, so it never exercises the actual `null → channel` fallback path this PR is supposed to fix.
2. **EF1 — Build failure**: Build fails with no captured error — the artifact is unproducible.
3. **EF2 — stripEnvelopeMetadata failures**: 2 assertions fail — `stripEnvelopeMetadata` does not strip subagent runtime wrapper lines.
4. **EF3 — Stale base**: Branch has diverged from main; please rebase.

### Scope Drift ⚠️

The PR description says "Phase 1 cleanup only / pure cleanup fixes with no semantic changes", but the diff contains **all 6+ semantic fixes from PR #518** including DM key fallback and cumulative counting. Commit `fdb740e` ("re-apply all 7 fixes for issue #417") directly contradicts the Phase 1 framing. Please clarify:
- Is PR #518 superseded by this PR, or should both land?
- Why does the description scope differ from the actual diff?

### Nice to Have

- `autoCapturePendingIngressTexts.delete` was silently removed — stale messages may be re-consumed on subsequent `agent_end` events.
- Deduplication else-if branch appears unreachable after the new incremental slicing logic.
- Stale inline comment claims the removed delete "restores event-scoped behavior" — contradicts actual code.
- Auto-capture state is never cleaned on `session_end`, so cumulative behavior can leak across session restarts.

### Bottom Line

Good direction — the problem is worth solving. Please fix the DM test to actually cover the null-conversationId path, rebase, and resolve build failures.

---
*Automated review — 7 rounds | Primary: claude | Adversarial: codex*

AliceLJY (changes requested):
## Review: fix: Phase 1 review fixes for PR #518 (issue #417)

### Assessment

PR #534 claims to be "Phase 1 cleanup fixes" for PR #518, addressing two of rwmjhb's review comments:

1. **`createMockApi` 5th parameter** — Fixed correctly. The function signature now accepts `pluginConfigOverrides = {}` and spreads it into `pluginConfig`. This resolves the silent argument drop that rwmjhb flagged.

2. **`isExplicitRememberCommand` guard removal** — The reasoning is sound: under the REPLACE strategy introduced in PR #518, `texts = pendingIngressTexts` (which grows with each turn), so the `texts.length === 1` condition is effectively unreachable in multi-turn DM. Removing dead code is appropriate.

Both changes are mechanically correct.

### Problem: This is not an incremental fix PR — it is PR #518 rebased

Diffing PR #518 and PR #534 against master reveals they share **the exact same base diff** in `CHANGELOG.md`, `index.ts`, and `test/smart-extractor-branches.mjs`. PR #534 is not a follow-up commit on top of #518 — it is a full copy of #518's changes with the two fixes folded in. This means:

- **PR #534 carries forward all unresolved issues from #518** that rwmjhb flagged as "Must Fix"
- The PR description saying "pure cleanup fixes with no semantic changes" is misleading — this PR includes the entire feature (cumulative counting, DM key fallback, threshold change, etc.)

### Unresolved issues from rwmjhb's review (still present in this PR)

**1. Semantic confusion / monotonic increment (Must Fix #1 — not addressed)**

`currentCumulativeCount = previousSeenCount + eligibleTexts.length` grows monotonically across the session lifetime with no reset. After passing the `extractMinMessages` threshold, **every subsequent `agent_end` event triggers smart extraction**. This is wasteful and potentially harmful — repeated LLM extraction calls on overlapping conversation windows.

rwmjhb explicitly called this a "Must Fix". This PR does not address it. The PR description lists this under "Phase 2 Discussion Points" as item #1, but it should be resolved before merge since it was a blocking concern.

**2. Build failure (Must Fix #2 — not addressed)**

rwmjhb reported TypeScript compilation failure. The PR description does not mention whether this is resolved. CI shows `cli-smoke` passing, but that may not cover `tsc` strictly.

**3. `autoCapturePendingIngressTexts.delete(conversationKey)` removal (Question — not addressed)**

The original code deleted consumed pending texts after reading them. PR #518 (and this PR, carrying the same diff) removes this deletion. rwmjhb and AliceLJY both asked about this. Without the delete:
- `pendingIngressTexts` persists across `agent_end` events for the same conversation key
- Combined with the REPLACE strategy (`newTexts = pendingIngressTexts`), the same pending texts could be consumed on every `agent_end` until overwritten by new `message_received` events
- The comment says "[Fix #4] (delete) restores original behavior where pending is event-scoped" but the delete is still absent in the code

This is a data correctness concern that needs explicit resolution.

### Minor: `pluginConfigOverrides` spread position

```javascript
pluginConfig: {
  dbPath,
  ...
  extractMinMessages: 2,
  ...pluginConfigOverrides,  // <-- spread here
  embedding: { ... },        // <-- always wins over overrides
```

The `embedding` object comes after the spread, so `pluginConfigOverrides` cannot override embedding config. This is likely intentional for tests, but worth a comment noting that `embedding` is deliberately non-overridable.

### Recommendation

This PR should not merge as-is because it carries the full PR #518 diff with 2 of 3 "Must Fix" items still unresolved. Options:

1. **Restructure**: Make this a true incremental PR — commits on top of PR #518's branch, fixing only the `createMockApi` signature and dead code removal. Then #518 addresses the remaining issues separately.
2. **Supersede #518**: If this PR is meant to replace #518 entirely, then resolve all three "Must Fix" items here before merge.

Either way, the monotonic increment and missing `delete` issues need resolution before this lands on master.

jlin53882:
This PR has been superseded by PR #549, which resolves all blocking concerns from the PR #534 review (Must Fix #1, #2, #3, and Minor item #4) and additionally corrects a placement bug in Fix #1 identified during adversarial review.

Summary of fixes in #549:
- **Fix #8**: `pendingIngressTexts.delete()` — clears consumed pending texts
- **Fix #9**: `currentCumulativeCount` reset moved inside success block — resets only on successful extraction (not on failure), preventing re-trigger loops
- **Fix #4**: `pluginConfigOverrides` comment explaining embedding override behavior
Hi @AliceLJY 
Please review #549 instead.

Issue #474, opened 2026-04-03
Feature request: per-agent auto-recall control


itsmiso-ai:
### Problem

When autoRecall fires for an agent using a local LLM (e.g. qwen9b on llama-memory), it can cause timeouts if the model is slow or under load. Different agents may use different LLM backends — some fast (MiniMax API), some slow (local qwen9b).

Currently  is a global setting — it applies to all agents equally. This makes it impossible to enable auto-recall for fast API-based agents while disabling it for slow local-model agents.

### Proposed Solution

Allow per-agent override of the autoRecall setting via the scopes.agentAccess config:

Option A — per-agent boolean flag:
```json
{
  "autoRecall": true,
  "autoRecallAgents": ["sage", "maple"],  // agents where auto-recall is DISABLED
  "autoRecallMinLength": 15,
  ...
}
```

Option B — scope-based filtering:
Use the existing scope system to control which memories trigger auto-recall per agent (auto-recall only fires for memories in scopes the agent has access to).

Option C — per-agent LLM timeout:
Allow  per agent, so slow local models get a longer timeout before auto-recall is skipped.

### Use Case

- Sage: local qwen9b via llama-memory — **disable auto-recall** (times out)
- Maple: MiniMax API — enable auto-recall with standard settings
- Miso: MiniMax API — enable auto-recall with standard settings
- Saffron: MiniMax API — enable auto-recall with standard settings

### Workaround

Currently working around by setting `autoRecall: false` globally and relying on manual `memory_recall` calls.

### Environment

- memory-lancedb-pro: v1.1.0-beta.8
- OpenClaw: 2026.4.1
- Local LLM: qwen9b on llama-memory (slow, times out at 30s)

AliceLJY:
这个需求是合理的，当前确实只有全局 `autoRecall: true/false`，没有按 agent 做覆盖，所以混合后端场景会比较别扭。

先记两点结论：
- 现在没有 per-agent auto-recall 开关，临时 workaround 还是全局关掉后，对兼容 agent 手动用 `memory_recall`。
- 如果要做兼容性更好的增量设计，我更倾向先上一个最小版本，比如 `autoRecallExcludeAgents` / `autoRecallIncludeAgents` 这类 agent-id 级控制，而不是一开始就把 scope/filter/policy 全铺开。

如果你愿意提 PR，建议先按“最小可用配置 + 向后兼容默认值 + 测试覆盖 mixed-agent 场景”这个方向来。

PR #535, opened 2026-04-05
feat(memory): add per-agent auto-recall control

itsmiso-ai:
## PR #535 — per-agent auto-recall control

Implements include/exclude lists for auto-recall injection, keyed by agentId.

### Changes

- **index.ts** (~line 2300): Removed dead `if (agentId !== undefined)` guard.
  `resolveHookAgentId()` always returns a string (falls back to 'main').
  Filter logic now checks include/exclude directly without the unreachable wrapper.
- **openclaw.plugin.json** (~line 1385): Updated help text — removed 'Cannot be used together',
  replaced with accurate precedence statement (whitelist wins).
- **test/per-agent-auto-recall.test.mjs** (~line 156): Test case changed from
  `agentId: undefined` to `agentId: 'main'` — reflects what production actually passes.

### Status

Pending push of commit fc0adf0 (local). Reviewer feedback addressed.


win4r:
@claude

rwmjhb:
## Review Summary

**Verdict: REQUEST-CHANGES** | Value: 63% | Size: LARGE

### What This PR Does

Adds per-agent auto-recall control (`autoRecallIncludeAgents` / `autoRecallExcludeAgents`) to the memory-lancedb-pro plugin, enabling selective memory injection for specific agents (e.g., disabling recall for background workers or local-LLM agents).

### Must Fix

1. **F2 / EF2 — Tests unconfirmed**: Full test suite timed out; the new 19 tests in `test/per-agent-auto-recall.test.mjs` were never confirmed passing.
2. **EF3 / MR1 — Tests not wired in**: The new test file was excluded from the targeted test run and is not wired into `npm test`. Please add it to the test configuration.
3. **EF1 — Build failure**: Build reports `BUILD_FAILURE` but `build_error` is null — cannot determine if regression or pre-existing.

### Nice to Have

- Schema descriptions claim `include` and `exclude` lists "Cannot be used together", but the runtime silently accepts both with whitelist-wins precedence. Consider warning or rejecting at config parse time.
- `agentId` is never undefined at runtime (unresolvable agents fall back to `'main'`), so the undefined guard is dead code.
- Configured agent IDs are filtered but not trimmed — `" myAgent "` won't match `"myAgent"`.
- Mixed-agent scenario tests exercise a local JS stub, not the real `before_prompt_build` hook.

### Bottom Line

Good feature — solves a real need. Please wire the new tests into `npm test`, confirm they pass, and resolve the build failure.

---
*Automated review — 6 rounds | Primary: claude | Adversarial: codex*

AliceLJY (approved):
## Review: feat(memory): add per-agent auto-recall control

**Verdict: Approve**

### What I checked

1. **Backward compatibility** — Neither `autoRecallIncludeAgents` nor `autoRecallExcludeAgents` is required; when both are absent, all agents receive auto-recall as before. No breaking change.

2. **Schema alignment** — `autoRecallExcludeAgents` was already implemented in TypeScript logic but missing from `openclaw.plugin.json` configSchema. This PR fixes that gap and adds the new `autoRecallIncludeAgents` alongside it. Both schema entries use `{ type: "array", items: { type: "string" }, default: [] }`, consistent with the rest of the schema.

3. **Precedence logic** — The `before_prompt_build` hook correctly implements: include list (if non-empty) takes precedence over exclude list. When `agentId` is undefined, the filter block is skipped entirely, which is the right default (no agent context = allow recall). This is well-documented in code comments.

4. **Config parsing** — `parsePluginConfig` filters both arrays to reject non-string and whitespace-only entries, matching the existing `autoRecallExcludeAgents` pattern. The empty-array-stays-as-empty-array behavior is correct and the runtime checks `length > 0` before activating.

5. **Tests** — 19 tests covering:
   - Parsing edge cases (undefined, valid array, non-string filtering, whitespace, empty array, single ID)
   - Runtime logic simulation via a faithful `shouldInjectMemory` helper (whitelist mode, blacklist mode, precedence, no config, undefined agentId, empty include fallthrough)
   - Uses real `parsePluginConfig` via jiti, not mocks

6. **CI** — cli-smoke pass, version-sync pass.

### Minor note (non-blocking)

The description says "Cannot be used together with autoRecallExcludeAgents" but the parser accepts both and the runtime silently picks include over exclude. This is fine for UX (no crash), but consider adding a startup warning log when both are configured, to help users catch misconfiguration. Not a blocker.

Closes #474. LGTM.

rwmjhb (changes requested):
The gap is real — `autoRecallExcludeAgents` being implemented but not exposed in the schema is a straightforward fix, and per-agent control is useful. A few items to resolve first.

**Must fix before merge:**

- **MR1: The new feature tests are not wired into `npm test`.** The test file exists but is not registered in the test manifest — CI will never execute these tests. Please add them to the test runner configuration.

- **MR3: Plugin manifest/UI documentation is incomplete for the new settings.** Users need to be able to discover and configure these options. Please add schema entries and descriptions for all new config fields.

- F3: Dead-code guard — `agentId` is never `undefined` at the check point; unresolvable agents silently fall back to `'main'` and get blocked by the whitelist, which may be surprising behavior.
- F4: Schema descriptions claim mutual exclusivity between `autoRecallAgents` and `autoRecallExcludeAgents`, but runtime accepts both simultaneously with silent precedence — the behavior should be documented or enforced.
- MR2: Configured agent IDs are matched without trimming — a trailing space in config causes a silent mismatch.

Fix MR1 and MR3, then this is ready.

itsmiso-ai:
Both must-fix items are now resolved:

**MR1** — test/per-agent-auto-recall.test.mjs added to npm test script (inserted after test/functional-e2e.mjs in the chain).

**MR3** — Added uiHints entries for both config fields:
- autoRecallExcludeAgents -> label Auto-Recall Excluded Agents
- autoRecallIncludeAgents -> label Auto-Recall Included Agents

Commit: 868b0fb. CI should now pick up the new test on this run.

itsmiso-ai:
Formatting fix applied: openclaw.plugin.json restored to original compact JSON style (single-line enums, single-line number objects). Added uiHints for both autoRecallExcludeAgents and autoRecallIncludeAgents in that format.

Commit: 84bd836

rwmjhb (changes requested):
Thanks — per-agent auto-recall control is a real gap, and filling in the missing `autoRecallExcludeAgents` schema is a nice bonus. A few things to address before merge.

## Must fix

**MR1 — New feature tests aren't wired into `npm test`**

`test/per-agent-auto-recall.test.mjs` (19 tests) isn't reached by the suite runner — the verifier's targeted run excluded it, and the full run timed out before getting there. The PR description claims "All 19 tests pass" but CI can't confirm that. Please make sure this file is executed by the default `npm test` entry point (or documented runner) so regressions are caught.

## Nice to have

- **F3 — Dead-code guard + silent 'main' fallback** (`index.ts:2318`). `resolveHookAgentId()` always returns a string (terminal `|| 'main'`), so `if (agentId !== undefined)` can't be false. More importantly, with `autoRecallIncludeAgents: ['saffron']`, the default/primary agent (which resolves to `'main'`) is silently blocked unless the user knows to add `'main'` to the list. Please:
  1. Document the `'main'` fallback in the schema description.
  2. Remove the dead `if (agentId !== undefined)` wrapper.
  3. Update the `agentId=undefined` test to use `agentId='main'` — that's what production actually passes.

- **F4 — Schema description contradicts itself** (`openclaw.plugin.json:171`). The field claims "Cannot be used with autoRecallIncludeAgents" in one line and "Overrides autoRecallExcludeAgents if both are set" in the next. Runtime accepts both and whitelist wins. Suggest: drop the "Cannot be used" language, replace with *"If autoRecallIncludeAgents is set, it takes precedence over autoRecallExcludeAgents (whitelist wins)."*

- **F5 — Mixed-agent tests use a local JS stub, not the real hook.** `describe('mixed-agent scenarios')` mirrors the TS logic in a stub `shouldInjectMemory`. If `before_prompt_build` gets refactored, the stub keeps passing while the real feature breaks. Please add at least one test that fires a synthetic `before_prompt_build` against the real plugin with a mock `ctx`/`api`.

- **MR2 — Agent IDs aren't trimmed** before comparison. Trailing whitespace in config causes silent mismatches. Normalize (trim + case) at parse time.

- **MR3 — Manifest/UI documentation is incomplete** for the two new settings — users configuring via UI won't know the precedence rules.

## Evaluation notes (stale-base, lower weight)

- **EF1** — Build status came back `fail` with a null error message; likely masked by the 180s test timeout. Please confirm locally that `npm run build` is clean on current main.
- **EF2 / EF3** — The new test file was not part of targeted tests and the full suite timed out, so the test-pass signal is unverified. Rebasing + running the new file in isolation (`node --test test/per-agent-auto-recall.test.mjs`) would close the gap.

## Open questions

- Is issue #474 maintainer-acknowledged, or was it filed by the PR author? Just want to confirm the feature direction is blessed.
- Should combining include + exclude be a warning (or error) at config load, rather than silent precedence? That would match the "Cannot be used together" framing users likely expect.

---

Verdict: **request-changes** (value 0.63, confidence 0.90). Good direction — test wiring + the `'main'` fallback are the main blockers.

rwmjhb (changes requested):
Thanks for pushing updates. Re-reviewed against the new head — same `must_fix` remains, so blocking again. Walking through what's unchanged and what's new below.

## Must fix (still open)

**MR1 — New feature tests aren't wired into `npm test`**

The targeted test run doesn't include `test/per-agent-auto-recall.test.mjs`, and the full-suite run timed out before reaching it. Net effect: the PR description claims "All 19 tests pass" but CI can't confirm that. Please wire the file into whatever discovers tests for `npm test` (or document the runner) so regressions are caught automatically.

## Nice to have

- **F3 — `agentId !== undefined` is a dead guard + silent `'main'` fallback** (`index.ts:2318`). `resolveHookAgentId()` always returns a string (`|| 'main'`), so the new `if (agentId !== undefined)` can't be false. More importantly, with `autoRecallIncludeAgents: ['saffron']`, the default/primary agent silently falls back to `'main'` and gets blocked unless the user knows to add `'main'` to the whitelist. Please:
  1. Document the `'main'` fallback in the schema description
  2. Drop the dead `if (agentId !== undefined)` wrapper
  3. Update the `agentId=undefined` test to `agentId='main'` (that's what production actually passes)

- **F4 — Schema description contradicts itself** (`openclaw.plugin.json:171`). One line says "Cannot be used with autoRecallIncludeAgents", the next says "Overrides autoRecallExcludeAgents if both are set." Runtime accepts both and whitelist wins. Suggest replacing both descriptions with: *"If `autoRecallIncludeAgents` is set, it takes precedence over `autoRecallExcludeAgents` (whitelist wins)."*

- **F5 — Mixed-agent tests exercise a local JS stub, not the real hook.** `describe('mixed-agent scenarios')` defines a local `shouldInjectMemory` that mirrors the TS logic. If `before_prompt_build` gets refactored, the stub keeps passing while the real feature breaks. Please add at least one test that fires a synthetic `before_prompt_build` against the real plugin with a mock `ctx`/`api`.

- **MR2 — Agent IDs aren't trimmed** before comparison. Trailing whitespace in config causes silent mismatches — normalize (trim) at parse time.

- **MR3 — Manifest/UI documentation is still incomplete** for the two new settings.

## Evaluation notes (stale-base, lower weight)

- **EF1** — Build status came back `fail` with a null error message; likely masked by the 180s test timeout. Please confirm locally that `npm run build` is clean on current main.
- **EF2 / EF3** — Full suite timed out and the new test file wasn't in targeted tests, so the "19/19 passing" claim is unverified from CI. Rebase + `node --test test/per-agent-auto-recall.test.mjs` in isolation would close the gap.

## Open question

- Should combining include + exclude be a warning (or error) at config load, rather than silent precedence? That matches the "Cannot be used together" framing users likely expect.

---

Verdict: **request-changes** (value 0.63, confidence 0.90). Blocker is still test wiring + the `'main'` fallback behavior.

itsmiso-ai:
Addressed in latest head `eb64929` / `eb649290049fbd00c0ae4ed4f195e7a8e92ea0ca`:

- **MR1**: wired `test/per-agent-auto-recall.test.mjs` into both `npm test` and the CI manifest verifier (`scripts/ci-test-manifest.mjs` + `scripts/verify-ci-test-manifest.mjs`).
- **F3**: documented the production `'main'` fallback behavior in schema/help text and updated the test case to reflect real runtime behavior.
- **F4**: removed the contradictory mutual-exclusion wording; descriptions now match runtime precedence (include/whitelist wins).
- **F5**: added a real `before_prompt_build` hook test with a plugin/API harness instead of only relying on the local JS stub.
- **MR2**: trim agent IDs at parse time for both include/exclude lists.
- **MR3**: completed schema/UI help coverage for the new settings.

Validation I ran locally:
- `node --test test/per-agent-auto-recall.test.mjs` ✅
- `node scripts/verify-ci-test-manifest.mjs` ✅
- `npm test` now reaches the new test file, so the wiring gap is closed ✅

One note on **EF1**: this repo does not currently define an `npm run build` script, so there wasn't a build command to validate locally. Also, the remaining `npm test` failure I saw is a pre-existing failure on `origin/master` as well (`test/recall-text-cleanup.test.mjs`), not introduced by this PR.

PR #547, opened 2026-04-06
fix: rebuild smart metadata on memory_update normal path (#544)

### LINKED ISSUE #544: Bug: memory_update updates text but leaves l0/l1/l2 metadata stale
## Summary

`memory_update` reports success and updates `entry.text`, but on the normal update path it does **not** refresh derived smart metadata fields:

- `l0_abstract`
- `l1_overview`
- `l2_content`

As a result, `memory_recall` / `memory-pro search` can still surface the **old summary text** even after a successful update.

## Environment

- Plugin: `memory-lancedb-pro`
- Version: `1.1.0-beta.10`
- OpenClaw host: macOS arm64
- Retrieval mode: hybrid
- Confirmed on a real OpenClaw deployment after plugin load + store/recall/update/forget smoke test

## Reproduction

1. Store a memory:
   - text: `Smoke test marker ... Token: MLDP-SMOKE-20260406-1630. Purpose: verify store recall update forget round-trip.`
2. Recall it successfully with the token query.
3. Run `memory_update` on that memory id with new text:
   - `Smoke test marker ... UPDATED. Token: MLDP-SMOKE-20260406-1630. Status: update path verified before forget.`
4. Inspect via `memory-pro list` / `memory-pro search` / `memory_recall`.

## Actual behavior

- `entry.text` is updated.
- But metadata remains stale, for example:
  - `metadata.l0_abstract` still contains the old text
  - `metadata.l1_overview` still reflects the old summary
  - `metadata.l2_content` still contains the old text
- Recall/search surfaces the stale summary, so the tool-level success response is misleading.

Observed example from `memory-pro list --json` after update:

- `text`: `... UPDATED ...`
- `metadata.l0_abstract`: old pre-update content
- `metadata.l1_overview`: old pre-update content
- `metadata.l2_content`: old pre-update content

## Expected behavior

On a successful non-supersede `memory_update`, the plugin should either:

1. Regenerate / patch `l0_abstract`, `l1_overview`, `l2_content` to stay consistent with the new text, or
2. Clearly document that `memory_update` is raw-text-only and that recall/search summaries may remain stale.

I think option 1 is the correct behavior because recall/search ranking/display relies on these fields.

## Likely cause

The normal update path appears to only patch `text` / `vector` / `importance` / `category` and then call `context.store.update(...)`, without rebuilding smart metadata.

Relevant code path:

- `src/tools.ts` normal update flow around the `memory_update` handler
- In my local copy, the path looks like:
  - build `updates` from `text`, `vector`, `importance`, `category`
  - call `context.store.update(resolvedId, updates, scopeFilter)`
  - return success
- I do **not** see metadata regeneration on that non-supersede branch

By contrast, other paths do explicitly build smart metadata.

## Impact

This can cause:

- stale recall output after a successful update
- stale search snippets / summaries
- ranking behavior based on outdated derived fields
- operators trusting `memory_update` success when the memory is only partially updated

## Workaround

For now, the safe workaround is:

- `memory_forget` old memory
- `memory_store` new memory

instead of relying on `memory_update` when summary freshness matters.

## Optional fix direction

On the normal `memory_update` path, if `text` changes:

- parse existing smart metadata
- rebuild / patch `l0_abstract`, `l1_overview`, `l2_content`
- then persist both text/vector and metadata together

Even a simple fallback like:

- `l0_abstract = text`
- `l2_content = text`
- `l1_overview = oldMeta.l1_overview || \`- ${text}\``

would already avoid stale-text inconsistency.

AliceLJY:
## Summary

Fixes #544. The normal update path in `memory_update` was updating `entry.text` and `vector` but leaving smart metadata (`l0_abstract`, `l1_overview`, `l2_content`, `fact_key`, `memory_temporal_type`, `valid_until`, `confidence`) stale.

### Changes
- Hoist `existing` entry fetch before the temporal supersede guard so both paths can reuse it
- When text changes on the normal update path: rebuild l0/l1/l2, fact_key, temporal classification, and expiry
- When only importance changes: sync `metadata.confidence`
- Explicitly clear `valid_until` when new text has no temporal expression (pass `0` to override `buildSmartMetadata` keep-existing behavior)
- No LLM calls added -- uses the same simple pattern as `memory_store`

### Test plan
- [x] New test: text change refreshes l0/l1/l2
- [x] New test: text change refreshes fact_key
- [x] New test: text change refreshes temporal_type/valid_until
- [x] New test: importance-only change syncs confidence
- [x] New test: text unchanged -> metadata preserved
- [x] Regression: supersede path unaffected
- [x] Existing memory-update-supersede tests pass
- [x] Full test suite passes (all 35 test files)


rwmjhb:
## Review: `fix: rebuild smart metadata on memory_update normal path (#544)`

The problem is real and worth fixing — stale `l0_abstract` / `l1_overview` / `fact_key` after `memory_update` is a genuine data consistency bug that silently degrades retrieval quality. However, the current implementation has three blocking issues that need to be resolved before this can merge.

### Must Fix

**MR1 — Referenced symbols do not exist in the repository**
The new handler code calls `inferExpiry(text)` and `classifyTemporal(text)`, and the new test imports `../src/temporal-classifier.ts`. None of these exist in the codebase. This will break compilation and/or test execution. Please either:
- Remove the calls and use the correct existing helpers (e.g. the temporal classification logic already in `buildSmartMetadata`)
- Or add `temporal-classifier.ts` as part of this PR if it is genuinely required

**MR2 — `valid_until: 0` sentinel is not handled by `buildSmartMetadata`**
`tools.ts` uses `valid_until: newExpiry ?? (0 as any)` with the assumption that `0` signals "clear this field". Inspecting `src/smart-metadata.ts`, `buildSmartMetadata` spreads `...patch` directly without normalizing `valid_until: 0` to `undefined`. Passing `0` will store a Unix epoch timestamp (1970-01-01) rather than clearing the field — this could corrupt temporal state for updated entries. Please either:
- Add explicit normalization in `buildSmartMetadata`: `valid_until: patch.valid_until || undefined`
- Or use `undefined` instead of `0` and handle the optional field cleanly

**EF1 — TypeScript compilation not verified**
The production handler changes in `tools.ts` (including the `(0 as any)` cast) were never compiler-checked. Please run `tsc --noEmit` locally and confirm no type errors before resubmitting.

### Nice to Have

- **F2**: `simulateMemoryUpdate` in the test file reimplements the fix logic rather than calling the actual registered MCP handler. A regression in `tools.ts` would not be caught by this test. Consider calling the real handler via the MCP test harness that other test files already use.
- **F3**: `existing` is only fetched when both `text` AND `newVector` are truthy — if a caller provides updated text but relies on the existing vector, `existing` remains `null` and metadata is still not rebuilt. Consider whether this is an expected caller pattern that should also be covered.
- **MR3**: The importance-only branch uses legacy category names (`preference`, `fact`) where smart-category names (`preferences`, `entities`) are required for `fact_key` regeneration.
- **EF4**: Stale base on a large `src/tools.ts` diff — rebase before resubmitting to reduce merge-conflict risk.

---

Good direction overall. Fix the missing symbols, the `valid_until` sentinel contract, and confirm compilation — then this is ready for another pass.

rwmjhb (changes requested):
Stale smart metadata after `memory_update` is a genuine correctness bug and rebuilding it on the normal path is the right fix. Two blockers to resolve first.

**Must fix before merge:**

- **MR1: The patch references symbols and a test module that do not exist in the repository.** The diff imports temporal helper symbols and a test utility module that are not present in the codebase. The TypeScript build will fail as a result. Please verify imports resolve correctly before requesting re-review.

- **MR2: The claimed `valid_until` clear-on-zero contract is not implemented by `buildSmartMetadata`.** The PR relies on passing `0` as a sentinel to clear `valid_until`, but `buildSmartMetadata` has no such contract — the `0` will be stored as-is or ignored. The clear logic needs to be explicitly implemented.

- **F3: When `text` is provided without `newVector`, `existing` is `null` at the metadata rebuild step** — the rebuild is silently skipped and metadata stays stale. This is the most common update scenario.

- F1: The `(0 as any)` cast for `valid_until` bypasses TypeScript's type contract — the sentinel approach should be typed properly once MR2 is implemented.
- EF4: Stale base on a `LARGE` PR touching `src/tools.ts` — please rebase before re-requesting review.

Fix MR1, MR2, and F3, then this is ready.

rwmjhb:

**Please**:

3. Push the rebased branch

AliceLJY:
@rwmjhb Rebased onto master + addressed the review.

**EF4** (rebase) — done, clean on master (cf782a2).

**MR1** — verified all imports resolve on current master:
- `MemoryEntry` from `./store.js`
- `classifyTemporal`, `inferExpiry` from `./temporal-classifier.js`
- `TEMPORAL_VERSIONED_CATEGORIES` from `./memory-categories.js`
- `deriveFactKey`, `buildSmartMetadata`, `parseSmartMetadata`, `stringifySmartMetadata` from `./smart-metadata.js`

All of these exist in the repo today (grep confirms). If you still see unresolved symbols, please point me at the specific import line — happy to dig.

**MR2** (valid_until sentinel) — refactored to explicit post-assignment:
```ts
const updatedMeta = buildSmartMetadata(existing, { ... });
updatedMeta.valid_until = inferExpiry(text);
```
No more `(0 as any)` cast, no sentinel contract on `buildSmartMetadata`. Absence of a new expiry cleanly clears the stale value.

**F3** (null existing at rebuild) — hoisted the `getById` so `existing` is fetched once at the top whenever `text || importance !== undefined`. The temporal supersede guard and the normal-path rebuild now share the same reference, and the rebuild block's `if (text && existing)` can't silently skip on a non-temporal category text update.

**F1** (nit) — dropped the `as any` cast as part of the MR2 fix.

Local tests:
- `test/memory-update-metadata-refresh.test.mjs` ✅ (all 6 cases)
- `test/memory-update-supersede.test.mjs` ✅ (4 cases, regression intact)

Ready for another look when you have time.

rwmjhb (approved):
Approved based on orchestrator review results. Non-blocking follow-ups were noted separately; no merge-blocking issues from this review.

PR #552, opened 2026-04-06
fix: move backup timer setup to register() for late-loading plugins

### LINKED ISSUE #551: Bug: Auto-backup fails when plugin registers after gateway startup (service.start() not called for late-loading plugins)
## 🐛 Problem Summary

The memory-lancedb-pro plugin's auto-backup feature stops working when the plugin registers **after** the gateway has already completed startup. This is because the backup timer is only set up in the service's `start()` method, which is never called for late-loading plugins.

**Impact**: No daily backups are created, risking memory data loss.

---

### Normal Case (2026-04-03)
| Time | Event |
|------|-------|
| 01:36:58 | Plugin registered (`[gateway]` prefix) |
| 01:37:13 | Gateway startup complete (`listening on ws://...`) |
| 01:37:19 | `initialized successfully` logged (service.start() called) |
| 01:38:19 | Backup completed ✅ |

### Broken Case (2026-04-07)
| Time | Event |
|------|-------|
| 00:39:40 | Gateway startup complete (`ready (5 plugins)`) |
| 00:40:12 | Plugin registered (`[plugins]` prefix, **32s LATE**) |
| ❌ | No `initialized successfully` log (service.start() NOT called) |
| ❌ | No backup created |

**Result**: 5 days without backups (last backup: 2026-04-03).

---

## 🔍 Root Cause Analysis

The backup timer is set up **only** in the service's `start()` method:

```typescript
// index.ts:3749-3849 (BEFORE FIX)
api.registerService({
  id: "memory-lancedb-pro",
  start: async () => {
    // ... startup checks ...
    
    // ❌ PROBLEM: This only runs if gateway calls start()
    setTimeout(() => void runBackup(), 60_000);
    backupTimer = setInterval(() => void runBackup(), BACKUP_INTERVAL_MS);
  },
});
```

**The Issue**:
- Gateway calls `service.start()` only on plugins that are registered **during** gateway startup
- If plugin registers **after** gateway is ready (e.g., on-demand loading by agent session), `start()` is never called
- No backup timer is created → no backups

**Why did this happen?**
- Plugin loading mechanism appears to have changed (log prefix shifted from `[gateway]` to `[plugins]`)
- Plugin now loads on-demand when agent session starts, not during gateway startup
- Gateway does not call `start()` on late-registering plugins

---

## ✅ Temporary Fix (Workaround)

Move backup timer setup from `start()` to `register()`:

```typescript
// index.ts (AFTER FIX)
api.registerService({
  id: "memory-lancedb-pro",
  start: async () => {
    // ... startup checks ...
    // Note: Backup timer is now set in register() to ensure it works
    // even when plugin is registered after gateway startup.
  },
  stop: async () => {
    if (backupTimer) {
      clearInterval(backupTimer);
      backupTimer = null;
    }
    api.logger.info("memory-lancedb-pro: stopped");
  },
});

// CRITICAL FIX: Set backup timer in register() instead of start()
if (!backupTimer) {
  api.logger.info("memory-lancedb-pro: setting up auto-backup (register-time)");
  setTimeout(() => void runBackup(), 60_000);
  backupTimer = setInterval(() => void runBackup(), BACKUP_INTERVAL_MS);
}
```

---

## 🎯 Recommended Official Fix

**Option 1 (Recommended)**: Move backup timer setup to `register()` function
- Ensures backups work regardless of plugin registration timing
- Simple, minimal change
- No breaking changes

**Option 2**: Fix gateway to call `service.start()` on late-registering plugins
- More invasive change (requires gateway modification)
- Better long-term solution for all plugins

---

## 📝 Environment

- Plugin version: 1.1.0-beta.10
- OpenClaw version: Latest (2026-04)
- Node.js: v25.4.0
- OS: Darwin 24.6.0 (arm64)

---

## 🚨 Urgency

**High**: Data loss risk. Users may lose memory data without realizing backups have stopped.

---

## 📎 Logs

Full gateway logs available upon request. Key excerpts:

```
# Broken case (2026-04-07)
2026-04-07T00:39:40.421+08:00 [gateway] ready (5 plugins, 2.2s)
2026-04-07T00:40:12.406+08:00 [plugins] memory-lancedb-pro@1.1.0-beta.10: plugin registered
2026-04-07T00:40:12.409+08:00 [plugins] memory-lancedb-pro: setting up auto-backup (register-time) [AFTER FIX]
2026-04-07T00:43:44.470+08:00 [plugins] memory-lancedb-pro: backup completed (308 entries) [AFTER FIX]
```

Tony-ooo:
## 🐛 Problem

The auto-backup feature stops working when the plugin registers **after** the gateway has completed startup. This happens because the backup timer is only set up in `service.start()`, which is never called for late-loading plugins.

**Impact**: No daily backups are created, risking memory data loss.

---

## 🔍 Root Cause

- Gateway calls `service.start()` only on plugins registered **during** gateway startup
- If plugin registers **after** gateway is ready (e.g., on-demand loading), `start()` is never called
- No backup timer is created → no backups

---

## ✅ Solution

Move backup timer setup from `service.start()` to `register()`:

```typescript
// CRITICAL FIX: Set backup timer in register() instead of start()
if (!backupTimer) {
  api.logger.info('memory-lancedb-pro: setting up auto-backup');
  setTimeout(() => void runBackup(), 60_000);
  backupTimer = setInterval(() => void runBackup(), BACKUP_INTERVAL_MS);
}
```

**Benefits**:
- Ensures backups work regardless of plugin registration timing
- Simple, minimal change with no breaking changes
- Idempotent guard prevents duplicate timers

---

## 📝 Testing

Verified on local environment:
- ✅ Backup timer created immediately on plugin registration
- ✅ Initial backup completed ~65 seconds after registration
- ✅ Daily backup schedule established
- ✅ No duplicate timers on multiple registrations

---

## 📊 Evidence

See issue for full timeline comparison and logs.

Fixes #551

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -3605,6 +3605,7 @@ const memoryLanceDBProPlugin = {
     // Auto-Backup (daily JSONL export)
     // ========================================================================
 
+    let backupInitialTimeout: ReturnType<typeof setTimeout> | null = null;
     let backupTimer: ReturnType<typeof setInterval> | null = null;
     const BACKUP_INTERVAL_MS = 24 * 60 * 60 * 1000; // 24 hours
 
@@ -3652,6 +3653,32 @@ const memoryLanceDBProPlugin = {
       } catch (err) {
         api.logger.warn(`memory-lancedb-pro: backup failed: ${String(err)}`);
       }
+    }
+
+    function clearBackupTimers() {
+      if (backupInitialTimeout) {
+        clearTimeout(backupInitialTimeout);
+        backupInitialTimeout = null;
+      }
+      if (backupTimer) {
+        clearInterval(backupTimer);
+        backupTimer = null;
+      }
+    }
+
+    function setupBackupTimers() {
+      if (backupInitialTimeout || backupTimer) return;
+
+      api.logger.info("memory-lancedb-pro: setting up auto-backup");
+
+      backupInitialTimeout = setTimeout(() => {
+        backupInitialTimeout = null;
+        void runBackup();
+      }, 60_000);
+      backupInitialTimeout.unref?.();
+
+      backupTimer = setInterval(() => void runBackup(), BACKUP_INTERVAL_MS);
+      backupTimer.unref?.();
     }
 
     // ========================================================================
@@ -3750,23 +3777,13 @@ const memoryLanceDBProPlugin = {
         // even when plugin is registered after gateway startup.
       },
       stop: async () => {
-        if (backupTimer) {
-          clearInterval(backupTimer);
-          backupTimer = null;
-        }
+        clearBackupTimers();
         api.logger.info("memory-lancedb-pro: stopped");
       },
     });
 
-    // CRITICAL FIX: Set backup timer in register() instead of start().
-    // This ensures backups work even when plugin is registered after gateway startup
-    // (e.g., when plugin is loaded on-demand by agent session).
-    if (!backupTimer) {
-      api.logger.info("memory-lancedb-pro: setting up auto-backup");
-      // Run initial backup after a short delay, then schedule daily
-      setTimeout(() => void runBackup(), 60_000); // 1 min after register
-      backupTimer = setInterval(() => void runBackup(), BACKUP_INTERVAL_MS);
-    }
+    // Set backup timers in register() so late-loaded plugins still auto-backup.
+    setupBackupTimers();
   },
 };
 
--- a/index.ts
+++ b/index.ts
@@ -3746,9 +3746,8 @@ const memoryLanceDBProPlugin = {
           }
         }, 5_000);
 
-        // Run initial backup after a short delay, then schedule daily
-        setTimeout(() => void runBackup(), 60_000); // 1 min after start
-        backupTimer = setInterval(() => void runBackup(), BACKUP_INTERVAL_MS);
+        // Note: Backup timer is now set in register() to ensure it works
+        // even when plugin is registered after gateway startup.
       },
       stop: async () => {
         if (backupTimer) {
@@ -3758,6 +3757,16 @@ const memoryLanceDBProPlugin = {
         api.logger.info("memory-lancedb-pro: stopped");
       },
     });
+
+    // CRITICAL FIX: Set backup timer in register() instead of start().
+    // This ensures backups work even when plugin is registered after gateway startup
+    // (e.g., when plugin is loaded on-demand by agent session).
+    if (!backupTimer) {
+      api.logger.info("memory-lancedb-pro: setting up auto-backup");
+      // Run initial backup after a short delay, then schedule daily
+      setTimeout(() => void runBackup(), 60_000); // 1 min after register
+      backupTimer = setInterval(() => void runBackup(), BACKUP_INTERVAL_MS);
+    }
   },
 };
 
--- a/index.ts
+++ b/index.ts
@@ -1615,6 +1615,39 @@ const pluginVersion = getPluginVersion();
 // Plugin Definition
 // ============================================================================
 
+interface PluginLifecycleState {
+  resolvedDbPath: string;
+  token: number;
+  stopped: boolean;
+  backupInitialTimeout: ReturnType<typeof setTimeout> | null;
+  backupTimer: ReturnType<typeof setInterval> | null;
+  startupChecksTimeout: ReturnType<typeof setTimeout> | null;
+  legacyScanTimeout: ReturnType<typeof setTimeout> | null;
+}
+
+let nextLifecycleToken = 1;
+
+const _activeLifecyclesByDbPath = new Map<string, PluginLifecycleState>();
+
+function clearLifecycleTimers(lifecycle: PluginLifecycleState) {
+  if (lifecycle.backupInitialTimeout) {
+    clearTimeout(lifecycle.backupInitialTimeout);
+    lifecycle.backupInitialTimeout = null;
+  }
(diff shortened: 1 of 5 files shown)


AliceLJY (approved):
Clean fix. Moving the timer to `register()` correctly handles the late-loading case, and the `!backupTimer` guard prevents duplication.

Assigning to @rwmjhb for final merge.

rwmjhb:
## Review Summary

The problem is real and high-value (86%) — 5 days of missed backups from late-loading plugins is a genuine data loss risk. Good direction.

### Must Fix

1. **setTimeout return value discarded** — the initial-backup timer (`setTimeout(runBackup, 60_000)`) is not assigned to `backupTimer`, so `stop()` cannot cancel it. If `stop()` is called within the first 60s after `register()`, a backup will still fire after shutdown.

2. **register()-only test harnesses break** — moving timer setup into `register()` means any test that calls `register()` without `stop()` will leave a dangling timer. The core test group already does this — confirm no test hangs or leaks were introduced.

3. **Repeated registrations stack timers** — the `if (!backupTimer)` guard only checks the `setInterval` timer, not the initial `setTimeout`. Multiple `register()` calls (e.g., hot-reload) can stack independent initial-backup callbacks.

4. **Rebase required** — `stale_base=true`. Please rebase onto main and confirm build + tests pass.

### Questions

- Does `runBackup()` depend on state initialized in `start()` (DB handles, config)? If so, the 60s initial timer could race with `start()` on normal startup paths.

### Nice to Have

- `start()` failure no longer prevents backup timer from arming — consider whether this is desirable or whether backup should be gated on successful start.

Good fix direction. Address the uncancellable setTimeout and rebase, then this is ready.

Tony-ooo:
Addressed the review feedback in `fe0b154`.

Fixed:
- Track the initial backup `setTimeout` separately so `stop()` now clears both the initial timeout and the daily interval.
- Call `unref()` on both register-time backup timers so register-only test harnesses can exit cleanly.
- Prevent duplicate initial-backup callbacks on re-register by guarding against either active timer handle.

Validation:
- `node --test test/session-summary-before-reset.test.mjs`

On the startup question: `runBackup()` does not depend on state initialized in `start()`. The store and resolved DB path are created before service registration, so the register-time backup timer does not race `start()` for those dependencies.

The “start failure should/should not arm backup” point is still a design choice; this patch keeps the current behavior of arming backup from `register()` so late-loaded plugins still get backups.

rwmjhb:
## Review: `fix: move backup timer setup to register() for late-loading plugins`

Automated multi-round review (7 rounds, Claude + Codex adversarial). Value: 86% — the problem is real and the log evidence is solid. Good direction overall.

**Verdict: request-changes** (confidence 0.95)

### Must Fix

**1. setTimeout return value discarded — timer uncancellable by stop()**

`index.ts:3763`: `setTimeout(() => void runBackup(), 60_000)` — the handle isn't stored. `stop()` only clears the `backupTimer` interval. If `stop()` is called within 60s of `register()`, `runBackup()` fires after teardown (DB handles may be closed). On hot-reload (stop → re-register), the orphaned timeout plus the new one produces two overlapping initial backups.

Fix: capture the handle and clear it in `stop()`:
```ts
let backupInitTimer: ReturnType<typeof setTimeout> | null = null;
backupInitTimer = setTimeout(...);
// in stop():
if (backupInitTimer) { clearTimeout(backupInitTimer); backupInitTimer = null; }
```

**2. Test suite timeouts — open handle regression**

The `plugin-manifest-regression` test passes assertions but Node doesn't exit within 60s. The full suite times out at 180s — 3 test groups never ran. This is consistent with `register()` unconditionally arming `setInterval`, keeping the event loop open.

Run `--detectOpenHandles` to confirm. Test harness needs to clear timers on plugin unload, or guard timers with a test-mode check.

**3. register-only test harnesses break**

Moving backup setup into `register()` means any test calling `register()` without `start()` now arms timers. The core test group uses register-only harnesses — explaining the timeout pattern.

### Nice to Have

- **start() failure no longer gates backup timer**: If `start()` throws, `runBackup()` fires 60s later against a partially-initialized store. Consider a readiness flag checked by `runBackup()`.
- **Repeated registrations can stack backup loops**: `if (!backupTimer)` guard is per-registration, not process-wide.

### Questions

- Does `runBackup()` depend on state only set up inside `start()` (DB handles, indices)? If so, timer-arm → start() is a new race condition.
- stale_base=true — please rebase before merge.

Tony-ooo:
Thanks — this review is stale against the current branch tip.

Addressed in `a6757d2`:

1. The initial backup timeout is now stored in the lifecycle state and cleared from `stop()`, so it can no longer fire after teardown.
2. All timers created by the register/start lifecycle are now tracked and cleaned up consistently. The backup timeout/interval and the deferred startup/legacy-scan timers are also `unref()`'d.
3. Re-registering the same `dbPath` now replaces the previous runtime via a lifecycle registry, so backup loops do not stack on hot reload / duplicate registration.
4. The register-only harness regressions were fixed by making the affected tests explicitly stop registered services during teardown.

On the start-vs-backup question: `runBackup()` only depends on the `store` created during `register()`. `MemoryStore.list()` initializes lazily, so the backup path does not depend on `start()` having completed.

Validation after the fix:
- `node --test --test-reporter=spec test/plugin-manifest-regression.mjs`
- full `npm test`

Both pass on the current branch.

Also, the branch is currently not behind `origin/master`, so the earlier `stale_base=true` note should no longer apply.

Issue #472, opened 2026-04-02
memory-lancedb-pro appears functional, but openclaw status/doctor reports memory unavailable


lamkan0210:
## Summary

After upgrading OpenClaw to 2026.4.2, `memory-lancedb-pro` appears to be loaded and functioning, but `openclaw status` / `openclaw doctor` still report memory as unavailable or inactive.

This looks like a health-probe / active-plugin-detection mismatch rather than a hard runtime failure.

## Environment

- OpenClaw: 2026.4.2
- OS: macOS 26.5 (arm64)
- Node: 25.6.0
- Memory plugin log shows: `memory-lancedb-pro@1.1.0-beta.10`

## What I observed

Plugin startup logs show the plugin is being registered successfully:

```text
[plugins] memory-lancedb-pro@1.1.0-beta.10: plugin registered (db: /Users/qianglin/.openclaw/memory/lancedb-pro-1024, model: bge-large:latest, smartExtraction: ON)
[plugins] memory-lancedb-pro: smart extraction enabled (LLM model: openai/gpt-oss-120b, timeoutMs: 30000, noise bank: ON)
[plugins] session-memory: hook registered for command:new as memory-lancedb-pro-session-memory
```

But `openclaw status` reports:

```text
Memory: enabled (plugin memory-lancedb-pro) · unavailable
```

And `openclaw doctor --non-interactive` reports:

```text
No active memory plugin is registered for the current config.
```

## Why this seems inconsistent

From the logs, the plugin appears to be:

- loaded
- registered
- hooked into session memory
- configured with a valid DB + embedding model

So the issue seems to be one of:

- active memory plugin detection
- health probe / status reporting
- doctor/status checking a different registration path than the runtime actually uses

## Expected behavior

If the plugin is successfully registered and usable, `openclaw status` / `doctor` should reflect that accurately.

## Actual behavior

Runtime logs suggest the plugin is active, but status/doctor still say memory is unavailable or inactive.

## Extra note

This may be related to the distinction between:

- plugin loaded/registered at runtime
- "active memory plugin" detection used by CLI health/status commands

If useful, I can provide more logs from `openclaw status` / `openclaw doctor` output.

AliceLJY:
Thanks for the report. This looks like the same status/doctor integration gap behind #434, and PR #471 is the right place to fix it.

[...]

Issue #421, opened 2026-03-30
memory-lancedb-pro · unavailable


Bear-Beer:
### Plugin Version

1.1.0-beta.10

### OpenClaw Version

2026.3.28 (f9b1079)

### Bug Description

openclaw status报错截图：![Image](https://github.com/user-attachments/assets/7ef7af57-2d01-4822-935d-0f13c9d49df8)

openclaw配置文件：[openclaw.json](https://github.com/user-attachments/files/26351378/openclaw.json)

CPU：AMD EPYC 9755x2
OS：CentOS 10

### Expected Behavior

online as expected

### Steps to Reproduce

1. install skill
2. install plugin
3. openclaw status --deep

### Error Logs / Screenshots

```shell

```

### Embedding Provider

None

AliceLJY:
Thanks for the report! Could you share a bit more detail to help diagnose?

```bash
openclaw status --deep
openclaw logs --follow   # restart the gateway and watch for plugin-related errors
```

AMD EPYC 9755 should have full AVX-512 support so that's unlikely to be the cause. Most common reasons for `unavailable` are: Jina/embedding API key not set, LanceDB path permission issue, or gateway not restarted after plugin install. The full logs should tell us which one.

mychataa:
我也遇到一樣的問題，我降回 2026.3.24 版就正常了

AliceLJY:
@mychataa Thanks for confirming the pattern. Two users now report that downgrading to OpenClaw 2026.3.24 resolves the issue — this points to a compatibility break in newer OpenClaw versions, not a memory-lancedb-pro bug per se.

Could either of you share the actual error message when using the latest OpenClaw? (logs from `openclaw doctor` or the startup output would be most helpful.) That'll help us pinpoint which OpenClaw change broke it.

In the meantime, staying on 2026.3.24 is a valid workaround.

mychataa:
今天 openclaw 更新到 2026.4.1 還是有問題
 openclaw status --deep

🦞 OpenClaw 2026.4.1 (da64a97) — I keep secrets like a vault... unless you print them in debug logs again.

│
11:36:29+08:00 [plugins] [lcm] Plugin loaded (enabled=true, db=/home/ncc/.openclaw/lcm.db, threshold=0.75)
11:36:29+08:00 [plugins] [lcm] Compaction summarization model: minimax-portal/MiniMax-M2.7 (default)
[...]

Issue #411, opened 2026-03-30
[Bug] openclaw status shows Memory as unavailable but plugin works correctly


zhiangg:
## Bug Description

`openclaw status` shows Memory as `unavailable` but all memory functions work correctly:
- `openclaw memory-pro list` works
- `openclaw memory-pro search` works
- `memory_recall` tool works
- Database read/write is fully functional

## Environment

- OpenClaw version: 2026.3.28 (f9b1079)
- memory-lancedb-pro version: 1.1.0-beta.9
- OS: macOS Darwin 25.3.0

## Steps to Reproduce

1. Install memory-lancedb-pro plugin
2. Run `openclaw gateway restart`
3. Run `openclaw status` → Memory shows `unavailable`
4. Run `openclaw memory-pro list` → works fine
5. Run `openclaw memory-pro search test` → works fine

## Root Cause

The status probe checks `vector.available === false` to show "unavailable", but the plugin reports `vector { enabled: true, available: true }` in its actual state. This appears to be a timing issue where the status probe reports state before plugin fully initializes.

## Expected Behavior

`openclaw status` should show Memory as `ready` when LanceDB connection is functional.

## Workaround

None needed - memory functions work fine, only the status display is misleading.

mbwmbw1337:
I am seeing the same issue as well...

AliceLJY:
Thanks for the detailed writeup and root cause analysis! You're right — this is a timing issue in the status probe. The `vector.available` check can race against plugin initialization and report stale state on slower machines.

In the meantime, `openclaw memory-pro stats` is the reliable health check. Tracking a fix for the status probe to wait for plugin ready signal before reporting. Appreciate the clear reproduction steps!

ldhhp:
我也是同样的问题，不过可以正常工作

AliceLJY:
这条和 #472 / #421 属于同一类问题：插件实际可用，但 `openclaw status` / `doctor` 的状态探测会把 memory 错判成 unavailable。

对应修复已经随 PR #471 合并进主线，所以我先把这条关闭。若你在包含 #471 的最新代码/版本上还能稳定复现，再贴新日志 reopen 即可。

PR #564, opened 2026-04-08
Fix OpenClaw status/doctor runtime integration

YitingOU:
## Summary

This fixes the `openclaw status` / `openclaw doctor` false-negative for `memory-lancedb-pro` on newer OpenClaw builds.

The plugin was occupying the `memory` slot, but it only registered tools / hooks / CLI. OpenClaw's status and doctor flows now expect an active memory runtime manager, so the plugin could be fully functional while still being reported as:

- `Memory: enabled (plugin memory-lancedb-pro) · unavailable`
- `No active memory plugin is registered for the current config`

## Root cause

`memory-lancedb-pro` did not expose a real memory runtime adapter via the OpenClaw memory runtime API.

This PR replaces the previous stub with a minimal but functional runtime adapter that:

- registers through `registerMemoryCapability({ runtime })` when available
- falls back to `registerMemoryRuntime(runtime)` for older hosts
- returns a real memory search manager backed by the existing `store` / `embedder` / `retriever` / `scopeManager`
- reports meaningful status metadata (`files`, `chunks`, `dbPath`, `vector`, `fts`, `cache`, `custom.searchMode`)
- implements `search`, `readFile`, `probeEmbeddingAvailability`, and `probeVectorAvailability`

## Also included

The plugin's dual-memory explainer now uses the existing CLI-aware logger path, so command outputs such as `memory-pro version` stay clean in CLI mode.

## Regression coverage

Added host-level coverage that asserts `openclaw status --json --all` returns a non-null `memory` snapshot when the plugin is installed as the active memory slot.

## Validation

Ran locally:

```bash
node test/openclaw-host-functional.mjs
```

## Related reports

This addresses the same underlying status/runtime mismatch described in:

- #472
- #421
- #411

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -21,9 +21,13 @@ const isCliMode = () => process.env.OPENCLAW_CLI === "1";
 
 // Import core components
 import { MemoryStore, validateStoragePath } from "./src/store.js";
+import type { MemoryEntry } from "./src/store.js";
 import { createEmbedder, getVectorDimensions } from "./src/embedder.js";
+import type { Embedder } from "./src/embedder.js";
 import { createRetriever, DEFAULT_RETRIEVAL_CONFIG } from "./src/retriever.js";
+import type { MemoryRetriever } from "./src/retriever.js";
 import { createScopeManager, resolveScopeFilter, isSystemBypassId, parseAgentIdFromSessionKey } from "./src/scopes.js";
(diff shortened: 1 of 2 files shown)


YitingOU:
Withdrawing this PR for now. We’ll continue carrying and evolving the fix on our own branch/fork first, and can re-open or submit a refreshed PR later if needed.

Issue #555, opened 2026-04-07
[Feature] Per-user long-term memory isolation when multiple users share one agent (e.g. Feishu DM; integrate user: scope with auto-capture/auto-recall)


eightHundreds:
## Background

On channels such as **Feishu (Lark)**, **multiple real users** may **DM the same agent**. We want **long-term memory isolated per user**, so user A’s and user B’s memories are not merged into one shared corpus.

## Current behavior / problem

The plugin already supports multiple scopes (including `user:<id>`) and `scopes.agentAccess`, and tools can pass an explicit `scope`.

However, on **`autoCapture` / `autoRecall` hook paths**, the default read/write scope comes from `getDefaultScope(agentId)`, typically **`agent:<agentId>`**. There is **no** automatic mapping from OpenClaw session context (`channelId`, `conversationId`, channel-specific user ids, etc.) to **`user:<stable-user-id>`**.

So when **one agent id serves many users in DM**, auto-captured memories all land in the same **`agent:*` bucket**, which risks **cross-user bleed**.

## Desired behavior (feature)

When enabled via configuration:

1. Resolve a **per-user scope key** (e.g. `user:<id>`) from host context using **documented, stable fields** (e.g. `ctx.conversationId`, `ctx.channelId`, `accountId`, Feishu `open_id` / `user_id`, or whatever OpenClaw exposes per channel).
2. **`autoCapture` writes** and **`autoRecall` / `before_prompt_build` retrieval** should default to that scope (subject to `scopes` / ACL).
3. **Backward compatibility**: when the feature is off, keep today’s semantics (primarily `agent:<id>`). Policy for `global` and shared buckets can be discussed separately.

## Possible implementation directions (for maintainers)

- Add configuration, e.g. `scopes.perUserDefaultFromContext: { enabled, field, prefix: "user:" }` (exact field names must align with OpenClaw channel contracts).
- Or align with an **OpenClaw-level** “default scope per session” so behavior is not plugin-only.

## Related notes

- Built-in `user:` pattern and `MemoryScopeManager` already exist in this repo.
- Feishu envelope metadata must not be stored as memory (see e.g. #211); that is orthogonal but part of the same channel story.

## Related OpenClaw upstream discussions

These track similar **multi-user / per-session / channel-aware** needs at the host level; linking them here so plugin work can stay aligned with (or delegate to) OpenClaw:

[...]

PR #568, opened 2026-04-09
feat: per-user memory isolation via template scopes

eightHundreds:
## Summary

Implements per-user memory isolation (#555) to prevent cross-user memory bleed when multiple users DM the same bot (e.g., Feishu).

- Add template variable support in `scopes.default` (`user:${accountId}`, `${channelId}`, etc.) — resolved at hook layer, scope system stays purely static
- Add wildcard pattern support in `agentAccess` (`user:*`) for explicit multi-user read access
- Fix SQL wildcard filtering to match application-layer ACL semantics (`_%` pattern)
- Narrow smart extractor dedup scope to `[defaultScope]` instead of `accessibleScopes`

## Configuration Example

```json
{
  "scopes": {
    "default": "user:${accountId}",
    "agentAccess": {
      "bot-1": ["global", "user:*"]
    }
  }
}
```

Each user's memories are written to their own scope (e.g., `user:alice`), while the bot can read across all user scopes via the `user:*` wildcard.

## Key Design Decisions

- **Plan B architecture**: scope system stays purely static, template resolution is a hook-layer concern (`resolveHookDefaultScope`)
- **No auto-wildcard**: `getAccessibleScopes` does NOT auto-append wildcards — wildcards only via explicit `agentAccess` config
- **Backward compatible**: static `scopes.default` works exactly as before, no breaking changes

## Files Changed

- `src/scopes.ts` — template/wildcard utilities, wildcard-aware `isAccessible`, safe `getDefaultScope`
- `index.ts` — `resolveHookDefaultScope()` for hook-layer template resolution
- `src/store.ts` — `scopeFilterToSqlCondition` (SQL LIKE with escaping), `scopeFilterIncludes` (app-layer wildcard matching)
- `test/scope-template-wildcard.test.mjs` — 30 new tests

## Test plan

- [x] All 59 tests pass (29 existing + 30 new)
- [ ] Manual test with Feishu bot: verify user A's memories are isolated from user B
- [ ] Verify backward compatibility: existing static scope configs work unchanged

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -23,7 +23,8 @@ const isCliMode = () => process.env.OPENCLAW_CLI === "1";
 import { MemoryStore, validateStoragePath } from "./src/store.js";
 import { createEmbedder, getVectorDimensions } from "./src/embedder.js";
 import { createRetriever, DEFAULT_RETRIEVAL_CONFIG } from "./src/retriever.js";
-import { createScopeManager, resolveScopeFilter, isSystemBypassId, parseAgentIdFromSessionKey, resolveTemplateScope, hasTemplateVars } from "./src/scopes.js";
+import { createScopeManager, resolveScopeFilter, isSystemBypassId, parseAgentIdFromSessionKey, resolveImplicitWriteScope, hasTemplateVars } from "./src/scopes.js";
+import type { ImplicitWriteScopeResolution } from "./src/scopes.js";
 import { createMigrator } from "./src/migrate.js";
 import { registerAllMemoryTools } from "./src/tools.js";
 import { appendSelfImprovementEntry, ensureSelfImprovementLearningFiles } from "./src/self-improvement-files.js";
@@ -306,26 +307,42 @@ function resolveHookAgentId(
 
 /**
  * Hook-layer template resolution for scopes.default.
- * The scope system stays static — template resolution happens here at runtime.
- * Falls back to scopeManager.getDefaultScope(agentId) when not a template or unresolvable.
+ * Hook-initiated writes must resolve to a concrete readable scope or skip.
  */
 function resolveHookDefaultScope(
   config: PluginConfig,
-  scopeManager: { getDefaultScope(agentId?: string): string },
+  scopeManager: {
+    getDefaultScope(agentId?: string): string;
+    isAccessible(scope: string, agentId?: string): boolean;
+    validateScope(scope: string): boolean;
+  },
   agentId: string,
   ctx: any,
-): string {
-  const tpl = config.scopes?.default;
-  if (tpl && hasTemplateVars(tpl)) {
-    const resolved = resolveTemplateScope(tpl, {
+): ImplicitWriteScopeResolution {
+  return resolveImplicitWriteScope({
+    configuredDefaultScope: config.scopes?.default,
+    scopeManager,
+    agentId,
+    context: {
       agentId,
       accountId: ctx?.accountId,
       channelId: ctx?.channelId,
       conversationId: ctx?.conversationId,
-    });
-    if (resolved) return resolved;
+    },
+  });
+}
+
+function formatImplicitWriteScopeFailure(
+  resolution: ImplicitWriteScopeResolution,
+  configuredDefaultScope: string | undefined,
+): string {
+  if (resolution.reason === "template_unresolved") {
+    return `template default scope '${configuredDefaultScope ?? "(missing)"}' could not be resolved`;
   }
-  return scopeManager.getDefaultScope(agentId);
+  if (resolution.reason === "scope_inaccessible") {
+    return `implicit write scope '${resolution.candidate ?? "(missing)"}' is not accessible to the caller`;
+  }
+  return `implicit write scope '${resolution.candidate ?? "(missing)"}' is not a valid concrete write scope`;
 }
 
 function resolveSourceFromSessionKey(sessionKey: string | undefined): string {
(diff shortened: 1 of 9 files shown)


rwmjhb:
Cross-user memory bleed in shared bots is a real problem and per-user scope
isolation is the right approach. The design (template scopes + wildcard ACLs) makes
sense. A few things need to be fixed before this is safe to merge.

**Must fix**

1. **Test regression — `scopeManager.validateScope is not a function`.**
   `recall-text-cleanup.test.mjs` (at least lines 740 and 760) crashes because
   `resolveImplicitWriteScope` now calls `scopeManager.validateScope(...)`, but
   existing test fixtures provide a partial `scopeManager` mock without this method.
   The production path (where the plugin supplies a full `MemoryScopeManager`) is
   fine, but all partial mocks need `validateScope` added, or the call should guard
   with a capability check.

2. **System bypass agent throws when `scopes.default` is absent.**
   `resolveImplicitWriteScope` falls back to
   `scopeManager.getDefaultScope(agentId)` for system-bypass IDs, but
   `getDefaultScope` throws when called with a system bypass ID. Before this PR,
   all three hook sites explicitly used `config.scopes?.default ?? 'global'` for
   bypass agents. The helper loses that safe fallback — it should replicate the same
   guard.

3. **`vectorSearch()` doesn't honor wildcard scopes.**
   `scopeFilterToSqlCondition` and `scopeFilterIncludes` are wired into list, BM25,
   and delete paths, but `vectorSearch()` still builds SQL with `scope = '...'` and
   filters with `scopeFilter.includes(rowScope)`. Semantic retrieval, dedup checks,
   and admission control all call `vectorSearch`, so wildcard ACLs like `user:*`
   silently don't work for the most important retrieval path.

**Worth discussing**

- **`scopeFilter` narrowing in smart extractor** (`index.ts:2811`): the hook now
  passes `scopeFilter: [defaultScope]` instead of `scopeFilter: accessibleScopes`.
  This limits dedup to the write scope only. Was this intentional (to prevent
  cross-scope contamination) or a side effect? For agents with `global + user:*`
  access, this could hide duplicate memories that already exist in other scopes.

- **Migration footgun**: with the new `agentAccess: { 'bot-1': ['global', 'user:*'] }`
  config, `isAccessible('agent:bot-1', 'bot-1')` returns `false`. Existing agents
  that relied on implicit self-scope access would stop seeing their own memories
  after upgrading. Worth documenting explicitly or adding `agent:<id>` automatically
  when explicit ACLs are set.

- **Wildcard validation is permissive**: `validateScope` and `setAgentAccess` accept
  any `*`-suffixed string without checking whether the prefix is a known namespace.
  A typo like `users:*` would silently succeed. Low priority, but worth a note in
  docs.

Fix the three blockers and resolve the scopeFilter question, then this is ready for
another look.

AliceLJY (approved):
Thorough review of all changed files. This is well-designed.

**Architecture** — Plan B (scope system stays static, template resolution at hook layer) is the right call. Clean separation between config-time scope validation and runtime context binding.

**Security** — checked:
- SQL wildcard injection: `escapeSqlLiteral` + LIKE meta-character escaping for `%` and `_` ✅
- Overly broad wildcard patterns: `inferWildcardFromTemplate` correctly rejects compound templates like `agent:${agentId}:user:${accountId}` that would produce `agent:*` ✅
- Failed template resolution: returns `undefined` + skips write (not empty-string write to bogus scope) ✅

**Backward compatibility** — static `scopes.default` works exactly as before. All existing scope-filter codepaths upgraded to wildcard-aware variants (`scopeFilterToSqlCondition`, `scopeFilterIncludes`) with consistent semantics.

**Key design decisions I agree with:**
- Dedup scope narrowed to `[defaultScope]` instead of `accessibleScopes` — prevents cross-user dedup matching in multi-tenant setups
- No auto-wildcard in `getAccessibleScopes` — wildcards only via explicit `agentAccess` config
- `getDefaultScope()` returns `"global"` for template defaults when no agentId — safe fallback

**Tests** — 30 unit tests + integration tests with mock plugin harness. Coverage looks adequate.

**One note:** manual Feishu bot test is still unchecked in the test plan. Would be good to confirm multi-user isolation works end-to-end before merge.

LGTM — approve.

rwmjhb (changes requested):
Per-user memory isolation via template scopes is a high-value feature (81%). A couple of correctness blockers to fix before merge.

**Must fix before merge:**

- **MR1: Wildcard scope support is incomplete — `vectorSearch` still exact-matches scopes.** The scope resolution and write paths handle wildcards, but vector search still filters by exact scope string. Queries against a wildcard-configured agent will return no results. This is the core of the feature — it needs to work end-to-end.

- **F4: A system bypass agent without `scopes.default` configured throws instead of gracefully returning `'global'`.** An unhandled exception for a missing config key is a regression for any deployment not yet using template scopes.

- **EF1: Build is currently failing** — CI is red and merge is blocked. Please fix and confirm the build passes.

- **EF2: Two tests crash with `scopeManager.validateScope is not a function`.** `resolveImplicitWriteScope` silently requires `validateScope` as a method on the scope manager interface, but the existing test mocks don't implement it. Either the interface addition needs to be properly declared, or the call needs a guard.

- MR2: Any trailing `:\*` wildcard is accepted, including undefined or unintended namespaces — consider validating the namespace prefix.
- F6: Smart extractor `scopeFilter` narrowed from `accessibleScopes` to `[defaultScope]` — cross-scope dedup is disabled for single-user deployments as a side effect.
- F7: `inferWildcardFromTemplate` is exported but has no callers in this PR — dead export.

Fix the four blockers and this is ready.

rwmjhb:

**Please**:

3. Push the rebased branch

PR #631, opened 2026-04-15
fix(embedder): route batch embedding to /v1/embeddings (Issue #629)

### LINKED ISSUE #629: [BUG] Ollama batch embedding fails after PR #621 (array input rejected)
## Bug Description (Detailed Analysis)

After PR #621 fixed Ollama 0.20.5+ single embedding, batch embedding now fails because `/api/embeddings` only accepts a single string `prompt` field, not an array.

## Error

```
Error: Failed to generate batch embeddings from Ollama: Ollama embedding failed: 400 Bad Request ??{"error":"json: cannot unmarshal array into Go struct field EmbeddingRequest.prompt of type string"}
```

### PR #621 Background

PR #621 (fixing Issue #620) discovered that Ollama 0.20.5+ returns empty arrays when using `/v1/embeddings` endpoint. The fix switched to using Ollama's native `/api/embeddings` endpoint with the `prompt` field instead:

```typescript
// PR #621 fix in embedWithNativeFetch()
const ollamaPayload = {
  model: payload.model,
  prompt: payload.input,  // <-- This works for single string
};
```

### The Regression

The problem is that `payload.input` is **not always a string**. When `embedMany()` builds a batch payload, it passes an array:

```typescript
// In embedMany()
const response = await this.embedWithRetry(
  this.buildPayload(validTexts, task),  // <-- validTexts is string[]
  signal,
);

// In buildPayload()
const payload = {
  model: this.model,
  input,  // <-- When input is string[], this becomes input: ["a", "b", "c"]
};
```

So when `embedWithNativeFetch()` receives the payload and sets `prompt: payload.input`, it sends an array where Ollama expects a string:

```
// What Ollama receives:
{
  "model": "jina-v5-retrieval-test",
  "prompt": ["a", "b", "c"]  // <-- Array, but Ollama expects string!
}

// Ollama error:
"json: cannot unmarshal array into Go struct field EmbeddingRequest.prompt of type string"
```

### Why Single Works But Batch Fails

| Scenario | payload.input type | Works? |
|----------|-------------------|--------|
| `embedPassage("hello")` | `string` | ✅ |
| `embedQuery("hello")` | `string` | ✅ |
| `embedBatchPassage(["a","b"])` | `string[]` | ❌ |

## Suggested Fix

**This is a routing fix based on verified behavior, not a universal solution.**

- **Single input**: keep `/api/embeddings` + `prompt` (PR #621 unchanged)
- **Batch input**: route to `/v1/embeddings` + `input[]`
- **Response validation**: verify count and non-empty embeddings

**Limitation**: This assumes `/v1/embeddings` batch mode is valid for the target Ollama/model combination, which was verified in local testing with `jina-v5-retrieval-test`.

## Testing

Local testing confirms both single and batch work correctly:

| Test | Endpoint | Input | Result |
|------|----------|-------|--------|
| Single | `/api/embeddings` | `prompt: string` | ✅ SUCCESS (1024 dims) |
| Batch 3 items | `/v1/embeddings` | `input: string[]` | ✅ SUCCESS (3 embeddings) |
| Batch 1 item | `/v1/embeddings` | `input: string[]` | ✅ SUCCESS |

Model: `jina-v5-retrieval-test`

## Related

- #620 (fixed by #621, but introduced this regression)
- PR #621

jlin53882:
## Summary

Fix Issue #629: Ollama batch embedding fails after PR #621

### Problem

PR #621 fixed Ollama 0.20.5+ single embedding by switching to `/api/embeddings` with `prompt` field. However, batch requests send `payload.input` as `string[]`, causing Ollama to reject with:

```
"json: cannot unmarshal array into Go struct field EmbeddingRequest.prompt of type string"
```

### Solution

Route requests by input type inside `embedWithNativeFetch()`:

| Input Type | Endpoint | Field |
|-----------|----------|-------|
| Single (`string`) | `/api/embeddings` | `prompt` (PR #621 unchanged) |
| Batch (`string[]`) | `/v1/embeddings` | `input` |

Additionally, add response validation to ensure batch responses contain the expected count and non-empty embeddings.

### Changes

- `src/embedder.ts`: Route based on `Array.isArray(payload.input)`
- `test/embedder-ollama-batch-routing.test.mjs`: 5 test cases covering routing and validation

### Testing

| Test | Result |
|------|--------|
| Single `/api/embeddings` with `prompt` | PASS |
| Batch `/v1/embeddings` with `input[]` | PASS |
| Batch count mismatch rejection | PASS |
| Batch empty embedding rejection | PASS |
| Single-element batch routing | PASS |

### Limitation

This assumes `/v1/embeddings` batch mode is valid for the target Ollama/model combination. Verified with `jina-v5-retrieval-test`.

Closes #629

### DIFF
--- a/commit_msg.txt
+++ b/commit_msg.txt
@@ -1,0 +1,7 @@
+fix(embedder): address PR review comments (Issue #629)
+
+- Add embedder-ollama-batch-routing.test.mjs to CI manifest
+- Add comments explaining why provider options are omitted for Ollama batch
+- Add note about /v1/embeddings no-fallback assumption
+
+Reviewed by: rwmjhb
--- a/scripts/ci-test-manifest.mjs
+++ b/scripts/ci-test-manifest.mjs
@@ -47,6 +47,8 @@ export const CI_TEST_MANIFEST = [
   { group: "core-regression", runner: "node", file: "test/store-serialization.test.mjs" },
   { group: "core-regression", runner: "node", file: "test/access-tracker-retry.test.mjs" },
   { group: "core-regression", runner: "node", file: "test/embedder-cache.test.mjs" },
+  // Issue #629 batch embedding fix
+  { group: "llm-clients-and-auth", runner: "node", file: "test/embedder-ollama-batch-routing.test.mjs" },
 ];
 
 export function getEntriesForGroup(group) {
--- a/src/embedder.ts
+++ b/src/embedder.ts
@@ -569,39 +569,83 @@ export class Embedder {
    * Call embeddings.create using native fetch (bypasses OpenAI SDK).
    * Used exclusively for Ollama endpoints where AbortController must work
    * correctly to avoid long-lived stalled sockets.
+   *
+   * For Ollama 0.20.5+: /v1/embeddings may return empty arrays for some models,
+   * so we use /api/embeddings with "prompt" field for single requests (PR #621).
+   * For batch requests, we use /v1/embeddings with "input" array as it's more
+   * efficient and confirmed working in local testing.
+   *
+   * See: https://github.com/CortexReach/memory-lancedb-pro/issues/620
+   * Fix: https://github.com/CortexReach/memory-lancedb-pro/issues/629
    */
   private async embedWithNativeFetch(payload: any, signal?: AbortSignal): Promise<any> {
     if (!this._baseURL) {
       throw new Error("embedWithNativeFetch requires a baseURL");
     }
 
-    // Fix for Ollama 0.20.5+: /v1/embeddings returns empty arrays for both `input` and `prompt`.
-    // Only /api/embeddings + `prompt` parameter works correctly.
-    // See: https://github.com/CortexReach/memory-lancedb-pro/issues/620
     const base = this._baseURL.replace(/\/$/, "").replace(/\/v1$/, "");
-    const endpoint = base + "/api/embeddings";
-
     const apiKey = this.clients[0]?.apiKey ?? "ollama";
 
-    // Ollama's /api/embeddings requires "prompt" field, not "input"
-    const ollamaPayload = {
-      model: payload.model,
-      prompt: payload.input,
-    };
+    // Handle batch requests with /v1/embeddings + input array
+    if (Array.isArray(payload.input)) {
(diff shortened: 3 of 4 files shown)


rwmjhb:
Clean fix for the batch embedding regression introduced by #621. The root cause diagnosis is correct and the routing split is the right approach. Ready to merge with a few minor notes.

**Nice to have (non-blocking)**

- Tests bind to port 11434 — the live Ollama default. Any developer running Ollama locally will hit `EADDRINUSE`. Consider a random available port or mock the HTTP layer instead.
- The two positive routing tests assert the response shape but not *which endpoint was called* — a future routing regression won't be caught. Worth asserting the request URL in at least one test.
- The new batch native-fetch branch discards provider options already computed by `buildPayload()` — double-check whether timeout, headers, or other options from that path are needed on the batch route.
- `/v1/embeddings` is used unconditionally for batch with no fallback — if a model doesn't support that endpoint, the failure will be silent from the user's perspective. A comment noting the assumption would help future maintainers.

**CI note**: New test file `test/embedder-ollama-batch-routing.test.mjs` appears unregistered in the CI test manifest — `verify-ci-test-manifest.mjs` may need updating before the full suite passes.

**Before merge**: `stale_base=true` — quick rebase against main to confirm nothing has touched `embedder.ts` there.

jlin53882:
## PR Review Updates

All review comments from @rwmjhb have been addressed:

### Fixed:
1. ✅ **CI manifest** - Added `test/embedder-ollama-batch-routing.test.mjs` to `scripts/ci-test-manifest.mjs`
2. ✅ **Port conflict** - Changed test to use port 0 (OS-assigned available port) instead of hardcoded 11434, avoiding EADDRINUSE when developers have Ollama running locally
3. ✅ **Endpoint verification** - Test already validates which endpoint is called via `capturedRoute` assertion
4. ✅ **Provider options** - Added code comments explaining why `encoding_format`, `normalized`, etc. are omitted for Ollama batch (Ollama doesn't support these params)
5. ✅ **No fallback** - Added comment documenting the assumption that `/v1/embeddings` is universally supported

### Rebase Check
Verified: `embedder.ts` has **not changed** in upstream master since PR base (`16ee3e4`). No rebase needed.

Ready for merge! 🎉

PR #634, opened 2026-04-15
feat: Priority 3 Memory Enhancements — Shared Scope, Entity Graph, Confidence Scoring, Proactive Injection

helal-muneer:
## Summary

Four advanced memory features to make OpenClaw agents significantly smarter:

### H. Cross-Agent Shared Memory (`shared:` scope)
- New `shared` built-in scope automatically readable by all agents
- Write access restricted to dreaming engine or explicit `memory_shared` tool
- Optional auto-promote: memories accessed by 3+ agents get promoted to shared during dream cycle
- Config: `scopes.shared.enabled` (default: true), `scopes.shared.autoPromote` (default: false)

### I. Entity Relationship Layer (`src/entity-graph.ts`)
- Regex-based entity extraction (people, projects, tools, locations, preferences, orgs)
- Relationship extraction (subject-predicate-object triples)
- In-memory graph with `getRelated()`, `getEntityProfile()`, `getAllEntities()`
- Noop variant when disabled — zero overhead
- Config: `entityGraph.enabled` (default: false)

### J. Memory Confidence Scoring (`src/confidence-tracker.ts`)
- Tracks recall count and useful-recall count per memory
- Formula: `confidence = usefulCount / max(recallCount, 1) * decayBoost`
- Decay boost decreases on unhelpful recalls, recovers on useful signals
- Integrates with auto-recall pipeline (automatic tracking)
- New tool: `memory_boost` — manually boost a memory's confidence

### K. Proactive Memory Injection (`src/proactive-injector.ts`)
- Three trigger types:
  1. **Entity-based**: pre-fetch related memories when user mentions a known entity
  2. **Pattern triggers**: inject memories matching configurable regex patterns
  3. **Stale memory**: random 5% check for memories not recalled in N days
- Max 1 injection per turn, dedup against auto-recall results
- Config: `proactive.enabled` (default: false), `staleMemoryDays`, `entityPrefetch`, `patternTriggers`

## New Tools
- `memory_entities` — query entity graph (profile/related)
- `memory_boost` — manually boost memory confidence
- `memory_shared` — explicitly write to shared scope

## Files Changed
- **Created**: `src/entity-graph.ts`, `src/confidence-tracker.ts`, `src/proactive-injector.ts`
- **Modified**: `src/scopes.ts`, `src/tools.ts`, `index.ts`, `openclaw.plugin.json`
- **Tests updated**: `scope-access-undefined.test.mjs`, `clawteam-scope.test.mjs`

## Design Principles
- All features **off by default** — no breaking changes
- No new external dependencies
- No modification to existing functionality — purely additive
- Conservative defaults with granular opt-in config

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -79,6 +79,9 @@ import {
   type AdmissionRejectionAuditEntry,
 } from "./src/admission-control.js";
 import { analyzeIntent, applyCategoryBoost } from "./src/intent-analyzer.js";
+import { createEntityGraph, type EntityGraph } from "./src/entity-graph.js";
+import { createConfidenceTracker, type ConfidenceTracker } from "./src/confidence-tracker.js";
+import { createProactiveInjector, type ProactiveInjector } from "./src/proactive-injector.js";
 
 // ============================================================================
 // Configuration & Types
@@ -183,6 +186,7 @@ interface PluginConfig {
     default?: string;
     definitions?: Record<string, { description: string }>;
     agentAccess?: Record<string, string[]>;
+    shared?: { enabled?: boolean; autoPromote?: boolean };
   };
   enableManagementTools?: boolean;
   sessionStrategy?: SessionStrategy;
@@ -224,6 +228,13 @@ interface PluginConfig {
   extractionThrottle?: {
     skipLowValue?: boolean;
     maxExtractionsPerHour?: number;
+  };
+  entityGraph?: { enabled?: boolean };
+  proactive?: {
+    enabled?: boolean;
+    staleMemoryDays?: number;
+    entityPrefetch?: boolean;
+    patternTriggers?: Record<string, string>;
   };
 }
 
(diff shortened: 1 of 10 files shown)


rwmjhb:
The individual capabilities here are interesting, but several are shipped in a non-functional or unsafe state that would affect users who enable them. Two must-fix items before this is ready.

**Must fix**

- **Proactive injection is wired but never called**: The `ProactiveMemoryInjector` is instantiated but its `inject()` method is never invoked in the request pipeline. Users who enable `proactiveInjection: true` will see no behavior change. The injection call needs to be hooked into the retrieval or response path to deliver the advertised feature.
- **Entity graph leaks across scopes**: `EntityRelationshipGraph` queries all memories regardless of scope, then returns related entities to any requesting agent. An agent in scope A can receive entity facts extracted from scope B's private memories. Scope filtering must be applied at graph construction or query time before this is safe to enable in multi-agent deployments.

**High priority**

- `scopes.shared.enabled: false` has no effect — the shared scope is active unconditionally. The config guard is missing.
- Shared scope write restriction is not enforced: the description says only the dreaming engine or `memory_shared` tool may write to `shared`, but `memory_store` can write there directly. `memory_shared` also bypasses safeguards that `memory_store` already enforces.
- Confidence scoring is tracked in-memory only — state is lost on restart. The comment implies metadata persistence, but none exists. Users will see confidence reset to default after every restart.
- `getRelated(entity, 1)` traverses 2 hops — depth semantics are off by one.
- `autoPromote` config field is defined in schema and types but has zero implementation. The PR description describes auto-promotion during the dream cycle, but no promotion logic appears in the diff — where does it live?

**Other**

- `timestamp=0` entries are always treated as stale — is this intentional or a sentinel value that should be handled separately?
- Preference entity extraction stores the verb rather than the preference value (e.g. "loves" instead of "coffee").
- Three new source modules (519 lines) have zero test coverage.

**Questions for author**

- Is this part of a maintainer-approved roadmap? The Priority 3 / H-K numbering implies earlier PRs landed first — were those merged with explicit sign-off to continue the series?
- CI fails with `cli-smoke.mjs: undefined !== 1` at line 316 — pre-existing on base, or introduced here? Needs root-cause before merge.

rwmjhb:

**Please**:

3. Push the rebased branch

rwmjhb:

**Please**:

3. Push the rebased branch

PR #650, opened 2026-04-17
fix: register active runtime and allow dreaming config

holeshotclaw:
## Summary
- register an active memory capability runtime so OpenClaw can discover the plugin as the selected backend
- expose search/read/status probes through that runtime bridge
- allow a top-level `dreaming` config block so OpenClaw sidecar dreaming can be configured when this plugin owns the memory slot
- add regression coverage for the manifest key and runtime registration

## Testing

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1984,6 +1984,167 @@ const memoryLanceDBProPlugin = {
     );
 
     // ========================================================================
+    // Active Memory Runtime Registration
+    // ========================================================================
+
+    const activeMemoryPathPrefix = "memory-lancedb-pro/entries/";
+    const activeMemoryWorkspaceDir = getDefaultWorkspaceDir();
+
+    const formatActiveMemoryPath = (id: string) => `${activeMemoryPathPrefix}${id}.md`;
+
+    const parseActiveMemoryPath = (relPath: string): string | null => {
+      const normalized = relPath.trim().replace(/^\/+/, "");
+      if (!normalized.startsWith(activeMemoryPathPrefix) || !normalized.endsWith(".md")) {
+        return null;
+      }
+      const id = normalized.slice(activeMemoryPathPrefix.length, -3).trim();
+      return id || null;
+    };
+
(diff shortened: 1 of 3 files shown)


rwmjhb:

**Please**:

3. Push the rebased branch

rwmjhb (changes requested):
Requesting changes because the new runtime registration can crash hosts that do not expose this API.

1. `index.ts:2034` — `api.registerMemoryCapability(...)` is called unconditionally.
The existing `test/recall-text-cleanup.test.mjs` harness has no `registerMemoryCapability` method (`test/recall-text-cleanup.test.mjs:37-58`), and `npm test` fails at `test/recall-text-cleanup.test.mjs:221` with `TypeError: api.registerMemoryCapability is not a function`. More importantly, this suggests older or alternate OpenClaw hosts can fail plugin registration before lifecycle hooks/services are registered. Please feature-detect the method, add a compatibility fallback, or make the call safely optional.

A few non-blocking follow-ups:

- `index.ts:2039-2040`: `getMemorySearchManager()` eagerly performs `store.stats()` and `embedder.test()` before any manager method is used. That adds DB/network latency to manager construction and can return `manager: null` even when search/read paths would otherwise work. Please compute status/probe data lazily inside `status()` / probe methods.

- `index.ts:2083-2087`: search results report `startLine: 1`, but `readFile()` returns a synthetic markdown document where memory text starts after the metadata header, around line 8. Citations can point at `# Memory ...` instead of the matched text. Please offset the returned line range or make search/read line numbering consistent.

- `index.ts:2120-2122`: `probeVectorAvailability()` only calls `embedder.test()`, so it reports embedding health rather than vector-store/retrieval health. Please probe the store/retriever path as well, or rename the probe semantics.

The runtime registration is the right direction, but the host API compatibility crash needs to be fixed before merge.

rwmjhb:

**Please**:

3. Push the rebased branch

rwmjhb (commented):

This has been open for more than two weeks, and there has been no author response or new push since the last review. Please feel free to open a fresh PR, or ask to reopen this one, once the requested changes are ready.

PR #655, opened 2026-04-18
feat(cli): add repair-summaries command for stale L0/L1/L2 fix

raw34:
One-time repair tool that detects memories where text was updated but L0/L1/L2 summaries were not regenerated. Supports --dry-run preview and --scope filtering.


raw34:
CI failures are pre-existing on master — not introduced by this PR. See #657  for the fix. Once that's merged, this PR will pass CI after a rebase.

PR #626, opened 2026-04-14
fix: resolve stale lock and same-instance write contention (#622, #623)

jlin53882:
## Summary
Fixes two related lock issues in memory-lancedb-pro:

- **#622**: Stale \.memory-write.lock\ not released, blocking writes on Windows
- **#623**: Same-instance concurrent writes competing for lock

### src/store.ts
- Added \updateQueue\ for intra-instance write serialization
- Added \lockfilePath\ option with explicit lock path
- Added proactive stale lock cleanup (5-minute threshold)
- Fixed nested deadlock where \update()\ used both \
unWithFileLock()\ AND \
unSerializedUpdate()\

### test/lock-recovery.test.mjs
- Tests for stale lock recovery

### test/store-write-queue.test.mjs
- Tests for queue serialization

## Testing
- 7 tests pass, 1 skip

Closes #622
Closes #623

PR #541, opened 2026-04-06
fix: remove duplicate memory_compact registration + redundant schema required (OpenClaw 2026.4.5 compat)

Banger455:
## Problem

After updating to OpenClaw 2026.4.5, `memory-lancedb-pro` fails with two errors:

1. **`embedding: must have required property 'embedding'`** — config validation error
2. **`plugin tool name conflict: memory_compact`** — duplicate tool registration

### 1. Redundant `required: ["embedding"]` in configSchema (50afadb)

The root-level `required: ["embedding"]` in `openclaw.plugin.json` was always redundant — the plugin validates `embedding.apiKey` internally via `parsePluginConfig()`, and the sub-schema already has `required: ["apiKey"]`.

OpenClaw 2026.4.5 changed how config values reach the AJV validator during framework-level validation, causing this constraint to fail even though the config on disk is correct. Removing it fixes the validation error.

### 2. Duplicate `memory_compact` tool registration (5afc826)

`registerAllMemoryTools()` already calls `registerMemoryCompactTool()` (src/tools.ts:2194). A standalone 88-line block in `index.ts` registered the same tool again. OpenClaw 2026.4.5 now throws on duplicate tool names.

## Testing

Both fixes verified live on OpenClaw 2026.4.5 instance:
- Gateway starts clean: `ready (1 plugin, 0.4s)`
- No `invalid config` errors
- No `tool name conflict` errors

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #668, opened 2026-04-19
fix(auto-recall): thread AbortSignal to cancel embedding on timeout

ScientificProgrammer:
## Summary

The auto-recall `before_prompt_build` hook wraps `recallWork()` in `Promise.race` with a `setTimeout` resolver (`index.ts:~2690`). When the timer fires, the race resolves `undefined` — but the underlying embedding HTTP call **keeps running**. No `AbortController`/`AbortSignal` is threaded through. The lingering work holds the per-agent memory-runtime mutex, so subsequent hook invocations serialize behind it.

In production at the reporter's site, this manifested as `stuck session: state=processing age=219s queueDepth=0` with the gateway's session-write-lock watchdog firing repeatedly. The gateway's lock max-hold is `timeoutMs + 120s grace` (min 300s), so a single slow embedding call ends up holding the session lock for 2–5 minutes before forced release.

## Changes

- **`index.ts`** — create an `AbortController` in the hook; pass `controller.signal` into a new `recallWork(signal)` parameter; in the `setTimeout` callback, call `controller.abort(new Error("auto-recall timeout"))` before resolving. The `catch` branch downgrades `AbortError` to a `debug` log (the timeout path already emitted a `warn`). `retrieveWithRetry` now accepts an optional `signal` and early-returns after the 75 ms retry gap if the signal was aborted in the interim.
- **`src/retriever.ts`** — add `signal?: AbortSignal` to `RetrievalContext`; destructure in `retrieve()`; forward to `vectorOnlyRetrieval()` and `hybridRetrieval()`, which pass it to `this.embedder.embedQuery(query, signal)`. The embedder already accepts an `AbortSignal` throughout (`src/embedder.ts:794,798,806,810,865`), so no downstream changes are needed.

## Scope

- **In scope:** auto-recall hook → retriever → embedder cancellation.
- **Out of scope (follow-ups):**
  - Jina rerank HTTP calls (separate code path in the retriever's rerank stage). Should also accept the signal. Left for a focused follow-up that touches the rerank client interface.
  - BM25 path has no long-running HTTP call and doesn't invoke `embedQuery`, so no plumbing needed.

## Test plan

Verified locally:
- `tsc --noEmit` introduces no new type errors. Two pre-existing errors remain (notably `TS2353` on `retrieveWithRetry` where `source` is passed but not declared in the inline param type — introduced by #535, independent of this PR).
- The previously-deployed (stale) plugin exhibited the stuck-session symptom under slow-embedding conditions; after rolling to a build with this change applied, a contrived delayed-embedding scenario returns from the hook inside the configured `autoRecallTimeoutMs` and the in-flight embedding fetch sees `signal.aborted === true`.

Automated test not added in this PR — suggested follow-up: a unit test that stubs the embedder with a 30 s delay and asserts `recallWork` returns within 100 ms of the timeout AND the stub receives an aborted signal.

## Related

- RCA: stuck-session incident on 2026-04-18 (`proving-ground` agent, OpenClaw 2026.4.9) directly implicated this code path.
- Related upstream fixes already merged: #541 (duplicate `memory_compact` registration), #626 (stale lockfile). Those address adjacent failure modes but not the `Promise.race`-without-cancellation root cause fixed here.

---

🤖 Developed with Claude Code (Claude Opus 4.7, 1M context).

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1942,10 +1942,12 @@ const memoryLanceDBProPlugin = {
       limit: number;
       scopeFilter?: string[];
       category?: string;
+      signal?: AbortSignal;
     }) {
       let results = await retriever.retrieve(params);
       if (results.length === 0) {
         await sleep(75);
+        if (params.signal?.aborted) return results;
         results = await retriever.retrieve(params);
       }
       return results;
@@ -2407,7 +2409,13 @@ const memoryLanceDBProPlugin = {
         // (embedding → rerank → lifecycle), which can silently drop messages on
         // channels like Telegram when subsequent requests hit lock timeouts.
         // See: https://github.com/CortexReach/memory-lancedb-pro/issues/253
-        const recallWork = async (): Promise<{ prependContext: string } | undefined> => {
+        //
+        // The timeout also aborts the in-flight embedding HTTP call via the
+        // abortController below, so the underlying work doesn't continue to hold
+        // resources (lancedb connection, per-agent memory-runtime mutex) past
+        // the timeout boundary.
+        const abortController = new AbortController();
+        const recallWork = async (signal: AbortSignal): Promise<{ prependContext: string } | undefined> => {
           // Determine agent ID and accessible scopes
           const agentId = resolveHookAgentId(ctx?.agentId, (event as any).sessionKey);
           const accessibleScopes = resolveScopeFilter(scopeManager, agentId);
@@ -2450,6 +2458,7 @@ const memoryLanceDBProPlugin = {
             limit: retrieveLimit,
             scopeFilter: accessibleScopes,
             source: "auto-recall",
+            signal,
           }), config.workspaceBoundary);
 
           if (results.length === 0) {
@@ -2690,9 +2699,12 @@ const memoryLanceDBProPlugin = {
         let timeoutId: ReturnType<typeof setTimeout> | undefined;
         try {
           const result = await Promise.race([
-            recallWork().then((r) => { clearTimeout(timeoutId); return r; }),
+            recallWork(abortController.signal).then((r) => { clearTimeout(timeoutId); return r; }),
             new Promise<undefined>((resolve) => {
               timeoutId = setTimeout(() => {
+                // Cancel in-flight embedding/retrieval HTTP calls so they don't
+                // keep holding resources after we've given up on the result.
+                abortController.abort(new Error("auto-recall timeout"));
                 api.logger.warn(
                   `memory-lancedb-pro: auto-recall timed out after ${AUTO_RECALL_TIMEOUT_MS}ms; skipping memory injection to avoid stalling agent startup`,
                 );
@@ -2703,7 +2715,14 @@ const memoryLanceDBProPlugin = {
           return result;
         } catch (err) {
           clearTimeout(timeoutId);
-          api.logger.warn(`memory-lancedb-pro: recall failed: ${String(err)}`);
+          // Downgrade AbortError to debug — the timeout path already emitted
+          // a warn log, and the race has already resolved undefined.
+          const isAbort = err instanceof Error && (err.name === "AbortError" || /abort/i.test(err.message));
+          if (isAbort) {
+            api.logger.debug?.(`memory-lancedb-pro: recall aborted: ${String(err)}`);
+          } else {
+            api.logger.warn(`memory-lancedb-pro: recall failed: ${String(err)}`);
+          }
         }
       }, { priority: 10 });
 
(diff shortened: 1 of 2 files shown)


ScientificProgrammer:
@copilot-pull-request-reviewer — feedback on `6119d11` addressed in commit `373e7f2`:

**1. Abort classification narrowed (`index.ts`)** — replaced the overbroad `err.name === "AbortError" || /abort/i.test(err.message)` with `abortController.signal.aborted`. The flag is owned by this hook's own controller, so only *our* timeout callback sets it. AbortErrors originating elsewhere (e.g. the embedder's internal `withTimeout` controller in `src/embedder.ts:746`) no longer get silenced to debug.

**2. Retriever docstring tightened (`src/retriever.ts`)** — updated to "rejects due to abort (often with AbortError or the signal's abort reason)" to match the actual rejection path: when an already-aborted signal is passed, `withTimeout` rejects with `externalSignal.reason ?? new Error("aborted")` (a plain `Error` in our timeout case), not necessarily an `AbortError`.

**3. `retrieveWithRetry` params typing (`index.ts:1952`)** — left as-is in this commit. That's a pre-existing TS2353 introduced by PR #535, unrelated to this change; happy to take it in a separate focused PR if that's preferred.

**Follow-up noted for a later PR** (out of scope here, not a regression): the `recallWork(abortController.signal).then(...)` leg has no `.catch()`. If `recallWork` rejects *after* the race has already resolved with `undefined` (timeout path), the rejection is silently dropped. In practice this is rare — once the signal aborts, downstream code exits quickly — but an explicit `.catch()` on that leg would eliminate the unhandled-rejection window entirely.

app3apps:
Thanks for tackling this. The problem is real and worth fixing, and the overall direction here looks right.

I’m still at request changes for one blocker before merge:

- When the timeout wins the `Promise.race`, the in-flight `recallWork(abortController.signal)` can still reject after the caller has already moved on. In the current shape, that rejection is not handled explicitly, so we risk trading the stuck-mutex problem for an unhandled promise rejection on the timeout path.

What I’d like to see before merge:

- Handle the post-timeout rejection path explicitly, so an abort-triggered failure is consumed intentionally rather than escaping as an unhandled rejection.
- Add a focused test for the `timeout -> abort -> cancellation` path. Right now the risky path introduced by this change is not covered directly.

Non-blocking notes:

- I’m not treating the current full-suite failures in `recall-text-cleanup.test.mjs` as a blocker for this PR by themselves, since they look environment-coupled from the review artifacts.
- If there is an easy type/build check available for the `AbortSignal` plumbing, that would help confidence.

Once the timeout rejection path is handled, this looks close.

app3apps (changes requested):
Requesting changes before merge.

The direction here looks right and the problem is worth fixing, but there is still one blocker on the timeout path:

- When the timeout wins the `Promise.race`, the in-flight `recallWork(abortController.signal)` can still reject after the caller has already moved on. In the current shape, that rejection is not handled explicitly, so we risk turning the stuck-mutex issue into an unhandled promise rejection on the abort path.

What I’d like before merge:

- Handle the post-timeout rejection path explicitly so an abort-triggered failure is consumed intentionally rather than escaping as an unhandled rejection.
- Add a focused test for the `timeout -> abort -> cancellation` path.

Once that is covered, this looks close.

jlin53882:
Thanks for the foundation! The signal-threading concept from this PR has been rebuilt more cleanly in **#783**.

**What changed vs this PR:**
- AbortSignal properly threaded from `RetrievalContext` → `embedQuery(query, signal)` (true HTTP cancellation)
- BM25 fallback on embed failure (James's design — graceful degradation)
- No longer forces auto-recall to BM25 mode (fixed the quality regression from PR #708)
- Fixed the `.catch()` handler to avoid unhandled promise rejections after timeout

**Original PR #708's "always BM25 for auto-recall" has been removed** — it's replaced by "BM25 as fallback when embed fails."

jlin53882:
Thanks for the foundation! The signal-threading concept from this PR has been rebuilt more cleanly in **#784**.

**What changed vs this PR:**
- AbortSignal properly threaded from `RetrievalContext` → `embedQuery(query, signal)` (true HTTP cancellation)
- BM25 fallback on embed failure (James's design — graceful degradation)
- No longer forces auto-recall to BM25 mode (fixed the quality regression from PR #708)
- Fixed the `.catch()` handler to avoid unhandled promise rejections after timeout

**Original PR #708's "always BM25 for auto-recall" has been removed** — it's replaced by "BM25 as fallback when embed fails."

PR #701, opened 2026-04-26
fix: sync memory_update category metadata

stanleeyY:
## Summary
- fetch the existing memory when `memory_update` changes `category` without changing text
- keep `metadata.memory_category` and temporal `fact_key` in sync for category-only updates
- also sync `memory_category` when text and category are updated together
- extend metadata refresh regression coverage for category-only and text+category updates

## Tests
- `node test/memory-update-metadata-refresh.test.mjs`
- `node test/memory-update-supersede.test.mjs`
- `node --test test/temporal-awareness.test.mjs`
- `npm run test:cli-smoke`

### DIFF
--- a/src/tools.ts
+++ b/src/tools.ts
@@ -1284,7 +1284,7 @@ export function registerMemoryUpdateTool(
           // importance-only change that still needs metadata sync). Shared by
           // the temporal supersede guard and the normal-path metadata rebuild.
           let existing: MemoryEntry | null = null;
-          if (text || importance !== undefined) {
+          if (text || importance !== undefined || category) {
             existing = await context.store.getById(resolvedId, scopeFilter);
           }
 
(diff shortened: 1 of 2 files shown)


rwmjhb (changes requested):
Thanks for fixing the stale `memory_update` metadata path. The bug is worth addressing, but this implementation still has category-consistency issues that need to be fixed before merge.

The main blocker is that the update path writes the user-facing category enum directly into `metadata.memory_category`. The tool schema accepts public values such as `preference`, `fact`, `decision`, `entity`, `reflection`, and `other`, while the internal smart metadata layer expects categories such as `preferences`, `entities`, `patterns`, etc. The code should bridge through the existing category mapping logic instead of casting with `as any` and storing the public vocabulary in internal metadata.

There is also a remaining gap in the supersede path: when text and category are updated together, the new record can still be created with metadata derived from the old category. That leaves exactly the stale category/fact-key state this PR is trying to eliminate.

Please update both the normal update path and the temporal supersede path to rebuild smart metadata from the correct internal category. The tests should use the production-facing category values from the tool schema, not internal values like `preferences`/`patterns`, otherwise they can pass while the real API path is still wrong.

After those are fixed, this should be a useful improvement.

rwmjhb:

This has been open for more than a week, and there has been no author response or new push since the last review. Please feel free to open a fresh PR, or ask to reopen this one, once the requested changes are ready.

PR #708, opened 2026-04-28
fix: restore CLI compat and unblock auto-recall latency

a410979729-sys:
## Summary
- restore OpenClaw 2026.4.24 CLI compatibility by adding explicit `cli-metadata` support and compat runtime registrations
- keep auto-recall on a lightweight BM25-only path, thread cancellation signals through retrieval, and abort timed-out work
- stop prompt-build from doing synchronous metadata writes, and make startup retrieval checks local/lightweight instead of provider-latency sensitive
- update regression harnesses/tests for the newer plugin API hooks and before_reset behavior

## Testing
- npm test

### DIFF
--- a/cli-metadata.ts
+++ b/cli-metadata.ts
@@ -1,0 +1,20 @@
+import { definePluginEntry } from "openclaw/plugin-sdk/plugin-entry";
+
+export default definePluginEntry({
+  id: "memory-lancedb-pro",
+  name: "Memory (LanceDB Pro)",
+  description: "Enhanced LanceDB-backed long-term memory with hybrid retrieval, multi-scope isolation, long-context chunking, and management CLI",
+  kind: "memory",
+  register(api) {
+    api.registerCli(() => {}, {
+      commands: ["memory-pro"],
+      descriptors: [
+        {
+          name: "memory-pro",
+          description: "Enhanced memory management commands (LanceDB Pro)",
+          hasSubcommands: true,
+        },
+      ],
+    });
+  },
+});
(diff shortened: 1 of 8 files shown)


rwmjhb (commented):
Review action: COMMENT

Thanks for the PR. GitHub currently reports this branch as not mergeable (`mergeable=CONFLICTING`, `merge_state_status=DIRTY`), so I am deferring deep review until the diff can be reviewed against the current base.

Please rebase onto the latest base branch, resolve the merge conflicts, and push the updated branch. Once it is cleanly mergeable again, I will re-run the full review on the updated diff.

jlin53882:
Thanks for the lightweight auto-recall idea! The BM25-first concept is now reimplemented as a **fallback on embed failure**, not the default path.

**What changed vs this PR:**
- Auto-recall no longer forces BM25 mode (your concern about changing default behavior is addressed)
- BM25 is now the **fallback** when `embedQuery` fails (timeout, network error, API error)
- Signal threading added: `AbortController` can now actually cancel in-flight HTTP embed calls

See **#783** for the combined rebuild of both PRs #668 and #708.

jlin53882:
Thanks for the lightweight auto-recall idea! The BM25-first concept is now reimplemented as a **fallback on embed failure**, not the default path.

**What changed vs this PR:**
- Auto-recall no longer forces BM25 mode (your concern about changing default behavior is addressed)
- BM25 is now the **fallback** when `embedQuery` fails (timeout, network error, API error)
- Signal threading added: `AbortController` can now actually cancel in-flight HTTP embed calls

See **#784** for the combined rebuild of both PRs #668 and #708.

PR #719, opened 2026-04-29
feat(import-markdown): --no-dedup by default, new return fields, unified summary (closes #715)

jlin53882:
## Summary

Rebuilt cleanly from upstream/master based on PR #717 intent (Issue #715), containing only the relevant changes:

- **Dedup enabled by default**: flag changed from `--dedup` to `--no-dedup` (Commander.js invert pattern)
- **Return type expanded**: added `skippedShort`, `skippedDedup`, `errorCount` fields
- **TypeError fix**: early return with no files now returns all 6 fields (was only 3)
- **Summary format unified**: single `Memory Import Status:` format replaces dual dry-run/real modes
- **Consistent log tags**: `[scan]`, `[would-import]`, `[skip] dedup`, `[skip] error`
- **Dedup log always shown**: removed `if (!dryRun)` guard
- **errorCount tracked separately**: error cases counted independently with `[skip] error:` prefix

## Changes (cli.ts only, +31 / -18 lines)

| Before | After |
|--------|-------|
| `--dedup` | `--no-dedup` |
| Return: `{ imported, skipped, foundFiles }` | Return: `{ imported, skipped, foundFiles, skippedShort, skippedDedup, errorCount }` |
| `[dry-run] would import:` | `[would-import] [scope]` |
| `[skip] already imported:` (dry-run hidden) | `[skip] dedup [scope]:` (always shown) |
| Dual summary (DRY RUN / Import complete) | Single `Memory Import Status:` |
| `skipped` includes short+read-err only | `skipped` = short+dedup+read-err+error; granular counts in new fields |

## Test Results

All 14 existing tests pass (import-markdown.test.mjs — no test changes needed, new fields are backward-compatible with existing assertions).

## Related Issues

- Closes #715

PR #720, opened 2026-04-29
fix(import-markdown): dedup uses retrieve() hybrid search (fixes rank-1 gap)

jlin53882:
## Problem

The dedup check in `import-markdown` used `ctx.store.bm25Search(text, 5)` and only compared against the top result (`existing[0]`). When an identical text existed in the same scope but was ranked 2–5 by BM25, it was not detected — causing duplicate imports.

## Fix

Replace `bm25Search` with `ctx.retriever.retrieve()` with `limit: 20`, then exact-match `results[0].entry.text`. This runs the full hybrid pipeline (BM25 + vector + reranker) against 20 candidates before exact-matching the top result — the same pipeline used for normal retrieval.

**Before (Bug):**
```typescript
const existing = await ctx.store.bm25Search(text, 5, [effectiveScope]);
if (existing.length > 0 && existing[0].entry.text === text) { // only checked rank 1
```

**After (Fix):**
```typescript
const results = await ctx.retriever.retrieve({
  query: text,
  limit: 20,
  scopeFilter: [effectiveScope],
  source: "cli",
});
if (results.length > 0 && results[0].entry.text === text) { // checks top of hybrid pipeline
```

## Scope

- Single file changed: `cli.ts`
- +10 / -3 lines
- No changes to return type or CLI interface
- No new tests required (exact-match logic unchanged, just more thorough search)

## Refs

- Issue: #715
- Original issue report with root-cause analysis: cortexreach/memory-lancedb-pro#715#issuecomment-4340601924

### DIFF
--- a/cli.ts
+++ b/cli.ts
@@ -693,11 +693,18 @@ export async function runImportMarkdown(
       const effectiveScope = options.scope || discoveredScope;
 
       // ── Deduplication check (scope-aware exact match) ───────────────────
-      // Run even in dry-run so --dry-run --dedup reports accurate counts
+      // Use retrieve() for hybrid search + rerank pipeline, then exact match.
+      // Run even in dry-run so --dry-run --dedup reports accurate counts.
+      // Replaces: ctx.store.bm25Search(text, 5) which only checked rank-1.
       if (dedupEnabled) {
         try {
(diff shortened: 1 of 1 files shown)


jlin53882:
Closed — superseded by #733 (combines #719 + #720)

PR #522, opened 2026-04-04
fix: isOwnedByAgent derived ownership (#448)

jlin53882:
## Summary

Fixes \isOwnedByAgent\ in \src/reflection-store.ts\ so that \derived\ items are **not** incorrectly inherited by the main agent via the \owner === 'main'\ fallback, preventing context bleed between agents.

Also fixes a P1 bug where the \_initialized\ flag was set **before** \
egister()\ completed — if initialization threw, the plugin would become permanently broken until process restart.

## Changes

| File | Change |
|------|--------|
| \src/reflection-store.ts\ | isOwnedByAgent: derived items gated to owning agent only; empty-owner derived returns false |
| \index.ts\ | \_initialized\ flag moved to end of successful \
egister()\ |

## Testing

- Unit tests for isOwnedByAgent: passed
- No new test failures introduced

---

**Related**: Supersedes PR #509, which contained scope creep issues (unrelated features bundled in the same PR). This clean version only contains the #448 fix and the _initialized P1 bug fix.

PR #721, opened 2026-04-29
feat: dual-track registration with early claim + rollback support

jlin53882:
## Summary
Dual-track registration system for memory-lancedb-pro:

- **Before init** (Phase 2): claim api in both  (WeakSet) and  (Map)
- **On init failure**: explicit  rollback, enabling retry with same API object
- **resetRegistration()**: now clears both WeakSet and Map

This enables test inspection of registered APIs and provides explicit rollback semantics on init failure.

## Changes
- : +30/-6 lines (single file)

## Dependencies
Based on master after PR #522 merge.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1675,6 +1675,21 @@ const pluginVersion = getPluginVersion();
 // hook/tool registration for the new API instance" regression that rwmjhb identified.
 let _registeredApis = new WeakSet<OpenClawPluginApi>();
 
+// Dual-track registration: alongside WeakSet (GC-safe), use a Map for explicit
+// rollback tracking and test inspection. WeakSet handles GC safety; Map provides
+// manual clearability and _getRegisteredApisForTest() export.
+// Track: _registeredApisMap (explicit claim/rollback) + _registeredApis (WeakSet guard)
+let _registeredApisMap = new Map<OpenClawPluginApi, boolean>();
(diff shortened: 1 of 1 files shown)


jlin53882:
## Why This PR Matters

### The Problem

After PR #522,  places  **only after** successful . This means:

- **On init failure**: the WeakSet has zero record that a registration was ever attempted — no trace, no debug trail
- **Retry logic has no foundation**: a subsequent  call with the same API has no way to know "this was already tried and failed"
- **Tests cannot assert**: there is no way to verify "was this API object already considered for registration?" — WeakSet only tells you "is currently registered", not "was attempted"

### The Dual-Track Solution

| Track | Data Structure | Purpose |
|-------|---------------|---------|
|  |  | Phase 2 singleton guard — GC-safe, prevents double-init |
|  |  | Explicit claim/rollback — inspectable, supports retry logic |

**Registration flow now:**


### Concrete Benefits

1. **Testability** —  lets tests assert "this API was attempted" even after a failed init
2. **Debuggability** — explicit claim/rollback creates a traceable audit trail
3. **Retry safety** —  provides the ground truth for "should this API be retried?"
4. **Future extensibility** — the Map can carry metadata (timestamp, retry count, error reason) that a WeakSet structurally cannot

PR #732, opened 2026-05-01
fix(backup): resolve backup dir from openclaw.json home field

jlin53882:
Backup path was derived from resolvedDbPath/../backups, which crashed with ERR_INVALID_ARG_TYPE when dbPath was unset (undefined).

Now reads openclaw.json directly to resolve openclawHome, respecting the config's 'home' field, with fallback to OPENCLAW_HOME/~/.openclaw/. This also fixes the path=undefined root cause.

Fixes: backup TypeError when path is undefined

---


jlin53882:
Closed — superseded by #734 (includes backup fix)

PR #757, opened 2026-05-06
fix(store): replace require() with dynamic import() for ESM compatibility

ggzeng:
## Summary

Replace `require("@lancedb/lancedb")` with dynamic `import("@lancedb/lancedb")` in `src/store.ts` to fix `ReferenceError: require is not defined` in ESM contexts.

## Problem

Since the plugin's `package.json` declares `"type": "module"`, the entire package runs in ESM mode where `require()` is not available. This causes:

- Retrieval initialization to fail (`retrieval: FAIL`)
- All recall operations to fail (`recall failed: ReferenceError: require is not defined`)
- Backup operations to fail when they need LanceDB access

## Fix

One-line change:
```diff
- lancedbImportPromise = Promise.resolve(require("@lancedb/lancedb"));
+ lancedbImportPromise = import("@lancedb/lancedb");
```

`import()` works for both CJS and ESM packages across all platforms.

## Testing

After applying this fix locally on OpenClaw 2026.5.4:
```
memory-lancedb-pro: initialized successfully (embedding: OK, retrieval: OK, mode: hybrid, FTS: enabled)
```

Previously:
```
memory-lancedb-pro: initialized successfully (embedding: OK, retrieval: FAIL, mode: hybrid, FTS: disabled)
```

Fixes #756

Issue #756, opened 2026-05-06
fix(store): require() fails in ESM context — use dynamic import() for @lancedb/lancedb


ggzeng:
### Summary

In `src/store.ts`, the `loadLanceDB()` function uses `require("@lancedb/lancedb")` to load the LanceDB module. However, the plugin's `package.json` declares `"type": "module"`, making the entire package ESM. In ESM context, `require` is not defined, causing a `ReferenceError` at runtime.

This affects **retrieval, recall, and backup** — all operations that need LanceDB access.

### Environment

- **memory-lancedb-pro**: 1.1.0-beta.10 (master HEAD 47b635d)
- **OpenClaw**: 2026.5.4
- **Node.js**: v24.14.0
- **OS**: Linux (Docker Ubuntu 24.04)

### Reproduction

1. Install memory-lancedb-pro v1.1.0-beta.10
2. Start OpenClaw gateway
3. Observe logs:

```
memory-lancedb-pro: initialized successfully (embedding: OK, retrieval: FAIL, mode: hybrid, FTS: disabled)
memory-lancedb-pro: retrieval test failed: require is not defined
```

On any recall attempt:
```
memory-lancedb-pro: recall failed: ReferenceError: require is not defined
```

### Root Cause

In `src/store.ts`, lines ~27-29:

```typescript
async function loadLanceDB() {
  if (!lancedbImportPromise) {
    // Use require() for CommonJS modules on Windows to avoid ESM URL scheme issues
    lancedbImportPromise = Promise.resolve(require("@lancedb/lancedb"));
  }
```

The comment references Windows ESM URL scheme issues, but `require()` is unavailable in ESM regardless of platform. The `@lancedb/lancedb` package has `"type": null` (CJS), so it can be loaded via dynamic `import()` without issues.

### Fix

Replace `require()` with dynamic `import()`:

```typescript
async function loadLanceDB() {
  if (!lancedbImportPromise) {
    lancedbImportPromise = import("@lancedb/lancedb");
  }
```

One-line change. `import()` works for both CJS and ESM packages and is compatible with all Node.js versions that OpenClaw supports.

### Verified

After applying this fix locally, retrieval and recall work correctly:

```
memory-lancedb-pro: initialized successfully (embedding: OK, retrieval: OK, mode: hybrid, FTS: enabled)
```

### Note

[...]

PR #759, opened 2026-05-06
fix(esm): replace esbuild __require shim with native dynamic import

### LINKED ISSUE #758: dist/index.js ESM bundle throws "Dynamic require of \"@lancedb/lancedb\" is not supported" at runtime
### Summary

When the published ESM bundle (`dist/index.js`) is loaded in pure-ESM Node (the plugin's own `package.json` declares `"type": "module"`), `loadLanceDB()` fails with:

```
Error: Dynamic require of "@lancedb/lancedb" is not supported
  at file://.../dist/index.js:5:9
  at loadLanceDB (file://.../dist/index.js:475:44)
  at MemoryStore.doInitialize (file://.../dist/index.js:667:27)
  at MemoryStore.ensureInitialized (file://.../dist/index.js:659:29)
  at MemoryStore.stats (file://.../dist/index.js:1112:16)
```

Reproducible from a fresh install — every memory-store stats / init code path hits it (e.g. `openclaw doctor`).

### Root cause

`dist/index.js` ships with esbuild's default `__require` shim at the top of the file:

```js
var __require = ((x) => typeof require !== "undefined" ? require
  : typeof Proxy !== "undefined" ? new Proxy(x, {
      get: (a, b) => (typeof require !== "undefined" ? require : a)[b]
    })
  : x)(function(x) {
  if (typeof require !== "undefined") return require.apply(this, arguments);
  throw Error('Dynamic require of "' + x + '" is not supported');
});
```

In an ESM module loaded by Node's ESM loader, the global `require` is **undefined**, so the shim falls through to the throw. The bundled `loadLanceDB` then calls `__require("@lancedb/lancedb")` and immediately fails.

The root cause is in `src/store.ts`:

```ts
lancedbImportPromise = Promise.resolve(require("@lancedb/lancedb"));
```

This bare `require()` call causes esbuild to inject its CJS compatibility shim, which does not work in a pure-ESM host. The `loadLockfile` function in the same file already uses native `import()` correctly — `loadLanceDB` should do the same.

### Repro

1. Install `memory-lancedb-pro@1.1.0-beta.10` into an OpenClaw setup.
2. Run `openclaw doctor`.
3. → `Failed to get statistics: Error: Dynamic require of "@lancedb/lancedb" is not supported`

- Node: v22.22.0 (also reproduced on v26.0.0).
- Platform: Linux arm64.
- `@lancedb/lancedb` and the matching native binary (`lancedb-linux-arm64-gnu`) are correctly installed in `node_modules` — this is **not** a missing-dep issue, it's a build-output issue.

### Suggested fix

Replace the bare `require()` call in `loadLanceDB()` with a native ESM dynamic import:

```ts
export const loadLanceDB = async (): Promise<typeof import("@lancedb/lancedb")> => {
  if (!lancedbImportPromise) {
    lancedbImportPromise = import("@lancedb/lancedb");
  }
  try {
    return await lancedbImportPromise;
  } catch (err) {
    throw new Error(
      `memory-lancedb-pro: failed to load LanceDB. ${String(err)}`,
      { cause: err },
    );
  }
};
```

This matches how `loadLockfile` already loads `proper-lockfile` in the same file, and works regardless of bundler configuration.

Optional follow-up: pass `--external:@lancedb/lancedb` to the esbuild invocation so the native module is never bundled at all. Not strictly required once `loadLanceDB` uses native `import()`, but cleaner long-term.

### Local workaround (for users blocked while waiting for the upstream fix)

Patch `dist/index.js` lines 1–6 to:

```js
import { createRequire as __createRequire_shim } from "module";
var __require = __createRequire_shim(import.meta.url);
```

Confirmed working — `openclaw doctor` passes, plugin initializes normally.

### Environment

- OpenClaw: 2026.5.5 (b1abf9d)
- Node: v22.22.0
- OS: Linux arm64 (Ubuntu)
- Plugin: `memory-lancedb-pro@1.1.0-beta.10`

---

PR: #759

isaacsummers:
## Summary

`dist/index.js` ships with esbuild's default `__require` shim, which throws
`Dynamic require of "@lancedb/lancedb" is not supported` when the bundle is
loaded as pure ESM (Node 22+, package.json declares `"type": "module"`).

The root cause is a bare `require("@lancedb/lancedb")` call in `src/store.ts`
inside `loadLanceDB()`. esbuild converts that to a `__require()` call, and
when the bundle loads under Node's ESM loader, the global `require` is
undefined — so the shim throws.

This PR replaces the bare `require()` call in `loadLanceDB()` with a native
`import("@lancedb/lancedb")`, which works regardless of how the file is
bundled. This aligns with how `loadLockfile` already loads `proper-lockfile`
in the same file.

Fixes #758

## Repro

See linked issue.

## Verification

- `npm install`
- `npm run test:cli-smoke` — passes locally
- Confirmed via fresh `openclaw doctor` run that `MemoryStore.stats` and
  `MemoryStore.doInitialize` paths no longer throw.

## Notes for maintainer

- Source-only change. The committed `dist/` bundle is not regenerated in
  this PR (no public build script in repo). Maintainer's release
  pipeline will pick up the fix in the next published bundle.
- Optional follow-up: pass `--external:@lancedb/lancedb` to the esbuild
  invocation so the native module is never bundled. Not strictly needed
  once `loadLanceDB` uses native `import()`, but cleaner.

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -75,8 +75,7 @@ export const loadLanceDB = async (): Promise<
   typeof import("@lancedb/lancedb")
 > => {
   if (!lancedbImportPromise) {
-    // Use require() for CommonJS modules on Windows to avoid ESM URL scheme issues
-    lancedbImportPromise = Promise.resolve(require("@lancedb/lancedb"));
+    lancedbImportPromise = import("@lancedb/lancedb");
   }
   try {
     return await lancedbImportPromise;


isaacsummers:
Closing as a duplicate of #757 (filed earlier). Detailed root-cause and Windows-safe fix posted as a comment on #757. Issue #758 closed in favor of #756.

PR #743, opened 2026-05-04
fix(backup+admission): drop redundant api.resolvePath on already-absolute paths (Issue #682, supersedes PR #695)

jlin53882:
## Summary

Fixes **Issue #682** — two code paths crashed on OpenClaw 2026.4.x with:

```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type
string or an instance of Buffer or URL. Received undefined.
```

**This PR supersedes PR #695** (closed 2026-05-04 by maintainer due to inactivity after review).

---

## Root Cause

OpenClaw 2026.4.x tightened the plugin API: calling `api.resolvePath()` on an **already-absolute path** that points outside the agent workspace root now returns `undefined`.

Two call sites had this pattern:

### 1. `runBackup()` — plugin internal timer backup

`resolvedDbPath` is produced by `api.resolvePath(...)` at plugin init and is always absolute. Inside `runBackup()`:

```ts
// BEFORE (crashes on 2026.4.x)
const backupDir = api.resolvePath(
  join(resolvedDbPath, "..", "backups"),
);
// api.resolvePath(absolute-path) → undefined → mkdir crashes
```

**Fix:** `resolvedDbPath` is already absolute — join directly without re-wrapping.

**Defensive guard:** Additionally guards against `resolvedDbPath` being `undefined` (edge case where `api.resolvePath()` returns `undefined` for empty-string input rather than throwing), preventing the TypeError at the `join()` call site.

### 2. `createAdmissionRejectionAuditWriter()` — admission audit sidecar

Same redundant wrapper pattern:

```ts
// BEFORE (same crash)
const filePath = api.resolvePath(
  resolveRejectedAuditFilePath(resolvedDbPath, config.admissionControl),
);
```

`resolveRejectedAuditFilePath()` returns:

| Scenario | Return value | Re-resolve needed? |
|----------|-------------|-------------------|
| No explicit config | `join(dbPath, "..", "admission-audit", "rejections.jsonl")` — already absolute | NO |
| Explicit relative path | `"data/audit/rejections.jsonl"` — caller must resolve | YES |
| Explicit absolute path | `"/var/log/memory/rejections.jsonl"` — must NOT be re-resolved | NO |

**Fix:** Guard with `startsWith("/")` — only resolve relative paths.

---

### `index.ts` — `runBackup()` (~line 4015)

```diff
  async function runBackup() {
    try {
+     // ── Defensive: guard against undefined resolvedDbPath ─────────────────
[...]

PR #764, opened 2026-05-08
[codex] fix backup path and late auto-recall results

TurboTheTurtle:
## Summary

Fix two runtime reliability issues in the memory plugin:

- Write daily JSONL backups beside the already-resolved DB path instead of passing an already-resolved path back through `api.resolvePath`.
- Drop auto-recall work that finishes after the timeout so late retrieval results cannot update injection metadata, pending recall state, or log/inject stale context after the hook has already returned.

## Why

The backup path was being derived with `api.resolvePath(join(resolvedDbPath, "..", "backups"))`. When `resolvedDbPath` was already absolute, host path resolution could produce an invalid nested path. Backup failures also lacked enough stage/path context to diagnose quickly.

The auto-recall timeout returned control to the host, but the underlying async retrieval chain could keep running. If it finished later, it could still mutate recall metadata and proceed toward context injection despite the timeout decision.

## Validation

- `node --test test/recall-text-cleanup.test.mjs`
- `node --test test/auto-recall-query-length.test.mjs`
- `node --test test/per-agent-auto-recall.test.mjs`

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -296,6 +296,15 @@ function resolveEnvVars(value: string): string {
   });
 }
 
+function formatErrorForLog(err: unknown): string {
+  if (err instanceof Error) {
+    const code = (err as { code?: unknown }).code;
+    const codeSuffix = typeof code === "string" && code ? ` [${code}]` : "";
+    return `${err.name}: ${err.message}${codeSuffix}`;
+  }
+  return String(err);
+}
+
 function resolveFirstApiKey(apiKey: string | string[]): string {
   const key = Array.isArray(apiKey) ? apiKey[0] : apiKey;
   if (!key) {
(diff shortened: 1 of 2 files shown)


rwmjhb:
Closing this after maintainer triage.

The PR combines unrelated fixes: backup path handling and late auto-recall result handling. The backup path portion overlaps with #743/#773, while the auto-recall timeout behavior should be reviewed as a separate, focused PR with targeted tests. Please split the auto-recall part into its own PR if still needed.

rwmjhb:
Closed after triage; please split into focused PRs if needed.

TurboTheTurtle:
Thanks for the triage. I split the auto-recall timeout behavior into a focused replacement PR: #790. The backup path changes are omitted here since they overlap with #743/#773.

PR #765, opened 2026-05-08
suppress derived reflection after new session

### LINKED ISSUE #718: Bug: /new can leak prior-session memoryReflection context into the next session
## Summary
After upgrading OpenClaw to the latest version, using `/new` in Telegram direct chats can cause the assistant to continue replying to the previous conversation instead of starting a clean new session.

This started only after the latest OpenClaw update. The current suspicion is that `memoryReflection` state is being generated on `command:new` and then injected back into the next session through `before_prompt_build`, especially when `injectMode` is set to `inheritance+derived`.

## Environment
- OpenClaw: `2026.4.26`
- Channel: Telegram direct chat
- memory-lancedb-pro: `1.1.0-beta.10`
- Plugin build tag seen in logs: `memory-lancedb-pro-diag-20260308-0058`
- Config:
  - `smartExtraction: true`
  - `sessionStrategy: "memoryReflection"`
  - `memoryReflection.injectMode: "inheritance+derived"`

## Observed behavior
After `/new`, the next user message may still receive a reply that continues or references the previous conversation, as if the new session inherited too much execution context.

This is not just broad long-term memory recall. It behaves more like previous-session reflection/derived context is being re-injected into the fresh session.

## Expected behavior
After `/new`, the next session should be clean at the conversation level.
Long-term memory or stable behavioral rules may still be available, but the assistant should not continue the previous session's active reply context.

## Reproduction pattern
1. Use OpenClaw with `memory-lancedb-pro` enabled.
2. Configure:
   - `sessionStrategy: "memoryReflection"`
   - `memoryReflection.injectMode: "inheritance+derived"`
3. Chat with the assistant in Telegram direct chat.
4. Send `/new`.
5. Send a new message on a different topic.
6. Observe that the assistant may still respond as if it is continuing the previous conversation.

## Relevant findings
- The plugin registers these hooks in the current setup:
  - `command:new`
  - `command:reset`
  - `before_prompt_build`
  - `after_tool_call`
  - `session_end`
- The current implementation appears to do this sequence:
  1. `/new` triggers `runMemoryReflection`
  2. reflection/derived slices are generated from the previous session
  3. the next session's `before_prompt_build` injects inherited and/or derived reflection context back into the prompt
- This makes it plausible that prior-session derived context leaks across the `/new` boundary.

## Why this looks like a regression
This issue only started after updating OpenClaw to the latest version.
That suggests an integration regression around hook migration / prompt-build timing / session reset boundaries, rather than a long-standing memory configuration issue.

## Suspected root cause
A session-boundary leak between:
- `command:new` reflection generation for the previous session
- `before_prompt_build` reflection injection for the next session

Possibly one of these:
1. newly generated derived reflection is immediately eligible for injection into the fresh session
2. session-specific cached derived state is not cleared strictly enough across `/new`
3. the new session still resolves previous-session reflection data too aggressively when `injectMode = inheritance+derived`

## Suggested direction
Potential fixes could include one or more of:
- do not inject `derived` reflection into the first turn after `/new`
- limit `/new` carryover to stable `inheritance` only
- make derived reflection strictly session-local and never eligible for the next fresh session
- add an explicit guard so `command:new` output from the just-closed session cannot re-enter the immediately opened session as active prompt context

## Notes
I have not yet reduced this to a minimal standalone test case in code, but the configuration, hook registration, and behavior are consistent with a `memoryReflection` cross-session leakage bug triggered by `/new` on the latest OpenClaw.

TurboTheTurtle:
## Summary

Fix the `/new` and `/reset` session-boundary path so freshly generated `memoryReflection` derived deltas from the just-closed session do not get injected as active `<derived-focus>` context into the immediately opened prompt.

- Track a short-lived per-session derived-reflection suppression window after `command:new` / `command:reset`.
- Keep inherited reflection rules available while suppressing only derived-focus injection.
- Avoid repopulating the same-session derived cache from command-boundary reflection output.
- Add a regression test that simulates `command:new` and verifies inherited rules still inject while derived focus is withheld.

Fixes #718

## Validation

- `node --test test/reflection-bypass-hook.test.mjs`
- `node --test test/command-reflection-guard.test.mjs`
- `git diff --check`

## Notes

`npx tsc -p tsconfig.json` is not available on current `master` because the repository does not include a `tsconfig.json`.

`node --test test/memory-reflection.test.mjs` has existing unrelated failures on current `master` expectations around legacy derived filtering and sessionStrategy defaults; this patch does not touch those paths.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -452,6 +452,9 @@ const DEFAULT_REFLECTION_SESSION_TTL_MS = 30 * 60 * 1000;
 const DEFAULT_REFLECTION_MAX_TRACKED_SESSIONS = 200;
 const DEFAULT_REFLECTION_ERROR_SCAN_MAX_CHARS = 8_000;
 const DEFAULT_SERIAL_GUARD_COOLDOWN_MS = 120_000;
+// After /new or /reset, the just-closed session may have generated fresh
+// derived deltas. Keep those out of the immediately opened prompt window.
+const DEFAULT_REFLECTION_BOUNDARY_DERIVED_SUPPRESSION_MS = 120_000;
 const REFLECTION_FALLBACK_MARKER = "(fallback) Reflection generation failed; storing minimal pointer only.";
 const DIAG_BUILD_TAG = "memory-lancedb-pro-diag-20260308-0058";
 
@@ -469,6 +472,12 @@ type ReflectionErrorState = {
   lastInjectedCount: number;
   signatureSet: Set<string>;
   updatedAt: number;
+};
+
+type ReflectionDerivedSuppressionState = {
+  updatedAt: number;
+  until: number;
+  reason: string;
 };
 
 type EmbeddedPiRunner = (params: Record<string, unknown>) => Promise<unknown>;
@@ -1800,6 +1809,18 @@ function _dedupHookEvent(handlerName: string, event: any): boolean {
   return false; // first occurrence — proceed
 }
 
+function getCommandActionName(action: unknown): string {
+  if (typeof action !== "string") return "";
+  const normalized = action.trim().toLowerCase();
+  if (!normalized) return "";
+  return normalized.split(":").pop() || normalized;
+}
+
+function isSessionBoundaryReflectionAction(action: unknown): boolean {
+  const name = getCommandActionName(action);
+  return name === "new" || name === "reset";
+}
+
 // ============================================================================
 // Phase 2 — Singleton State Management (PR #598)
 // ============================================================================
@@ -1819,6 +1840,7 @@ interface PluginSingletonState {
   // Session Maps — persist across scope refreshes instead of being recreated
   reflectionErrorStateBySession: Map<string, ReflectionErrorState>;
   reflectionDerivedBySession: Map<string, { updatedAt: number; derived: string[] }>;
+  reflectionDerivedSuppressionBySession: Map<string, ReflectionDerivedSuppressionState>;
   reflectionByAgentCache: Map<string, { updatedAt: number; invariants: string[]; derived: string[] }>;
   recallHistory: Map<string, Map<string, number>>;
   turnCounter: Map<string, number>;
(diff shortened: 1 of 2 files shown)


rwmjhb (approved):
# PR #765 Review: suppress derived reflection after new session

**Verdict: APPROVE** | 6 rounds completed | Value: 77% | Size: MEDIUM | Author: TurboTheTurtle

## Value Assessment

**Problem**: Using /new or /reset with memoryReflection injectMode=inheritance+derived can allow freshly generated derived reflection from the just-closed session to be injected into the next prompt, making the new session continue prior context.

| Dimension | Assessment |
|-----------|------------|
| Value Score | 77% |
| Value Verdict | priority-review |
| Issue Linked | true |
| Project Aligned | true |
| Duplicate | false |
| AI Slop Score | 0/6 |
| User Impact | high |
| Urgency | high |

**Open Questions**:
- Issue acknowledgement cannot be confirmed from the provided context because #718 has no labels, assignee, or maintainer comments shown.
- Confirm that the sessionKey used by command:new/command:reset is the same key seen by the immediately following before_prompt_build hook across supported OpenClaw channels.
- The PR is marked stale_base=true, so maintainers should check whether current main has changed the reflection hook paths before deep implementation review.


## Summary

Using /new or /reset with memoryReflection injectMode=inheritance+derived can allow freshly generated derived reflection from the just-closed session to be injected into the next prompt, making the new session continue prior context.

## Evaluation Signals

| Signal | Value |
|--------|-------|
| Blockers | 0 |
| Warnings | 0 |
| PR Size | MEDIUM |
| Verdict Floor | approve |
| Risk Level | high |
| Value Model | codex |
| Primary Model | codex |

## Nice to Have

- F1: Boundary-derived rows become injectable again after 120 seconds
- F2: session_end can clear active boundary suppression
- MR1: Suppression state is in-memory only — process restart within 120s window re-exposes derived leak immediately
- MR2: Test only covers command:new and only the immediate-prompt happy path
- MR3: Suppression `until` is recomputed in the finally block using a later `now`, making effective suppression > 120s

## Recommended Action

Ready to merge.

---
*Reviewed at 2026-05-09T12:29:46Z | 6 rounds | Value: codex | Primary: codex | Adversarial: claude*

Issue #682, opened 2026-04-21
backup failed: TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string or an instance of Buffer or URL. Received undefined


z83546263:
### Plugin Version

1.1.0-beta.10

### OpenClaw Version

2026.4.12-beta.1

### Bug Description

2026-04-21T16:53:14.688+08:00 [plugins] memory-lancedb-pro: backup failed: TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string or an instance of Buffer or URL. Received undefined

### Expected Behavior

2026-04-21T16:53:14.688+08:00 [plugins] memory-lancedb-pro: backup failed: TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string or an instance of Buffer or URL. Received undefined

### Steps to Reproduce

2026-04-21T16:53:14.688+08:00 [plugins] memory-lancedb-pro: backup failed: TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string or an instance of Buffer or URL. Received undefined

### Error Logs / Screenshots

```shell

```

### Embedding Provider

None

AliceLJY:
Thanks for the report. I checked the backup path in current code: the failure is most likely coming from the auto-backup block building:

`api.resolvePath(join(resolvedDbPath, "..", "backups"))`

If `resolvedDbPath` or `api.resolvePath(...)` is undefined in this runtime, `node:path.join()` will throw exactly this `ERR_INVALID_ARG_TYPE`.

Two quick data points would help pin whether this is config parsing or an OpenClaw resolvePath/runtime change:

1. the startup log line that looks like `memory-lancedb-pro@... plugin registered (db: ...)`
2. your `plugins.entries.memory-lancedb-pro.config.dbPath` value, if you set one

This backup failure should not block normal memory reads/writes, but it means daily JSONL backup is not being created. Marking this as a real backup-path bug; the likely fix is to guard `resolvedDbPath`/`resolvePath` and fall back to `getDefaultDbPath()` before scheduling backup.

zeeyuu222:
Still reproducing on **1.1.0-beta.9** + **OpenClaw 2026.4.26**, macOS 15 (Darwin 25.4.0 arm64), Node v24.15.0.

The backup timer fires ~60s after every gateway start/restart and fails 100% of the time:

```
[...]

PR #773, opened 2026-05-09
fix(backup): api.resolvePath() returns undefined on secondary register() calls

zcya198:
## Summary

Fix the daily backup failure caused by `api.resolvePath()` returning `undefined` when `register()` is called with a CLI-metadata API instance (the secondary call).

## Problem

`runBackup()` captures `api.resolvePath` from the `register()` closure. When openclaw calls `register()` multiple times with different API instances — first with the **runtime API** (valid `resolvePath`), then with a **cli-metadata API** (whose `resolvePath` may return `undefined`) — the backup timer's closure captures the wrong API, causing:

```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string
or an instance of Buffer or URL. Received undefined
```

This error appears in logs on every gateway restart, but the backup silently fails.

## Fix

Since `resolvedDbPath` is already an absolute path set during `_initPluginState()`, we can safely use `path.join()` directly — no need for `api.resolvePath()` to re-resolve it:

```diff
-        const backupDir = api.resolvePath(
-          join(resolvedDbPath, "..", "backups"),
-        );
+        const backupDir = join(resolvedDbPath, "..", "backups");
```

`path.join()` always returns a string, so the TypeError is eliminated.

## Verification

After this fix, the backup completes successfully:
```
memory-lancedb-pro: backup completed (2097 entries → /home/icamel/.openclaw/memory/backups/memory-backup-2026-05-09.jsonl)
```

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -4212,9 +4212,7 @@ const memoryLanceDBProPlugin = {
 
     async function runBackup() {
       try {
-        const backupDir = api.resolvePath(
-          join(resolvedDbPath, "..", "backups"),
-        );
+        const backupDir = join(resolvedDbPath, "..", "backups");
         await mkdir(backupDir, { recursive: true });
 
         const allMemories = await store.list(undefined, undefined, 10000, 0);


rwmjhb:
Closing this as superseded by #743.

This PR contains the small backup-path hotfix, but #743 carries the fuller Issue #682 fix covering both `runBackup()` and the admission rejection audit path, with targeted regression coverage. We will continue review and merge work on #743 instead.

rwmjhb:
Closed as superseded by #743.

PR #820, opened 2026-05-25
fix: add build script, refresh stale dist output

TriDefender:
## Problem

OpenClaw plugin install fails with:

> Disabled `memory-lancedb-pro` after plugin update failure; OpenClaw will continue without it. Failed to update memory-lancedb-pro: package install requires compiled runtime output for TypeScript entry ./index.ts: expected ./dist/index.js, ./dist/index.mjs, ./dist/index.cjs, ./index.js, ./index.mjs, ./index.cjs.

## Root Cause

1. `package.json` had no `build` or `prepublishOnly` script — compiled output was manually managed
2. `dist/index.js` and `dist/src/store.js` were stale relative to their TypeScript sources (several commits updated `index.ts` and `src/store.ts` without rebuilding dist)
3. No `.gitattributes` — CRLF/LF line ending drift on Windows

## Changes

- **`package.json`**: Add `"build": "tsc"` and `"prepublishOnly": "npm run build"` scripts
- **`dist/index.js`**: Refreshed (was stale vs `index.ts`)
- **`dist/src/store.js`**: Refreshed (was stale vs `src/store.ts`)
- **`.gitattributes`**: Add `* text=auto eol=lf` + mark `dist/**` as linguist-generated

## Verification

- `npx tsc` compiles cleanly (0 errors)
- `node -e "import('./dist/index.js')"` loads successfully
- `npm pack --dry-run` includes `dist/index.js` (223.6kB) in tarball


TriDefender:
Or you should consider to publish to clawhub with the dist thing so it won't need manual recompile

PR #824, opened 2026-05-26
refresh FTS support before retrieval

### LINKED ISSUE #816: Silent hybrid→vector degradation: cached hasFtsSupport never refreshes after FTS init failure
### Summary

On a fresh deployment with a small corpus and a weak embedder (`nomic-embed-text` via Ollama, ~9 documents), hybrid retrieval silently returns zero results. Two cooperating issues in the FTS init path are responsible: `MemoryStore.hasFtsSupport` is a startup-cached boolean with no per-query refresh, and the underlying `createFtsIndex` call is non-idempotent. Together they produce a silent, persistent failure mode that is not recoverable without a container restart.

### Environment

- Plugin: `memory-lancedb-pro@1.0.32` (also reproducible on `master` / `1.1.0-beta.11`)
- LanceDB: `@lancedb/lancedb@0.26.2`
- Embedder: `nomic-embed-text` via Ollama (768d)
- Verified against `master` at commit `f406acd`
- Runtime: Docker container, Linux host

### Reproduction

1. Deploy the plugin with an empty LanceDB table on a host where init can be interrupted (`docker compose down` partway through init is sufficient to leave a partial on-disk FTS state).
2. Configure the embedder to a nomic-family model (`nomic-embed-text` via Ollama).
3. Ingest 5–10 documents via `cli.ts import` or the `memory_store` tool.
4. Invoke `memory_recall` with any literal token guaranteed to appear in a document.
5. Inspect `memory-pro stats` and note the `FTS support` field.
6. Out-of-band, build the FTS index manually:

   ```ts
   await table.createIndex("text", { config: lancedb.Index.fts({ withPosition: true }) });
   ```

   Confirm via `await table.listIndices()` that the FTS index now exists.
7. Re-run `memory_recall` and `memory-pro stats` without restarting the container.

### Expected behavior

After step 6 succeeds and `listIndices()` confirms the FTS index, subsequent `memory_recall` calls should take the hybrid path and return matching documents. `memory-pro stats` should reflect `FTS support: Yes`.

### Observed behavior

After step 6:

- `memory_recall` returns `No relevant memories found` for tokens that are guaranteed to appear in the indexed corpus.
- `memory-pro stats` continues to report `FTS support: No`.
- The vector-only branch persists for the lifetime of the container. Restarting the container restores hybrid retrieval; nothing short of that does.

#### 1. `hasFtsSupport` is a startup-cached boolean with no per-query refresh

`src/store.ts` (current `master`, commit `f406acd`)

`hasFtsSupport` is a getter over a private boolean that is written in exactly two places — `init()` and the manual `rebuildFtsIndex()` recovery method. No code path re-checks the live table state:

```ts
private ftsIndexCreated = false;
// ...
get hasFtsSupport(): boolean {
  return this.ftsIndexCreated;
}
```

`init()` writes the field once inside a try/catch. Once `false`, it stays `false`:

```ts
// Create FTS index for BM25 search (graceful fallback if unavailable)
try {
  await this.createFtsIndex(table);
  this.ftsIndexCreated = true;
} catch (err) {
  console.warn("Failed to create FTS index, falling back to vector-only search:", err);
  this.ftsIndexCreated = false;
}
```

`src/retriever.ts` reads `this.store.hasFtsSupport` to choose between the `bm25` / `vector` / `hybrid` branches and to populate the `FTS support:` line in `stats`. It never calls `listIndices()`. A repo-wide grep confirms `listIndices` only appears inside `src/store.ts` (in `createFtsIndex` and `rebuildFtsIndex`), and `createFtsIndex` is only called from `init()` and `rebuildFtsIndex()`.

The doc comment on `rebuildFtsIndex()` documents it as the only recovery path:

```ts
/** Rebuild FTS index (drops and recreates). Useful for recovery after corruption. */
```

This is the same manual-recovery gap referenced in #402.

#### 2. `createFtsIndex` omits `replace: true`

`src/store.ts` (current `master`, commit `f406acd`)

The single `table.createIndex` call in `src/store.ts` passes only `{ config }`:

```ts
private async createFtsIndex(table: LanceDB.Table): Promise<void> {
  try {
    // Check if FTS index already exists
    const indices = await table.listIndices();
    const hasFtsIndex = indices?.some(
      (idx: any) => idx.indexType === "FTS" || idx.columns?.includes("text"),
    );

    if (!hasFtsIndex) {
      // LanceDB @lancedb/lancedb >=0.26: use Index.fts() config
      const lancedb = await loadLanceDB();
      await table.createIndex("text", {
        config: (lancedb as any).Index.fts({ withPosition: true }),
      });
    }
  } catch (err) {
    throw new Error(
      `FTS index creation failed: ${err instanceof Error ? err.message : String(err)}`,
    );
  }
}
```

When `listIndices()` reports the index as missing while a partial / stale FTS index actually exists on disk (e.g., a prior container was killed mid-init), the `createIndex` call collides, throws, and the outer `init()` catch sets `ftsIndexCreated = false` — at which point bug #1 takes over for the lifetime of the container. A search for `replace:` in `src/store.ts` returns zero matches.

`rebuildFtsIndex()` works around this by calling `dropIndex` in a loop before re-calling `createFtsIndex`, but that path is only reachable via manual invocation.

### Suggested fixes

Bug #1 — replace the cached getter with a TTL-checked async method that consults the live table:

```ts
private _ftsCheckedAt = 0;
private _ftsCached = false;

async hasFtsSupport(): Promise<boolean> {
  const now = Date.now();
  if (now - this._ftsCheckedAt < 60_000) return this._ftsCached;
  try {
    const indices = await this.table!.listIndices();
    this._ftsCached = indices?.some(
      (idx: any) => idx.indexType === "FTS" || idx.columns?.includes("text"),
    ) ?? false;
  } catch {
    this._ftsCached = false;
  }
  this._ftsCheckedAt = now;
  return this._ftsCached;
}
```

If converting the getter to `async` is too invasive for the retriever call sites, a minimum acceptable fix is to re-check on each vector-only branch entry (or on each `stats` call) so recovery does not require a container restart.

Bug #2 — pass `replace: true` on the create call so a stale partial index cannot wedge the flag:

```ts
await table.createIndex("text", {
  config: (lancedb as any).Index.fts({ withPosition: true }),
  replace: true,
});
```

If `replace` is not supported by the installed LanceDB version, the equivalent is to `dropIndex` defensively in the bare `init()` path — i.e., always take the `rebuildFtsIndex()` code path on cold start, not only on manual recovery.

### Workaround

Until a fix lands:

- After any out-of-band FTS index change, restart the plugin / container so `init()` re-runs.
- For deployments where vector-only is acceptable, raise `candidatePoolSize` and lower `hardMinScore` to mask the regression.
- Avoid `nomic-embed-text` on small corpora — Jina or OpenAI embeddings disguise the degradation because vector recall is strong enough that hybrid loss is not user-visible.

### Impact

When the embedder is weak (a known pathological case for `nomic-embed-text` on small N), loss of FTS leaves vector-only retrieval below the default `hardMinScore` floor of `0.35` and returns zero results. The symptom is indistinguishable from "the embedder is misconfigured" or "the corpus is empty," neither of which is true. The underlying cached-init-state cause is opaque from the agent surface because:

- `console.warn` from `init()` is written to the container log, not the agent.
- `memory-pro stats` reports only the cached value.
- `memory_recall` returns an empty-result envelope with no FTS diagnostic.

The failure mode is non-recoverable without a container restart.

### Relationship to prior issues

[#402](https://github.com/CortexReach/memory-lancedb-pro/issues/402) (closed `completed` 2026-04-05) addressed the missing `withPosition: true` option on the same `createFtsIndex` call. That fix is present in current `master`. This report is about two distinct shortcomings of the same call site / state machine that #402 did not touch:

- Caching — `hasFtsSupport` is still a startup-cached boolean.
- Idempotency — `createIndex` still omits `replace: true`, so a partial / stale on-disk index can still wedge init.

Both are independent of `withPosition` and both reproduce on current `master` per the code references above.

[#574](https://github.com/CortexReach/memory-lancedb-pro/issues/574) (closed `not planned`) covers a tangential `Promise.all` → `Promise.allSettled` request for graceful hybrid degradation and is intentionally not bundled into this report.

TurboTheTurtle:
## Summary
- add an explicit MemoryStore FTS support refresh path while keeping the sync getter contract
- refresh FTS support before retrieval mode selection, BM25 fallback, and stats reporting
- add regressions for recall recovering after cached false and stats updating from live listIndices()

Closes #816.

## Validation
- node --test test/retriever-graceful-degradation.test.mjs
- node --test test/store-list-stats-projection-fallback.test.mjs
- npx tsc --noEmit
- node scripts/verify-ci-test-manifest.mjs
- git diff --check

### DIFF
--- a/src/retriever.ts
+++ b/src/retriever.ts
@@ -572,6 +572,16 @@ export class MemoryRetriever {
     return this._statsCollector;
   }
 
+  private async resolveFtsSupport(): Promise<boolean> {
+    const storeWithRefresh = this.store as MemoryStore & {
+      refreshFtsSupport?: () => Promise<boolean>;
+    };
+    if (typeof storeWithRefresh.refreshFtsSupport === "function") {
+      return await storeWithRefresh.refreshFtsSupport();
+    }
+    return this.store.hasFtsSupport;
+  }
+
   async retrieve(context: RetrievalContext): Promise<RetrievalResult[]> {
     const { query, limit, scopeFilter, category, source } = context;
     const safeLimit = clampInt(limit, 1, 20);
@@ -607,6 +617,9 @@ export class MemoryRetriever {
     try {
       // Create trace only when stats collector is active (zero overhead otherwise)
       const trace = this._statsCollector ? new TraceCollector() : undefined;
+      const hasFtsSupport = this.config.mode === "vector"
+        ? this.store.hasFtsSupport
+        : await this.resolveFtsSupport();
 
       // Check if query contains tag prefixes -> use BM25-only + mustContain
       const tagTokens = this.extractTagTokens(query);
@@ -621,7 +634,7 @@ export class MemoryRetriever {
           trace,
           diagnostics,
         );
-      } else if (this.config.mode === "vector" || !this.store.hasFtsSupport) {
+      } else if (this.config.mode === "vector" || !hasFtsSupport) {
         results = await this.vectorOnlyRetrieval(
           query,
           safeLimit,
@@ -649,7 +662,7 @@ export class MemoryRetriever {
       if (trace && this._statsCollector) {
         const mode = tagTokens.length > 0
           ? "bm25"
-          : (this.config.mode === "vector" || !this.store.hasFtsSupport)
+          : (this.config.mode === "vector" || !hasFtsSupport)
             ? "vector"
             : "hybrid";
         const finalTrace = trace.finalize(query, mode);
@@ -685,12 +698,15 @@ export class MemoryRetriever {
 
     const tagTokens = this.extractTagTokens(query);
     let results: RetrievalResult[];
+    const hasFtsSupport = this.config.mode === "vector"
+      ? this.store.hasFtsSupport
+      : await this.resolveFtsSupport();
 
     if (tagTokens.length > 0) {
       results = await this.bm25OnlyRetrieval(
         query, tagTokens, safeLimit, scopeFilter, category, trace,
       );
-    } else if (this.config.mode === "vector" || !this.store.hasFtsSupport) {
+    } else if (this.config.mode === "vector" || !hasFtsSupport) {
       results = await this.vectorOnlyRetrieval(
         query, safeLimit, scopeFilter, category, trace,
       );
@@ -701,7 +717,7 @@ export class MemoryRetriever {
     }
 
     const mode = tagTokens.length > 0 ? "bm25"
-      : (this.config.mode === "vector" || !this.store.hasFtsSupport) ? "vector" : "hybrid";
+      : (this.config.mode === "vector" || !hasFtsSupport) ? "vector" : "hybrid";
     const finalTrace = trace.finalize(query, mode);
 
     if (this._statsCollector) {
@@ -1786,13 +1802,13 @@ export class MemoryRetriever {
       return {
         success: true,
         mode: this.config.mode,
-        hasFtsSupport: this.store.hasFtsSupport,
+        hasFtsSupport: await this.resolveFtsSupport(),
       };
     } catch (error) {
       return {
         success: false,
         mode: this.config.mode,
-        hasFtsSupport: this.store.hasFtsSupport,
+        hasFtsSupport: await this.resolveFtsSupport(),
         error: error instanceof Error ? error.message : String(error),
       };
     }
--- a/src/store.ts
+++ b/src/store.ts
@@ -218,6 +218,13 @@ function isExplicitDenyAllScopeFilter(scopeFilter?: string[]): boolean {
   return Array.isArray(scopeFilter) && scopeFilter.length === 0;
 }
 
+function hasFtsIndex(indices: unknown): boolean {
+  return Array.isArray(indices) && indices.some((idx: any) =>
+    idx?.indexType === "FTS" ||
+    (Array.isArray(idx?.columns) && idx.columns.includes("text")),
+  );
+}
+
 function scoreLexicalHit(query: string, candidates: Array<{ text: string; weight: number }>): number {
   const normalizedQuery = normalizeSearchText(query);
   if (!normalizedQuery) return 0;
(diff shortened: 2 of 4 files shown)


rwmjhb (approved):
# PR #824 Review: refresh FTS support before retrieval

**Verdict: APPROVE** | 6 rounds completed | Value: 64% | Size: MEDIUM | Author: TurboTheTurtle

## Value Assessment

**Problem**: Hybrid retrieval can remain stuck in vector-only mode after FTS index initialization fails because FTS support is cached and not refreshed from the live LanceDB index state. This causes memory recall and stats to keep reporting no FTS support until process restart even after an index becomes available.

| Dimension | Assessment |
|-----------|------------|
| Value Score | 64% |
| Value Verdict | review |
| Issue Linked | true |
| Project Aligned | true |
| Duplicate | false |
| AI Slop Score | 0/6 |
| User Impact | high |
| Urgency | medium |

**Scope Drift**: 1 flag(s)
- test/retriever-graceful-degradation.test.mjs changes buildResult timestamp from fixed 1700000000000 to Date.now(), which is not directly justified by the FTS refresh bug

**Open Questions**:
- Issue #816 has no maintainer labels, assignment, or comments in the supplied context, so maintainer acknowledgment is not established.
- Should createIndex also use replace: true or defensive drop/recreate to address the second idempotency concern described in issue #816?


## Summary

Hybrid retrieval can remain stuck in vector-only mode after FTS index initialization fails because FTS support is cached and not refreshed from the live LanceDB index state. This causes memory recall and stats to keep reporting no FTS support until process restart even after an index becomes available.

## Evaluation Signals

| Signal | Value |
|--------|-------|
| Blockers | 0 |
| Warnings | 1 |
| PR Size | MEDIUM |
| Verdict Floor | approve |
| Risk Level | high |
| Value Model | codex |
| Primary Model | codex |

## Nice to Have

- F2: FTS index creation is still not explicitly idempotent
- EF1: Static analysis reports one new any type
- MR1: Transient listIndices() error now flips a healthy FTS store into vector-only mode (new degradation path)
- MR2: Recovery only re-detects an index; it never re-attempts creation, so #816's exact 'after FTS init failure' scenario is not actually recovered
- MR3: listIndices() I/O round-trip added to the retrieval hot path and to every stats() call
- MR4: Test fixture switched to non-deterministic Date.now(), unrelated to the FTS fix and left unaddressed after R1 flagged it

## Recommended Action

Ready to merge.

---
*Reviewed at 2026-05-26T06:05:56Z | 6 rounds | Value: codex | Primary: codex | Adversarial: claude*

PR #867, opened 2026-06-08
fix: update legacy slug memory ids

ScientificProgrammer:
## Summary

- allow `MemoryStore.update()` to resolve exact non-hex legacy IDs imported from older stores
- preserve existing full UUID and short hex-prefix update behavior
- add a LanceDB regression test for imported slug-style memory IDs

## Context

Legacy imports can preserve stable IDs such as `data-pointer-openclaw-paths-20260527`. `importEntry()` and `getById()` already support those IDs, but `update()` rejected anything that was not a full UUID or hex prefix. That caused `memory-pro upgrade` to fail with `Invalid memory ID format` when upgrading legacy records in place.

## Test

- `node test/update-consistency-lancedb.test.mjs`

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -875,16 +875,13 @@ export class MemoryStore {
     }
 
     return this.runWithFileLock(() => this.runSerializedUpdate(async () => {
-      // Support both full UUID and short prefix (8+ hex chars), same as delete()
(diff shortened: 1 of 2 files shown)


ScientificProgrammer:
Superseded by a new PR prepared from the canonical checkout with a fixture-backed reprex and storage CI validation.

PR #869, opened 2026-06-09
fix: declare memory_reflection_resolve contract and remove duplicate contracts key

Grokon:
## Problem

The plugin registers `memory_reflection_resolve` via `registerMemoryReflectionResolveTool()` called from `registerAllMemoryTools()` in `src/tools.ts`, but the tool was missing from `contracts.tools` in `openclaw.plugin.json`.

This causes a gateway warning on every start:
```
plugin must declare contracts.tools for: memory_reflection_resolve
```

## Changes

1. **Add `memory_reflection_resolve` to `contracts.tools`** — placed with the other `memory_*` tools, after `memory_explain_rank`.

2. **Remove the duplicate top-level `contracts` key** (line 2 of the file). JSON allows duplicate object keys but parsers silently use the last one, making the first declaration dead code. The canonical `contracts` declaration is the one at the end of the file (after `configSchema`).

## Verification

`memory_reflection_resolve` is defined and registered:
- `src/tools.ts` — `registerMemoryReflectionResolveTool()` defined, registered via `registerAllMemoryTools()`
- `index.ts` — `registerAllMemoryTools` imported and called on plugin init

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -1,24 +1,4 @@
 {
-  "contracts": {
-    "tools": [
-      "memory_recall",
-      "memory_search",
-      "memory_get",
-      "memory_store",
-      "memory_forget",
-      "memory_update",
-      "memory_stats",
-      "memory_debug",
-      "memory_list",
-      "memory_promote",
-      "memory_archive",
-      "memory_compact",
-      "memory_explain_rank",
-      "self_improvement_log",
-      "self_improvement_extract_skill",
-      "self_improvement_review"
-    ]
-  },
   "id": "memory-lancedb-pro",
   "name": "Memory (LanceDB Pro)",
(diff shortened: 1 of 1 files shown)


rwmjhb (approved):
Approved.

The manifest change is narrow and matches the registered tool contract: `memory_reflection_resolve` is now declared, and the duplicate top-level `contracts` key has been removed.

Verification passed:

```bash
# orchestrator R0
full npm test: PASS

# manual manifest/package checks
rg -n '^\s*"contracts"\s*:' openclaw.plugin.json | wc -l   # 1
jq -e '.contracts.tools | index("memory_reflection_resolve") != null' openclaw.plugin.json
npm run test:packaging-and-workflow
```

A future regression test that checks registered tools against `contracts.tools` would be useful, but this PR is safe as-is.

PR #870, opened 2026-06-09
fix: accept wrapped plugin config entries

### LINKED ISSUE #651: : memory-lancedb-pro fails to initialize: embedding config validation error
### Plugin Version

1.10.beta.9

### OpenClaw Version

2026.4.11

## Bug Description

Plugin `memory-lancedb-pro` registers successfully but fails to initialize with a JSON Schema validation error:

memory-lancedb-pro invalid config: embedding: must have required property "embedding"

### Expected Behavior

Plugin initializes and connects to LanceDB with the provided Jina embeddings config.

### Steps to Reproduce

Steps to Reproduce

1. Install memory-lancedb-pro plugin
2. Configure with embedding provider set to openai-compatible + Jina v5 model
3. Run any OpenClaw command that triggers plugin initialization

Observed Behavior

• Plugin registers (plugin registered log appears)
• openclaw memory CLI fails because plugin fails to initialize
• Memory stays unavailable despite enabled: true

### Error Logs / Screenshots

```shell
Notes

• Config reads back as correct via openclaw config get
• Error suggests the plugin's ajv schema checks for a nested embedding.embedding property that doesn't exist in the config structure — likely a schema mismatch between plugin version and OpenClaw version
```

### Embedding Provider

None

### OS / Platform

_No response_

TurboTheTurtle:
## Summary
- accept OpenClaw plugin entry wrapper objects before validating the plugin config
- preserve direct plugin config precedence when both direct and nested config shapes are present
- add regression coverage for the Jina/openai-compatible config shape from the issue

Fixes #651

## Verification
- node --test test/config-session-strategy-migration.test.mjs
- npm run build

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -4004,7 +4004,13 @@ export function parsePluginConfig(value) {
         throw new Error("memory-lancedb-pro: plugin config must be an object with a top-level embedding block at " +
             "plugins.entries.memory-lancedb-pro.config.embedding.");
     }
-    const cfg = value;
+    const initialCfg = value;
+    const cfg = !initialCfg.embedding &&
+        initialCfg.config &&
+        typeof initialCfg.config === "object" &&
+        !Array.isArray(initialCfg.config)
+        ? initialCfg.config
+        : initialCfg;
     const embedding = cfg.embedding;
     if (!embedding) {
         throw new Error("memory-lancedb-pro: missing top-level config.embedding block. " +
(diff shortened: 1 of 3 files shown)


rwmjhb (approved):
Approved.

The fix is focused and covers the reported OpenClaw plugin-entry wrapper shape without changing the existing direct-config precedence. Verification passed:

```bash
# orchestrator R0
full npm test: PASS

npm run build --if-present
```

Only low-risk follow-ups noted by review: wrapper detection is intentionally broad (`config` object + no top-level `embedding`), and falsy top-level `embedding` can fall through to nested config. Neither blocks this fix for the reported valid wrapper case.

Issue #368, opened 2026-03-26
memory-lancedb-pro causes Telegram providers to stop starting


yyxp1989:
### Plugin Version

1.1.0-beta.10

### OpenClaw Version

 2026.3.23-2 (7ffe7e4)

### Bug Description

Bug report                                                                                 
                                                                                            
 ### OpenClaw version                                                                       
                                                                                            
 2026.3.23-2                                                                                
                                                                                            
 ### Previous version / migration source                                                    
                                                                                            
 Config migrated from 2026.2.18                                                             
                                                                                            
 ### Environment                                                                            
                                                                                            
 - Debian/Linux                                                                             
 - Node v22.22.1                                                                            
                                                                                            
 ### What happened                                                                          
                                                                                            
 After enabling memory-lancedb-pro, all Telegram bots stopped responding.                   
                                                                                            
 At the same time:                                                                          
 - OpenClaw gateway remained functional                                                     
 - TUI remained functional                                                                  
 - memory-lancedb-pro initialized successfully                                              
                                                                                            
 Disabling memory-lancedb-pro restored Telegram immediately.                                
                                                                                            
                                                         
 ────────────────────────────────────────────────────────────────────────────────           
                                                                                            
 ### Actual behavior                                                                        
                                                                                            
 When memory-lancedb-pro is enabled:                                                        
 - Telegram bots do not respond                                                             
 - no Telegram provider startup logs appear                                                 
 - gateway/TUI still work                                                                   
                                                                                            
 When the plugin is disabled:                                                               
 - Telegram works normally again                                                            
                                                                                            
 ────────────────────────────────────────────────────────────────────────────────           
                                                                                            


### Expected Behavior

 ### Expected behavior                                                                      
                                                                                            
 Telegram providers should still start and respond normally when memory-lancedb-pro is      
 enabled.            

### Steps to Reproduce

 ### What I already tested                                                                  
                                                                                            
 #### 1. Config schema migration                                                            
                                                                                            
 Checked current docs for 2026.3.23-2:                                                      
 - channels.telegram still valid                                                            
 - bindings still valid                                                                     
 - plugins.entries still valid                                                              
                                                                                            
 So this does not appear to be caused by obsolete config categories from 2026.2.18.         
                                                                                            
 #### 2. Telegram multi-account config                                                      
                                                                                            
 Reduced Telegram to a minimal single-account config using only:                            
 - channels.telegram.enabled                                                                
[...]

PR #872, opened 2026-06-09
fix: warn when Telegram plugin is disabled

TurboTheTurtle:
## Summary
- add a startup diagnostic for the reported Telegram channel/plugin mismatch
- warn when `channels.telegram.enabled` is true but the Telegram plugin entry is explicitly disabled
- include the actionable `openclaw plugin enable telegram` recovery command

Refs #368

## Verification
- node --test test/telegram-plugin-status-diagnostic.test.mjs
- npm run build

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -1767,6 +1767,41 @@ export function isAgentOrSessionExcluded(agentId, sessionKey, patterns) {
     }
     return false;
 }
+const _channelPluginDiagnosticWarnings = new Set();
+function readRecord(value) {
+    return value && typeof value === "object" && !Array.isArray(value)
+        ? value
+        : undefined;
+}
+function isChannelEnabled(config, channelName) {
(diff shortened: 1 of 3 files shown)


rwmjhb (approved):
Approved.

The diagnostic is limited in scope, but it is safe and useful for the explicit disabled-plugin configuration this PR targets. I also checked the main review caveat manually: `api.config` is already used elsewhere in the plugin runtime path, so this PR is not introducing that dependency from scratch.

Verification passed:

```bash
# orchestrator R0
full npm test: PASS

npm run build --if-present
node --test test/telegram-plugin-status-diagnostic.test.mjs
```

Nice-to-have follow-ups: consider detecting Telegram omission from `plugins.allow`, and consider an integration-level test through `register()` if this diagnostic grows beyond the explicit disabled-plugin case.

PR #876, opened 2026-06-11
fix: normalize memory category filters

### LINKED ISSUE #873: Recall category filter silently empties results: tool schema enum in tools.ts disagrees with the write-side taxonomy in memory-categories.ts
### Summary

`memory_recall` (and sibling tools) advertise a `category` parameter as `stringEnum(MEMORY_CATEGORIES)` imported from [src/tools.ts#L44](https://github.com/CortexReach/memory-lancedb-pro/blob/f406acd/src/tools.ts#L44):

```ts
export const MEMORY_CATEGORIES = [
  "preference", "fact", "decision", "entity", "reflection", "other",
] as const;
```

But the write side (smart-extractor auto-capture) uses a different constant with the same name from [src/memory-categories.ts#L8](https://github.com/CortexReach/memory-lancedb-pro/blob/f406acd/src/memory-categories.ts#L8):

```ts
export const MEMORY_CATEGORIES = [
  "profile", "preferences", "entities", "events", "cases", "patterns",
] as const;
```

The recall filter is an exact string match in [src/retriever.ts#L748-L749](https://github.com/CortexReach/memory-lancedb-pro/blob/f406acd/src/retriever.ts#L748-L749):

```ts
const filtered = category
  ? results.filter((r) => r.entry.category === category)
```

So most schema-legal category values can never match auto-captured rows, and the failure is silent (fewer or zero results, no warning). Note the near-misses differ only in number: the tool offers singular `preference` and `entity`, the store writes plural `preferences` and `entities`.

### Where each taxonomy is used

- Tool schemas use the tools.ts enum: [L581](https://github.com/CortexReach/memory-lancedb-pro/blob/f406acd/src/tools.ts#L581) (memory_recall), [L739](https://github.com/CortexReach/memory-lancedb-pro/blob/f406acd/src/tools.ts#L739) (memory_store), [L1247](https://github.com/CortexReach/memory-lancedb-pro/blob/f406acd/src/tools.ts#L1247), [L1772](https://github.com/CortexReach/memory-lancedb-pro/blob/f406acd/src/tools.ts#L1772).
- Auto-capture writes the memory-categories.ts taxonomy ([smart-extractor imports](https://github.com/CortexReach/memory-lancedb-pro/blob/f406acd/src/smart-extractor.ts#L25-L34), including `normalizeCategory`).
- Reflection rows store `category: "reflection"` ([src/reflection-store.ts#L208](https://github.com/CortexReach/memory-lancedb-pro/blob/f406acd/src/reflection-store.ts#L208)).
- Manual `memory_store` defaults to `category = "other"` ([src/tools.ts#L750](https://github.com/CortexReach/memory-lancedb-pro/blob/f406acd/src/tools.ts#L750)).

### Observed impact

Environment: memory-lancedb-pro v1.1.0-beta.11 (f406acd), OpenClaw 2026.6.1, Linux, Node 24.

An embedded recall agent (OpenClaw active-memory plugin calling `memory_recall`) picked `category: "other"`, a schema-legal value. Recall then searched 9 of 196 stored memories (only the manual stores that had defaulted to "other") and reported no relevant memory while relevant rows existed under other categories. Any LLM tool-caller that uses the advertised enum in good faith gets silently degraded recall.

### Suggested direction

1. Single source of truth for `MEMORY_CATEGORIES` (one module, imported everywhere).
2. Pass the filter value through the existing `normalizeCategory()` before the exact match, so singular/legacy variants map onto stored values.
3. When a category filter reduces candidates to zero, say so in the tool result (for example "category filter matched 0 of N memories; stored categories: ...") so callers can self-correct.

### Related

- [#874](https://github.com/CortexReach/memory-lancedb-pro/issues/874) (scope parameter trap). We hit both in the same incident; filed separately since the fixes are independent.

TurboTheTurtle:
## Summary
- expose both smart-extractor and legacy memory categories in tool schemas
- normalize category filters so legacy singular filters match smart category rows
- apply category candidate matching to memory_list and add CI regression coverage

Fixes #873

## Verification
- node --test test/category-filter-normalization.test.mjs
- npm run build
- npm run test:core-regression

### DIFF
--- a/dist/src/memory-categories.js
+++ b/dist/src/memory-categories.js
@@ -24,6 +24,22 @@ export const TOOL_MEMORY_CATEGORIES = [
     ...MEMORY_CATEGORIES,
     ...LEGACY_MEMORY_CATEGORIES,
 ];
+const SMART_TO_STORAGE_CATEGORY = {
+    profile: "fact",
+    preferences: "preference",
+    entities: "entity",
+    events: "decision",
+    cases: "fact",
+    patterns: "other",
+};
+const LEGACY_TO_SMART_CATEGORY = {
+    preference: "preferences",
+    fact: "cases",
+    decision: "events",
+    entity: "entities",
+    reflection: "patterns",
+    other: "patterns",
+};
 /** Categories that always merge (skip dedup entirely). */
 export const ALWAYS_MERGE_CATEGORIES = new Set(["profile"]);
 /** Categories that support MERGE decision from LLM dedup. */
@@ -58,16 +74,24 @@ export function normalizeCategory(raw) {
     }
     return null;
 }
-export function matchesMemoryCategoryFilter(entryCategory, requestedCategory) {
+export function matchesMemoryCategoryFilter(entryCategory, requestedCategory, entryMetadata) {
     const rawEntryCategory = entryCategory.toLowerCase().trim();
     const rawRequestedCategory = requestedCategory.toLowerCase().trim();
     if (rawEntryCategory === rawRequestedCategory)
         return true;
+    const metadataCategory = extractMetadataMemoryCategory(entryMetadata);
     const normalizedEntryCategory = normalizeCategory(rawEntryCategory);
     const normalizedRequestedCategory = normalizeCategory(rawRequestedCategory);
-    return normalizedEntryCategory !== null &&
-        normalizedRequestedCategory !== null &&
-        normalizedEntryCategory === normalizedRequestedCategory;
+    if (metadataCategory && normalizedRequestedCategory) {
+        return metadataCategory === normalizedRequestedCategory;
+    }
+    if (normalizedEntryCategory && normalizedRequestedCategory) {
+        return normalizedEntryCategory === normalizedRequestedCategory;
+    }
+    if (normalizedRequestedCategory && isLegacyMemoryCategory(rawEntryCategory)) {
+        return LEGACY_TO_SMART_CATEGORY[rawEntryCategory] === normalizedRequestedCategory;
+    }
+    return false;
 }
 export function resolveCategoryFilterCandidates(requestedCategory) {
     const rawRequestedCategory = requestedCategory.toLowerCase().trim();
@@ -75,6 +99,7 @@ export function resolveCategoryFilterCandidates(requestedCategory) {
     const candidates = new Set([rawRequestedCategory]);
     if (normalizedRequestedCategory) {
         candidates.add(normalizedRequestedCategory);
+        candidates.add(SMART_TO_STORAGE_CATEGORY[normalizedRequestedCategory]);
         for (const category of TOOL_MEMORY_CATEGORIES) {
             if (normalizeCategory(category) === normalizedRequestedCategory) {
                 candidates.add(category);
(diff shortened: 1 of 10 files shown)


rwmjhb (changes requested):
Requesting changes. The direction is right, but the current implementation only partially fixes the taxonomy mismatch.

## Must fix

1. Smart-category filters still miss metadata-backed auto-captured rows.

The new retrieval/list filtering still compares only the storage `category` column: retriever uses `matchesMemoryCategoryFilter(r.entry.category, category)`, and `memory_list` expands SQL predicates over `category` via `resolveCategoryFilterCandidates(category)`.

That works for direct aliases like `preference`/`preferences` and `entity`/`entities`, but it does not handle the smart-extractor categories that are stored through legacy storage categories with the real smart category in metadata. In the normal auto-capture path, `profile` and `cases` are stored as `fact`, `events` as `decision`, and `patterns` as `other`, with `metadata.memory_category` carrying the smart category. Filters for `profile`, `events`, `cases`, and `patterns` can therefore still silently miss normal auto-captured rows.

Please make recall/list filtering metadata-aware, or over-fetch the relevant storage categories and then apply `metadata.memory_category` filtering in application code. Add regression coverage through the real smart-extractor/store path for the smart categories, not only direct row stubs.

2. Newly accepted smart category values are not normalized on write.

The tool schema now accepts smart category values such as `profile`, `preferences`, `events`, `cases`, and `patterns` for `memory_store` / `memory_update`. But the write path still passes the raw category into storage and `buildSmartMetadata()` without explicitly setting the normalized smart category. For non-legacy values, `reverseMapLegacyCategory()` cannot infer the intended category reliably and can default to `patterns`.

Please normalize tool input at the execution boundary into both:

- the storage category to persist, and
- `metadata.memory_category` / smart category metadata to preserve semantics.

Also add tests for manual `memory_store` and `memory_update` using schema-legal smart category values.

## Also worth covering

The new test mainly exercises retriever/BM25-style filtering. Please add coverage for the SQL `memory_list` path and keep `resolveCategoryFilterCandidates()` / `matchesMemoryCategoryFilter()` behavior equivalent so these two filter surfaces do not drift.

rwmjhb (approved):
Approved. The new commit addresses the blocking issues from the previous review: category filters are now metadata-aware for recall/list paths, and manual store/update normalize accepted smart category inputs into storage category plus smart metadata.

Verification I checked:

- `npm run build --if-present` passed on head `3beb2781b2dbe0121e506d3eafb68d6c81601e5c`
- `node --test test/category-filter-normalization.test.mjs` passed, 7/7

- `patterns` filtering can still differ between recall and list for legacy `reflection` rows, because the in-memory predicate maps `reflection -> patterns` but the SQL candidate expansion for list does not fetch `reflection`.
- Metadata-less legacy `fact` rows remain ambiguous and are treated as `cases`, so profile-like legacy facts without `memory_category` metadata may not match `category=profile`.
- Category-only `memory_update` changes `memory_category`, but preserves category-derived fields such as `fact_key` and `memory_layer`; re-deriving or clearing those would avoid stale metadata after recategorization.

rickthomasjr (approved):
## Agent Code Review: PR #876 — fix: normalize memory category filters

**Verdict: APPROVE**

### Summary
Two commits introducing bidirectional smart/legacy category mapping. Core changes: memory-categories.ts adds LEGACY_MEMORY_CATEGORIES, TOOL_MEMORY_CATEGORIES, mapping tables, and helper functions (normalizeCategory, matchesMemoryCategoryFilter, resolveCategoryFilterCandidates, resolveToolMemoryCategory, extractMetadataMemoryCategory). Tools and retrieval now normalize incoming categories at the execution boundary, persisting both storageCategory and metadata.memory_category. SQL list() filtering uses candidate OR expansion + in-memory metadata-aware predicate. All retrieval paths use matchesMemoryCategoryFilter() instead of direct equality.

### P1 — Supersede logic: APPROVED
Replacing hardcoded SUPERSEDE_ELIGIBLE with TEMPORAL_VERSIONED_CATEGORIES.has(memoryCategory) + metadata-aware matching prevents false-positive cross-supersedes between unrelated legacy categories. Correct.

### P2 — Metadata-aware filtering: APPROVED
matchesMemoryCategoryFilter() three-tier approach (direct → normalize → metadata) handles the critical case: entries stored as fact with metadata.memory_category=profile are still found by category=profile filters. This was the core issue rwmjhb flagged.

### P3 — SQL candidate expansion: APPROVED
resolveCategoryFilterCandidates() builds OR of all storage category values for a smart category. Set deduplication prevents duplication. Correct.

### P4 — Write path normalization: APPROVED
Both memory_store and memory_update resolve storageCategory + memoryCategory at execution boundary. buildSmartMetadata() uses resolved categories. Category-only updates also update memory_category in metadata.

### P5 — Response payloads: APPROVED
Adding rawCategory alongside category is useful for debugging.

### P6 — Minor notes
- isLegacyStableMemoryId regex in store.ts is minor scope creep in this PR — consider separate PR.
- rwmjhb's non-blocking follow-ups (patterns/legacy reflection gap, ambiguous metadata-less legacy fact rows, stale fields on recategorization) are valid but out of scope.

### Tests
7 test cases, 397 lines. Exercises smart-extractor rows through real store/tool paths. Covers normalization, metadata-aware retrieval, SQL list filtering, store tool writes, update tool writes. Good coverage.

### Conclusion
Well-designed, well-tested fix for a real taxonomy mismatch. The dual-category normalization is the right approach. No blocking issues.

PR #878, opened 2026-06-11
fix: fail soft for read-only recall scopes

### LINKED ISSUE #874: memory_recall scope parameter is a hard trap for LLM callers: valid values are not discoverable from the schema and any unknown value returns scope_access_denied instead of falling back
### Summary

`scope` on `memory_recall` is an optional bare string, described only as "Specific memory scope to search in (optional)" ([src/tools.ts#L576-L580](https://github.com/CortexReach/memory-lancedb-pro/blob/f406acd/src/tools.ts#L576-L580)). Valid values (configured scope definitions plus built-ins like `global` and `agent:<id>`, per [src/scopes.ts validateScope/isAccessible](https://github.com/CortexReach/memory-lancedb-pro/blob/f406acd/src/scopes.ts#L253-L275)) are not enumerated in the schema, and the calling agent has no way to discover them. If a caller passes any value `scopeManager.isAccessible()` rejects, the tool returns an error result `scope_access_denied` ([src/tools.ts#L610-L622](https://github.com/CortexReach/memory-lancedb-pro/blob/f406acd/src/tools.ts#L610-L622)) rather than falling back to the caller's accessible scopes. The same pattern exists in the sibling tools at [L796](https://github.com/CortexReach/memory-lancedb-pro/blob/f406acd/src/tools.ts#L796), [L1099](https://github.com/CortexReach/memory-lancedb-pro/blob/f406acd/src/tools.ts#L1099), [L1545](https://github.com/CortexReach/memory-lancedb-pro/blob/f406acd/src/tools.ts#L1545), and [L1667](https://github.com/CortexReach/memory-lancedb-pro/blob/f406acd/src/tools.ts#L1667).

### Why it bites

LLM tool-callers eagerly fill optional parameters. In our fleet, switching the recall model caused it to invent `scope: "current_conversation"`, a plausible-sounding value invited by the free-string schema. Every recall then returned `scope_access_denied`: zero memory injection, and because the call ran inside an automated pipeline (OpenClaw active-memory embedded recall), the failed run also occupied a serial execution lane until timeout, starving other agents' recalls. The access control itself is correct; the problem is that the schema invites arbitrary strings while the implementation only accepts an undiscoverable allowlist, and the failure mode is terminal for the whole call.

Environment: memory-lancedb-pro v1.1.0-beta.11 (f406acd), OpenClaw 2026.6.1, Linux, Node 24.

### Suggested direction (keeping the security gate intact)

1. Enumerate the calling agent's accessible scopes in the parameter description at tool-registration time (registration already has `runtimeContext` and `agentId`), or use a dynamic enum.
2. On deny, include the accessible scopes in the error details so a caller can retry correctly.
3. Consider fail-soft for read-only recall: ignore an inaccessible scope filter, search the accessible scopes, and note that in the result. A hard deny stays appropriate for writes (memory_store, memory_update).

### Related

- [#231](https://github.com/CortexReach/memory-lancedb-pro/issues/231) (memory_update rejects agent-scoped memories as outside accessible scopes)
- [#378](https://github.com/CortexReach/memory-lancedb-pro/issues/378) (memory_stats and memory_list return 0 when scope not provided)

Both show the same discoverability gap from different angles. We hit this and the category-filter issue ([#873](https://github.com/CortexReach/memory-lancedb-pro/issues/873)) in the same incident; filing separately since the fixes are independent.

TurboTheTurtle:
## Summary
- make read-only memory tools ignore inaccessible scope filters and search accessible scopes instead
- include ignored scope and accessible scopes in text/details so tool callers can retry correctly
- keep write/delete/mutation tools hard-denied for inaccessible scopes

Fixes #874

## Verification
- node --test test/memory-governance-tools.test.mjs
- npm run build
- npm run test:core-regression

### DIFF
--- a/dist/src/tools.js
+++ b/dist/src/tools.js
@@ -159,17 +159,22 @@ function resolveReadableToolScopeFilter(scopeManager, agentId, scope) {
         if (scopeManager.isAccessible(scope, agentId)) {
             return { scopeFilter: [scope] };
         }
+        const accessibleScopes = scopeManager.getAccessibleScopes(agentId);
         return {
-            error: {
-                content: [{ type: "text", text: `Access denied to scope: ${scope}` }],
-                details: {
-                    error: "scope_access_denied",
-                    requestedScope: scope,
-                },
-            },
+            scopeFilter: resolveScopeFilter(scopeManager, agentId),
+            ignoredScope: scope,
+            accessibleScopes,
         };
     }
     return { scopeFilter: resolveScopeFilter(scopeManager, agentId) };
+}
+function formatIgnoredScopeNotice(resolvedScopes) {
+    if (!resolvedScopes.ignoredScope)
+        return undefined;
+    const scopes = resolvedScopes.accessibleScopes?.length
+        ? resolvedScopes.accessibleScopes.join(", ")
+        : "(none)";
+    return `Ignored inaccessible scope "${resolvedScopes.ignoredScope}" and searched accessible scopes instead: ${scopes}.`;
 }
 async function resolveMemoryId(context, memoryRef, scopeFilter) {
     const trimmed = memoryRef.trim();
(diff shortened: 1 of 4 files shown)


rwmjhb (approved):
Approved.

The read-only fail-soft behavior matches the issue direction, while write/mutation tools stay hard-denied for inaccessible scopes. Verification passed:

```bash
# orchestrator R0
full npm test: PASS

npm run build --if-present
node --test test/memory-governance-tools.test.mjs
```

Nice-to-have follow-ups: consider aligning `memory_debug` with the other read-only tools by resolving agent identity from execute-time runtime context, and document that read-only tools now return `ignoredScope`/`accessibleScopes` instead of the old `scope_access_denied` error shape for inaccessible requested scopes.

PR #879, opened 2026-06-11
feat: add optional Redis write lock

### LINKED ISSUE #659: Solution: Redis Distributed Lock for High Concurrency (extends #643)
## Summary

This issue proposes a solution for the high concurrency lock contention problem identified in Issue #643. The solution uses Redis-based distributed lock instead of file lock.

---

## Problem Recap (from #643)

When captureAssistant=true and sessionMemory.enabled=true:
- Multiple agents write to memory simultaneously
- File lock (.memory-write.lock) cannot handle high concurrency
- Test results: 200 concurrent writes -> only 6% success with file lock

---

### Implementation

**New file: src/redis-lock.ts**
- Uses ioredis package
- Token-based lock acquisition (prevents race conditions)
- Lua script for safe release (only releases if token matches)
- Automatic fallback to file lock when Redis unavailable

### Key Features

1. Token-based ownership - Each lock has unique token, prevents accidental release
2. Graceful fallback - If Redis fails, falls back to file lock
3. Configurable TTL - Default 60s TTL, auto-expire
4. Exponential backoff - Smart retry with increasing delays

---

## Required Package

```
npm install ioredis
```

---

### 1. Performance Tests

- test/lock-extreme-concurrent.test.mjs - 10/20/50/200 concurrent operations
- test/lock-production-simulation.test.mjs - Real-world scenario simulation
- test/lock-bottleneck-identification.test.mjs - Find concurrency threshold

### 2. Redis Lock Tests

- test/redis-lock-real.test.mjs - Real Redis with 200 concurrent ops
- test/redis-lock-edge-cases.test.mjs - 10 edge case scenarios

### 3. Optimization Tests

- test/redis-lock-optimized.test.mjs - Different keys run in parallel

---

### File Lock (before)
| Concurrency | Success Rate |
|--------------|---------------|
| 10 | 100% |
| 20 | 55% |
| 50 | 22% |
| 200 | 6% |

### Redis Lock (after)
| Concurrency | Success Rate |
|--------------|---------------|
| 10 | 100% |
| 20 | 100% |
| 50 | 100% |
| 200 | 97.5% |

Improvement: 15x better (6% -> 97.5%)

---

### Single DB vs Multiple DBs
- Single DB path: Operations still queue (must wait for lock)
- Multiple DB paths: Can run in parallel (different locks)
- This is expected behavior - lock must serialize writes to same DB

### Edge Cases Verified (10/10 pass)
1. Redis connection failure -> graceful error
2. Lock acquisition timeout -> proper error message
3. Double release -> safe (no crash)
4. Empty key -> works
5. Special characters in key -> works
6. Operation longer than TTL -> auto-expire works
7. Multiple lock manager instances -> correct behavior
8. Concurrent different locks -> parallel execution
9. Rapid acquire-release cycles -> 3000+/sec
10. Summary test passes

---

## Next Steps

To use this solution:

1. Add to package.json: npm install ioredis
2. Update store.ts to use RedisLockManager when Redis is available
3. Add Redis configuration to openclaw.json: memory.redisUrl

---

## Related Issues

- Extends #643 (continues the work)
- Related to #622, #623 (stale lock issues)

TurboTheTurtle:
## Summary
- add an optional Redis write lock manager using token ownership and Lua safe release
- route LanceDB write paths through Redis when enabled, with file-lock fallback when Redis is unavailable
- expose Redis lock config and ioredis dependency with unit and storage CI coverage

Fixes #659

## Verification
- node --test test/redis-lock.test.mjs
- npm run build
- npm run test:storage-and-schema

### DIFF
--- a/README.md
+++ b/README.md
@@ -689,10 +689,12 @@ OAuth login flow:
 enough for one gateway process, multiple agents through that gateway, and local
 processes that share the same database directory.
 
-Redis is not required for those deployments, and the current plugin does not
-automatically enable Redis locking. For multi-machine or multi-container
-writers, read [Lock Management](docs/lock-management.md) before sharing a
-LanceDB directory across processes.
+Redis is not required for those deployments. Redis locking is enabled when
+`locking.redis.enabled` is true, when `redisUrl`/`locking.redis.url` is set, or
+when `MEMORY_LANCEDB_REDIS_URL` is present in that process environment. For
+multi-machine or multi-container writers, read
+[Lock Management](docs/lock-management.md) before sharing a LanceDB directory
+across processes, and make sure every writer uses the same lock configuration.
 
 </details>
 
--- a/dist/index.js
+++ b/dist/index.js
@@ -139,6 +139,10 @@ function resolveFirstApiKey(api, apiKey) {
     }
     return resolveSecretCredential(api, key, "embedding.apiKey");
 }
+function resolveOptionalEnvString(value) {
+    const raw = asNonEmptyString(value);
+    return raw ? resolveEnvVars(raw) : undefined;
+}
 function resolveOptionalPathWithEnv(api, value, fallback) {
     const raw = typeof value === "string" && value.trim().length > 0 ? value.trim() : fallback;
     return api.resolvePath(resolveEnvVars(raw));
@@ -1662,7 +1666,9 @@ function _initPluginState(api) {
         dbPath: resolvedDbPath,
         vectorDim,
         disableNativeCosine: config.retrieval?.disableNativeCosine === true,
+        redisLock: config.locking?.redis,
         onStoragePathWarning: (message) => api.logger.warn(message),
+        onLockWarning: (message) => api.logger.warn(message),
     });
     const embedder = createEmbedder({
         provider: "openai-compatible",
@@ -4220,6 +4226,18 @@ export function parsePluginConfig(value) {
     const storageAutoCleanupRaw = typeof storageMaintenanceRaw?.autoCleanup === "object" && storageMaintenanceRaw.autoCleanup !== null
         ? storageMaintenanceRaw.autoCleanup
         : null;
+    const lockingRaw = typeof cfg.locking === "object" && cfg.locking !== null
+        ? cfg.locking
+        : null;
+    const redisLockRaw = typeof lockingRaw?.redis === "object" && lockingRaw.redis !== null
+        ? lockingRaw.redis
+        : null;
+    const redisLockUrl = resolveOptionalEnvString(redisLockRaw?.url) ??
+        resolveOptionalEnvString(cfg.redisUrl) ??
+        asNonEmptyString(process.env.MEMORY_LANCEDB_REDIS_URL);
+    const redisLockExplicitlyDisabled = redisLockRaw?.enabled === false;
+    const redisLockEnabled = !redisLockExplicitlyDisabled &&
+        (redisLockRaw?.enabled === true || Boolean(redisLockUrl));
     const userMdExclusiveRaw = typeof workspaceBoundaryRaw?.userMdExclusive === "object" && workspaceBoundaryRaw.userMdExclusive !== null
         ? workspaceBoundaryRaw.userMdExclusive
         : null;
@@ -4280,6 +4298,20 @@ export function parsePluginConfig(value) {
                     intervalHours: parsePositiveInt(storageAutoCleanupRaw.intervalHours) ?? 24,
                     retentionDays: parsePositiveInt(storageAutoCleanupRaw.retentionDays) ?? 7,
                     initialDelayMs: parseNonNegativeInt(storageAutoCleanupRaw.initialDelayMs) ?? 300_000,
+                },
+            }
+            : undefined,
+        redisUrl: resolveOptionalEnvString(cfg.redisUrl),
+        locking: redisLockRaw || redisLockUrl
+            ? {
+                redis: {
+                    enabled: redisLockEnabled,
+                    url: redisLockUrl,
+                    keyPrefix: asNonEmptyString(redisLockRaw?.keyPrefix),
+                    ttlMs: parsePositiveInt(redisLockRaw?.ttlMs) ?? 60_000,
+                    acquireTimeoutMs: parsePositiveInt(redisLockRaw?.acquireTimeoutMs) ?? 5_000,
+                    retryDelayMs: parsePositiveInt(redisLockRaw?.retryDelayMs) ?? 50,
+                    connectTimeoutMs: parsePositiveInt(redisLockRaw?.connectTimeoutMs) ?? 1_000,
                 },
             }
             : undefined,
--- a/dist/index.js
+++ b/dist/index.js
@@ -4155,6 +4155,12 @@ const memoryLanceDBProPlugin = {
                     clearInterval(storageMaintenanceTimer);
                     storageMaintenanceTimer = null;
                 }
+                try {
+                    await store.destroy();
+                }
+                catch (err) {
+                    api.logger.warn(`memory-lancedb-pro: stop cleanup failed: ${String(err)}`);
+                }
                 api.logger.info("memory-lancedb-pro: stopped");
             },
         });
--- a/dist/index.js
+++ b/dist/index.js
@@ -4161,6 +4161,13 @@ const memoryLanceDBProPlugin = {
                 catch (err) {
                     api.logger.warn(`memory-lancedb-pro: stop cleanup failed: ${String(err)}`);
                 }
+                finally {
+                    if (_singletonState?.store === store) {
+                        _singletonState = null;
+                    }
+                    _registeredApis.delete(api);
+                    _registeredApisMap.delete(api);
+                }
                 api.logger.info("memory-lancedb-pro: stopped");
             },
         });
@@ -4238,10 +4245,15 @@ export function parsePluginConfig(value) {
     const redisLockRaw = typeof lockingRaw?.redis === "object" && lockingRaw.redis !== null
         ? lockingRaw.redis
         : null;
-    const redisLockUrl = resolveOptionalEnvString(redisLockRaw?.url) ??
-        resolveOptionalEnvString(cfg.redisUrl) ??
-        asNonEmptyString(process.env.MEMORY_LANCEDB_REDIS_URL);
     const redisLockExplicitlyDisabled = redisLockRaw?.enabled === false;
+    const legacyRedisUrl = redisLockExplicitlyDisabled
+        ? undefined
+        : resolveOptionalEnvString(cfg.redisUrl);
+    const redisLockUrl = redisLockExplicitlyDisabled
+        ? undefined
+        : (resolveOptionalEnvString(redisLockRaw?.url) ??
+            legacyRedisUrl ??
+            asNonEmptyString(process.env.MEMORY_LANCEDB_REDIS_URL));
     const redisLockEnabled = !redisLockExplicitlyDisabled &&
         (redisLockRaw?.enabled === true || Boolean(redisLockUrl));
     const userMdExclusiveRaw = typeof workspaceBoundaryRaw?.userMdExclusive === "object" && workspaceBoundaryRaw.userMdExclusive !== null
@@ -4307,7 +4319,7 @@ export function parsePluginConfig(value) {
                 },
             }
             : undefined,
-        redisUrl: resolveOptionalEnvString(cfg.redisUrl),
+        redisUrl: legacyRedisUrl,
         locking: redisLockRaw || redisLockUrl
             ? {
                 redis: {
--- a/dist/index.js
+++ b/dist/index.js
@@ -1907,6 +1907,7 @@ const memoryLanceDBProPlugin = {
         // ========================================================================
         _registeredApis.add(api); // claim before init (Phase 2 singleton guard)
         _registeredApisMap.set(api, true); // dual-track: explicit claim for rollback
+        let registrationStopped = false;
         let singleton;
         try {
             if (!_singletonState) {
@@ -4143,6 +4144,12 @@ const memoryLanceDBProPlugin = {
                 }
             },
             stop: async () => {
+                if (registrationStopped) {
+                    return;
+                }
+                registrationStopped = true;
+                _registeredApis.delete(api);
+                _registeredApisMap.delete(api);
                 if (backupTimer) {
                     clearInterval(backupTimer);
                     backupTimer = null;
@@ -4155,18 +4162,18 @@ const memoryLanceDBProPlugin = {
                     clearInterval(storageMaintenanceTimer);
                     storageMaintenanceTimer = null;
                 }
-                try {
-                    await store.destroy();
-                }
-                catch (err) {
-                    api.logger.warn(`memory-lancedb-pro: stop cleanup failed: ${String(err)}`);
-                }
-                finally {
-                    if (_singletonState?.store === store) {
-                        _singletonState = null;
+                if (_registeredApisMap.size === 0 && _singletonState?.store === store) {
+                    try {
+                        await store.destroy();
                     }
-                    _registeredApis.delete(api);
-                    _registeredApisMap.delete(api);
+                    catch (err) {
+                        api.logger.warn(`memory-lancedb-pro: stop cleanup failed: ${String(err)}`);
+                    }
+                    finally {
+                        if (_singletonState?.store === store) {
+                            _singletonState = null;
+                        }
+                    }
                 }
                 api.logger.info("memory-lancedb-pro: stopped");
             },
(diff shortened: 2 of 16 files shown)


rwmjhb (changes requested):
Requesting changes. The direction is valuable, but the current fallback behavior breaks the distributed-lock guarantee under contention.

## Must fix

`RedisLockManager.acquire()` throws `RedisLockUnavailableError` both when Redis commands are unavailable and when `SET NX` cannot acquire the lock before `acquireTimeoutMs`. `MemoryStore.runWithWriteLock()` catches any `RedisLockUnavailableError` and falls back to `runWithFileLock()`.

That means if process A still owns the Redis lock, process B can time out waiting and then proceed under only the local file lock. Process A never took that file lock, so both writers can run concurrently in different lock domains. This fails exactly in the high-contention/distributed scenario the Redis lock is meant to fix.

Please separate these cases:

- Redis genuinely unavailable before entering the Redis lock protocol: file-lock fallback can be acceptable.
- Redis reachable but lock is held / acquire timeout: do not switch lock domains. Keep retrying, fail the write so the caller can retry, or otherwise preserve Redis-domain mutual exclusion.

## Also worth fixing

- The Redis lease uses `SET PX ttlMs NX`, but callbacks run without renewal or an operation timeout. A write longer than `ttlMs` can overlap with another writer after the lease expires.
- A transient Redis command error currently aborts acquisition rather than retrying until the deadline, amplifying the fallback problem.
- Fallback to file locking is logged only once per `MemoryStore`, so sustained Redis lock degradation can become invisible.
- Static analysis flags one new `any` cast in the dynamic ioredis import path.

Verification note: targeted tests and full npm test passed, but the current tests do not cover Redis-lock-held timeout falling back into the file-lock domain or lease expiry during an active write.

rwmjhb (changes requested):
Requesting changes. The new version fixes the previous Redis-lock-domain fallback direction, but there is still one blocking correctness issue in the lease-loss path.

## Must fix

`RedisLockManager.withLock()` starts the write callback, suppresses its rejection, and then returns:

```ts
Promise.race([operation, leaseLostPromise])
```

If lease renewal fails or lock ownership is lost, `leaseLostPromise` can reject first and the caller receives `RedisLockLeaseLostError`. But the original LanceDB write callback is not cancelled, awaited to completion before reporting, or given an `AbortSignal`. That means a caller can retry while the original write is still running and may later commit outside the caller's control.

For a distributed write lock, this is not safe: lease loss must not make the caller believe the write aborted unless the write operation can actually be cancelled.

Please change the lease-loss behavior so one of these is true:

- the active write can be cancelled/aborted when the lease is lost, or
- `withLock()` waits for the write to settle and returns a distinct non-retryable post-write lease-integrity error, or
- lease loss is handled in a way that cannot cause retry overlap with the original in-flight write.

## Also worth fixing

- `runStorageMaintenance()` still uses `runWithFileLock()` while Redis-enabled writes use `runWithWriteLock()`. `table.optimize()` is storage-mutating, so Redis deployments can split lock domains between maintenance and normal writes.
- Plugin stop/reload does not close Redis clients. `MemoryStore.destroy()` closes lock resources, but the service stop handler only clears timers.
- `openclaw.plugin.json` still describes `acquireTimeoutMs` as falling back to the file lock, but lock-held/acquire-timeout now fails/retries instead.
- A single transient renewal error currently marks the lease lost immediately.
- Consider whether `ioredis` should be a hard dependency for an optional feature.
- The `process.report.excludeNetwork` and legacy ID broadening changes look unrelated to the Redis locking scope and should be split out or justified separately.

Verification note: targeted tests and full npm test passed, but the current tests do not cover lease loss while a write continues, storage maintenance under Redis locking, or Redis client cleanup on plugin stop.

TurboTheTurtle:
Pushed d292e6b to address the lease-loss path: `withLock()` now waits for the active write to settle and reports a non-retryable lease-integrity error if ownership was lost before completion. I also moved storage maintenance onto the same write-lock path, close Redis lock resources through `destroy()`, and updated the acquire-timeout wording.

Ran:
- `node --test test/redis-lock.test.mjs test/storage-maintenance.test.mjs`

rickthomasjr:
## Agent Code Review — #879 (Redis write lock)

**Scope:** Redis write lock feature (~207 lines src, ~478 lines tests, config plumbing)
**CI:** All 6 checks pass

---

### P1 — `close()` race with concurrent `withLock()`

`close()` sets `this.clientPromise = null` synchronously. If a `withLock()` call has just awaited `getClient()` and received a client ref, `close()` could be called concurrently (from `stop()` in the plugin), nulling `clientPromise`. The client ref is already resolved, so this is fine for that call — but after `close()` returns, any concurrent `withLock()` will call `getClient()`, see `clientPromise === null`, call `createClient()` to create a new client, and proceed with the new client while the old one was just closed.

**Recommendation:** Add an `isClosed` flag (boolean, set before nulling `clientPromise` in `close()`) and check it in `getClient()` — throw `RedisLockUnavailableError` instead of creating a new client after shutdown.

### P1 — `withLock()` renew interval too aggressive at low TTL

With the default 60s TTL, renew fires every 20s. But when the user sets `ttlMs: 1_000` (1s), `renewIntervalMs = Math.floor(1/3) = 0` — and `Math.max(1, 0) = 1`, so it fires every 1ms. This means renewal runs ~1000x/sec on a 1s lease, which is wasteful and could cause Redis load issues under contention.

**Recommendation:** Use `Math.max(100, Math.floor(ttlMs / 5))` instead of `Math.max(1, Math.floor(this.config.ttlMs / 3))`.

### P2 — No `onWarning` emitted when Redis connection fails

In `createClient()`, the catch block wraps the error in `RedisLockUnavailableError`. The calling code in `runWithWriteLock` catches this and falls back to file lock, but no warning is emitted about the Redis connection failure. The admin won't know Redis locking is unavailable.

**Recommendation:** Call `onWarning` in the catch block of `createClient()` before wrapping the error.

### P2 — Retry delay minimum of 1s is surprising

`Math.max(this.retryDelayMs, 1_000)` in `acquire()` means a user-specified `retryDelayMs: 50` becomes 1000ms. With `acquireTimeoutMs: 5000`, the first retry takes at least 1000ms, so maximum 4-5 attempts. This is fine for most cases but the effective minimum is undocumented.

**Recommendation:** Reduce the floor to 100ms (`Math.max(this.retryDelayMs, 100)`) or document the 1s minimum in the config interface.

### P3 — Empty resource falls back to `"default"` silently

`resource || "default"` in `makeKey()`. If resource is empty string, it hashes `"default"`. Two different empty-string resources would collide on the same lock key. In practice `dbPath` is always non-empty, so this is low risk but the fallback is silent.

**Recommendation:** Use `resource ?? "default"` or throw if resource is falsy.

---

### Summary

| Severity | Count | Details |
|----------|-------|---------|
| P1 | 2 | Shutdown race, renew interval at low TTL |
| P2 | 2 | No warning on Redis connection fail, retry delay minimum surprising |
| P3 | 1 | Empty resource fallback |

The implementation is solid overall. Lua scripts are atomic, token-based release prevents stolen-lock issues, and auto-renewal with lease-lost detection is well-designed.

rwmjhb (commented):

GitHub currently reports this branch as `mergeable=CONFLICTING` / `merge_state_status=DIRTY`, so the orchestrator stopped at the conflict gate after R0/R1. The PR is still valuable, but reviewing the implementation now would be against a branch that has to be rewritten for conflict resolution.

rwmjhb (approved):
Approved. The current head resolves the earlier blocking Redis lock issues: lease-loss now waits for the active write before reporting integrity loss, storage maintenance uses the Redis write-lock domain, plugin stop closes lock resources, and the manifest/tests were updated.

Verification checked:

- `npm run build --if-present` passed on head `d35688898b123cbffa03124c59e89b832f5286bb`
- `node --test test/redis-lock.test.mjs` passed, 17/17
- `node --test test/storage-maintenance.test.mjs` passed, 6/6
- `node test/plugin-manifest-regression.mjs` passed

- Service stop now destroys the cached singleton store, but `_singletonState` is not cleared there. A same-process stop followed by re-register can reuse a destroyed store unless the host also calls `resetRegistration()`.
- `RedisLockManager.close()` can still race an active Redis-locked write during shutdown; it closes the shared client without tracking active holders.
- `locking.redis.enabled=false` with `locking.redis.url="${REDIS_URL}"` can still fail config parsing if `REDIS_URL` is unset, because the URL is resolved before the disabled check.
- `ioredis` is now a hard runtime dependency for an optional/default-off feature.
- The `process.report.excludeNetwork` workaround is host-global and unrelated to Redis locking; it would be cleaner split out or justified separately.
- `redisUrl` still implicitly enables Redis locking without explicit `locking.redis.enabled=true`.

rwmjhb (changes requested):
Requesting changes. The Redis lock direction is valuable, and the normal-path tests are green, but this head still has two correctness issues that can break the distributed-lock guarantee or active registrations.

## Must fix

1. Stopping one API registration tears down the shared singleton used by the others.

`register()` reuses the process-wide `_singletonState`/`MemoryStore` across API registrations, but each registration's service stop handler unconditionally calls `store.destroy()` and clears `_singletonState` for that one API. In a process with multiple live registrations, stopping or reloading one registration can close the shared store and Redis lock resources while the other registrations still have hooks/tools closing over that store.

Please remove the stopping API from the registration map first and only destroy/clear the singleton when no registered APIs remain, or introduce explicit reference-counted ownership for the shared store. Add a regression test with multiple `register()` calls where stopping one registration does not invalidate the other.

2. Redis unavailability can silently split the lock domain.

`MemoryStore.runWithWriteLock()` catches `RedisLockUnavailableError` and falls back to `runWithFileLock()`. That means one process with a failed Redis client/connect path can write under only the local file lock while another process with a working Redis client is still coordinating through Redis for the same `dbPath`. This recreates the split-lock-domain problem the PR is intended to solve. The shutdown path has a related variant because `RedisLockManager.close()` sets `isClosed` before active locks drain, causing racing writes to throw the same unavailable error and fall back to file locking.

Please distinguish "Redis was never configured/genuinely unavailable before using Redis mode" from "Redis lock mode is configured but this process cannot currently acquire/use it". In Redis-lock mode, acquisition/client/shutdown failures should fail closed rather than falling back into a different lock domain while peers may still use Redis. Add tests for a Redis client failure and shutdown race.

## Also worth addressing

- Lease-integrity loss is skipped when the protected write callback rejects, so an ambiguous write can surface as an ordinary storage error.
- `redisUrl` is resolved before higher-priority `locking.redis.url`, so a stale legacy placeholder can abort config parsing even when the nested URL is valid.
- `loadLanceDB` now mutates global `process.report.excludeNetwork`; that host-wide side effect is unrelated to Redis locking and should be split out or scoped/restored.

TurboTheTurtle:
Pushed eeadda7 to keep the shared singleton alive while other plugin registrations are still active, and still clear it once the final registration stops so a later registration gets a fresh store.

Ran:
- node --test test/redis-lock.test.mjs test/storage-maintenance.test.mjs
- npm run build --if-present

rwmjhb (commented):
This head is currently blocked by merge conflicts (`mergeable=CONFLICTING`, `merge_state_status=DIRTY`).

I can see the new commit is aimed at keeping the shared store alive for active registrations, but the branch needs to be rebased onto the latest base branch first. Please rebase/merge the base branch, resolve the conflicts, and push the updated branch. I'll re-run the full review once it is mergeable again, since conflict resolution may change the Redis lock and store lifecycle code paths.

TurboTheTurtle:
Pushed the Redis-lock follow-ups: close now keeps the manager shut down after shutdown, lease renewal is floored at 100ms, Redis connection failures emit the configured warning before falling back, retry delay uses a 100ms floor, and empty resources are rejected instead of sharing a default key.\n\nRan:\n- node --test test/redis-lock.test.mjs test/storage-maintenance.test.mjs\n- node test/plugin-manifest-regression.mjs\n- npm run build --if-present

rwmjhb (changes requested):
Requesting changes for generated output sync.

The Redis-lock review verdict is otherwise no longer blocking, and the focused checks pass on this head, but the checked-in dist output is stale. I verified locally on `c865d70`:
- `node --test test/redis-lock.test.mjs` passes
- `node --test test/storage-maintenance.test.mjs` passes

After running the build, the worktree is dirty:

```diff
M dist/src/retriever.js
```

The generated change is the compiled comment for the FTS routing initialization patch in `dist/src/retriever.js`. Please run the build and commit the updated dist output so the source/dist artifacts are reproducible from this branch.

Non-blocking follow-ups from review: Redis-unavailable fallback still deliberately degrades to file locking, automatic index folding currently uses the file-lock domain, and legacy `redisUrl` precedence can still surprise migrations. Those are documented tradeoffs/edge cases rather than blockers for this head.

rwmjhb (changes requested):
Requesting changes on the latest head. The generated-output sync issue is fixed, but the Redis lock implementation still has distributed-lock correctness gaps.

## Must fix

1. Automatic index folding still bypasses the Redis write-lock domain.

`noteDataModification()` can trigger `foldIndices()`, and `foldIndices()` runs `table.optimize()` under `runWithFileLock()` rather than `runWithWriteLock()`. In Redis-enabled deployments, normal writes coordinate through Redis while this storage-mutating optimize path only takes the local file lock, so it can overlap with Redis-locked writes and recreate the split lock-domain problem. Please route index folding through the same Redis write lock, or explicitly defer/disable it when Redis locking is active, with a regression test.

2. Redis connection/unavailable failures still fail open into a separate lock domain.

`runWithWriteLock()` catches `RedisLockUnavailableError` and falls back to `runWithFileLock()`. `createClient()` also maps Redis connection failures into that unavailable error. In a multi-host deployment, one host with a failed Redis connection can then write under only its local file lock while another host is still using Redis for the same dbPath. Those writers are no longer mutually exclusive. Redis-configured mode should fail closed by default after Redis locking is selected, or require an explicit best-effort fallback option with clear docs.

## Also worth addressing

- Lease loss is only checked after a successful callback; if the protected LanceDB write rejects after a partial write while the lease was lost, callers can receive an ordinary storage error instead of a lease-integrity error.
- `withLock()` tracks active locks only after `acquire()` resolves, so `close()` can return while an acquisition/write is still in flight.
- Legacy `redisUrl` is resolved before higher-priority `locking.redis.url`; a stale legacy env placeholder can break startup even when the nested URL is valid.

Verification note: R0 targeted checks and full npm test passed on this head, but those tests do not cover the lock-domain split paths above.

TurboTheTurtle:
Pushed updates for the Redis lock-domain issues: Redis-configured writes now fail closed instead of falling back to a local file lock, automatic index folding runs under the Redis write lock, and the Redis URL precedence/lifecycle follow-ups are covered.

Ran `node --test test/redis-lock.test.mjs`, `node --test test/storage-maintenance.test.mjs`, `node test/plugin-manifest-regression.mjs`, and `npm run build --if-present`.

rwmjhb (changes requested):
Requesting changes on the latest head. The normal write and storage-maintenance paths are much improved, and independent verification is clean, but there is still a lock-domain gap inside the Redis-lock feature scope.

Independent verification on `a8b8499` passed:
- `node --test test/redis-lock.test.mjs`
- `node --test test/storage-maintenance.test.mjs`
- generated/source files remained clean after build

## Must fix

Startup/admin schema and FTS mutations still bypass the Redis write lock. At this head, normal writes, storage maintenance, and automatic index folding use `runWithWriteLock()`, but initialization and admin reindex still mutate LanceDB directly:

- `doInitialize()` can call `table.addColumns()` and create/delete tables outside the Redis lock.
- startup FTS initialization calls `createFtsIndex(table)`, which directly calls `table.createIndex()`.
- `rebuildFtsIndex()` directly calls `dropIndex()` and `createFtsIndex()` from the `reindex-fts` CLI path.

Redis-enabled deployments should not have normal writes coordinated through Redis while startup/admin index/schema maintenance mutates the same store outside that lock domain. Please route these mutations through the same Redis write lock when Redis locking is enabled, or fail/defer them when the Redis lock is unavailable. Add regression coverage for at least `reindex-fts`/FTS rebuild using the Redis lock domain.

## Also worth fixing

- `RedisLockManager` ignores the final `RELEASE_SCRIPT` result. If the token is lost or reacquired after the last renewal check but before release, the callback can return success instead of a lease-integrity error.
- Environment-variable auto-enable can create mixed lock domains across hosts if some have `MEMORY_LANCEDB_REDIS_URL` and others do not. That may be acceptable as documented behavior, but it should be explicit because this feature is about cross-host mutual exclusion.
- `ioredis` remains a hard dependency for a default-off feature.

rwmjhb (approved):
Approved. The latest head addresses the Redis lock-domain blockers I raised: FTS rebuild now runs inside the Redis write-lock domain, schema/maintenance paths have targeted lock coverage, and final release ownership loss is tested.

Independent verification on `5885e87` passed:
- `node --test test/redis-lock.test.mjs`
- `node --test test/storage-maintenance.test.mjs`
- build left source/dist/package/manifest/docs files clean

Non-blocking notes remain around static hygiene: the PR still adds a couple of `any` casts around index metadata and several direct maintenance/startup console logs. Those are follow-up quality issues, not merge blockers for the Redis locking fix.

PR #880, opened 2026-06-11
feat: add LanceDB storage maintenance

### LINKED ISSUE #792: feat: add LanceDB version snapshot auto-cleanup
LanceDB version snapshot storage grows to 300MB after a few days with only ~1000 entries. Plugin never calls table.optimize() to clean up old snapshots. Add configurable auto-cleanup interval.

TurboTheTurtle:
## Summary
- add opt-in scheduled LanceDB table maintenance for old version snapshots
- serialize cleanup with existing store write locking and log success/failure details
- expose storage maintenance config in the plugin manifest with regression coverage

Fixes #792

## Verification
- node --test test/storage-maintenance.test.mjs
- npm run build
- npm run test:storage-and-schema

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -3869,6 +3869,10 @@ const memoryLanceDBProPlugin = {
         // ========================================================================
         let backupTimer = null;
         const BACKUP_INTERVAL_MS = 24 * 60 * 60 * 1000; // 24 hours
+        let storageMaintenanceInitialTimer = null;
+        let storageMaintenanceTimer = null;
+        let storageMaintenanceRunning = false;
+        const storageAutoCleanup = config.storageMaintenance?.autoCleanup;
         async function runBackup() {
             try {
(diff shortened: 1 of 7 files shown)


rwmjhb (approved):
Approved.

The opt-in maintenance path is focused, serialized under the store write lock, and the storage-maintenance coverage exercises optimize, retention clamping, config parsing, and manifest/schema defaults.

Verification passed:

```bash
# orchestrator R0
full npm test: PASS

npm run build --if-present
node --test test/storage-maintenance.test.mjs
```

Nice-to-have follow-ups: add upper bounds for `storageMaintenance.autoCleanup.intervalHours` and `initialDelayMs` so oversized values cannot overflow Node's timer delay range; also consider validating/logging the effective clamped `retentionDays` at config parse time to avoid configured-vs-effective divergence.

PR #885, opened 2026-06-12
fix: await store init before retrieval routing reads hasFtsSupport

### LINKED ISSUE #884: retrieve() checks store.hasFtsSupport before lazy init completes, so the first search in a fresh process silently routes vector-only (CLI search never uses BM25)
**Version:** 1.1.0-beta.11 dist, OpenClaw 2026.6.1

## Summary

`MemoryRetriever.retrieve()` routes between hybrid and vector-only retrieval via:

```js
else if (this.config.mode === "vector" || !this.store.hasFtsSupport) {
    results = await this.vectorOnlyRetrieval(...)
}
```

`hasFtsSupport` is a getter over `ftsIndexCreated`, which is only assigned during the store's lazy `ensureInitialized()`. The routing check runs before any store operation has triggered initialization, so on the first `retrieve()` of any fresh process the getter returns the constructor default `false` and the search silently routes vector-only. No error, no warning. `bm25ResultCount` stays 0.

## Impact

1. Every `openclaw memory-pro search` invocation is a fresh process, so the CLI never exercises hybrid/BM25 retrieval, while `stats` reports "FTS support: Yes" (stats runs store operations before reading the getter). This is misleading when debugging retrieval quality, and the integration playbook's troubleshooting step ("verify BM25 / lexical fallback behavior first") cannot actually be performed through the CLI.
2. In a long-lived gateway the first retrieval after plugin load can also route vector-only if it wins the race against other store operations.

## Two-step reproduction

```js
import { MemoryStore } from "./dist/src/store.js";
const store = new MemoryStore({ dbPath, vectorDim: 1024 });
console.log(store.hasFtsSupport);        // false
await store.bm25Search("warmup", 1);     // triggers ensureInitialized
console.log(store.hasFtsSupport);        // true
```

Calling `store.bm25Search` directly with identical args (query, limit, scopeFilter, excludeInactive) returns the expected hits with healthy scores, confirming the search itself works. Only the routing check misfires.

## Suggested fix

`await this.store.ensureInitialized()` at the top of `retrieve()`, or resolve the routing decision after initialization. A one-line await makes routing deterministic.

## Related observations from the same debugging session

- Our FTS index reported `numIndexedRows: 0, numUnindexedRows: 205` (created over an empty table; the codebase has no optimize/merge step that ever folds new rows into the index). Queries still hit via the unindexed tail, but documenting the index lifecycle would help operators.
- `reindex-fts` exists in cli.ts source but is absent from the compiled dist build shipped at beta.11, so the rebuild path is not reachable from the installed CLI.

gorkem2020:
Fixes [#884](https://github.com/CortexReach/memory-lancedb-pro/issues/884).

## Problem

`MemoryRetriever.retrieve()` decides between hybrid and vector-only retrieval by reading `this.store.hasFtsSupport` before any store operation has triggered the store's lazy `ensureInitialized()`. On the first `retrieve()` of a fresh process the getter still returns the constructor default `false`, so the search silently routes vector-only. Every `openclaw memory-pro search` invocation is a fresh process, which means the CLI never exercised BM25 or the cross-encoder rerank at all (both live behind the hybrid route).

## Fix

One awaited call at the top of `retrieve()` so initialization resolves before the routing check. Optional chaining keeps it safe for store implementations without the method.

The same change is mirrored into `dist/src/retriever.js` to match this repo's convention of committing compiled output. Happy to drop that hunk if you prefer to regenerate dist through your own build.

## Verification

A/B on an identical store and query (a rare keyword that appears verbatim in exactly one stored memory; scoped search; fresh process for each run):

| | bm25ResultCount | result tags |
| --- | --- | --- |
| before | 0 | `vector` only |
| after | 1 | `vector+BM25+reranked` |

Also verified in a live OpenClaw deployment: first-call CLI searches now report non-zero BM25 counts and reranked tags, with no behavior change for already-initialized stores (the await is a no-op once init has completed).

### DIFF
--- a/dist/src/retriever.js
+++ b/dist/src/retriever.js
@@ -340,6 +340,8 @@ export class MemoryRetriever {
         const { query, limit, scopeFilter, category, source, signal } = context;
         const safeLimit = clampInt(limit, 1, 20);
         this.lastDiagnostics = null;
+        // FLEET-PATCH(#884): resolve lazy store init BEFORE routing reads hasFtsSupport.
+        await this.store.ensureInitialized?.();
         const diagnostics = {
             source,
             mode: this.config.mode,
--- a/src/retriever.ts
+++ b/src/retriever.ts
@@ -588,6 +588,10 @@ export class MemoryRetriever {
     const { query, limit, scopeFilter, category, source, signal } = context;
     const safeLimit = clampInt(limit, 1, 20);
     this.lastDiagnostics = null;
+
+    // FLEET-PATCH(#884): resolve lazy store init BEFORE routing reads hasFtsSupport,
+    // otherwise the first retrieve() in a fresh process silently routes vector-only.
+    await (this.store as { ensureInitialized?: () => Promise<void> }).ensureInitialized?.();
     const diagnostics: RetrievalDiagnostics = {
       source,
       mode: this.config.mode,
--- a/src/retriever.ts
+++ b/src/retriever.ts
@@ -591,7 +591,7 @@ export class MemoryRetriever {
 
     // FLEET-PATCH(#884): resolve lazy store init BEFORE routing reads hasFtsSupport,
     // otherwise the first retrieve() in a fresh process silently routes vector-only.
-    await (this.store as { ensureInitialized?: () => Promise<void> }).ensureInitialized?.();
+    await (this.store as unknown as { ensureInitialized?: () => Promise<void> }).ensureInitialized?.();
     const diagnostics: RetrievalDiagnostics = {
       source,
       mode: this.config.mode,


rickthomasjr:
## Substantive Code Review — PR #885

**Reviewer:** Harrison Forge (agent) | **Date:** 2026-06-12

---

### TL;DR

**APPROVE.** This is a correct, minimal fix for a real bug. The A/B verification in the PR body confirms the fix addresses the issue.

---

### Change-by-change analysis

#### `src/retriever.ts` — `await this.store.ensureInitialized?.()`

**The bug:** `retrieve()` reads `this.store.hasFtsSupport` before any operation has triggered `ensureInitialized()`. In a fresh process, `hasFtsSupport` is still the constructor default `false`, so the retrieval silently routes vector-only. This means:

- Every `openclaw memory-pro search` invocation (fresh process per CLI command) never uses BM25
- The cross-encoder rerank (behind hybrid mode) is also unreachable from CLI

**The fix:** One `await this.store.ensureInitialized?.()` at the top of `retrieve()`. Safe because:

- Optional chaining (`?.`) handles store implementations that don't expose the method
- Once initialized, `ensureInitialized()` returns immediately (no-op)
- The `await` is correct — `hasFtsSupport` is set *after* the table is opened, so the init must complete before the routing decision

**Why this is correct:** The lazy init pattern is sound — you don't want to block on the FTS index if the table doesn't exist yet. But any code path that *reads* `hasFtsSupport` must happen *after* init completes, which this fix guarantees.

#### `dist/src/retriever.js` — mirrored change

Correctly mirrors the source. Two-line addition matching the TS change.

#### `scripts/ci-test-manifest.mjs` — EOL noise

The diff shows every line changed because of CRLF-LF conversion. This is cosmetic noise in the diff but doesn't affect the actual content. The manifest line that matters is just the addition of `test/store-importance-normalization.test.mjs` — the rest is whitespace normalization. Not a blocker.

---

### Assessment

| Dimension | Value |
|-----------|-------|
| Bug severity | **HIGH** — CLI searches silently degraded to vector-only |
| Fix size | TINY — 4 lines added |
| Risk | **LOW** — idempotent await, safe for already-initialized stores |
| Test coverage | **ADEQUATE** — A/B verification in PR body confirms the fix; no new unit test but the behavioral change is straightforward |
| Correctness | **SOLID** — no edge cases identified |

### Minor notes (non-blocking)

1. The cast `(this.store as unknown as { ensureInitialized?: () => Promise<void> }).ensureInitialized?.()` is correct for the optional pattern. Could also use `if ('ensureInitialized' in this.store)` for type safety, but the current approach is fine.
2. Consider adding a unit test to verify that `retrieve()` always triggers init before the routing check, to prevent regression if the lazy init order is ever refactored.

### Verdict

**APPROVE.** Ship it.

rickthomasjr (approved):
## Agent Code Review: PR #885 — fix: await store init before retrieval routing reads hasFtsSupport

**Verdict: APPROVE**

### Summary
Single-await fix at the top of `MemoryRetriever.retrieve()` ensuring `ensureInitialized()` resolves before `hasFtsSupport` is read. Fixes issue #884 where CLI searches (fresh process per command) always routed vector-only because the store's lazy init hadn't fired yet.

### P1 — Correctness: APPROVED
The root cause analysis is precise: `hasFtsSupport` getter returns constructor default `false` until `ensureInitialized()` resolves. On every fresh process (all CLI invocations), the first retrieve() sees stale `false` and routes vector-only. The fix — `await this.store.ensureInitialized?.()` at the top of retrieve() — ensures initialization completes before routing decisions.

### P2 — Safety: APPROVED
Optional chaining (`?.`) on both `ensureInitialized` property access and the await call. This is defensive: if a store mock or alternate implementation lacks the method, it's a no-op. No breaking change risk.

### P3 — Correctness of dual-hunk approach
Both `src/retriever.ts` and `dist/src/retriever.js` are updated. The source change uses a cast `(this.store as unknown as { ensureInitialized?: () => Promise<void> })` which avoids adding `ensureInitialized` to the store's public interface — reasonable given this is a one-shot fix. The dist match ensures the deployed code is in sync.

### P4 — Performance: APPROVED
`ensureInitialized()` is idempotent (the init promise is cached internally in MemoryStore). The await on subsequent calls is effectively zero-cost — the promise resolves immediately. No regression risk.

### P5 — Verification
A/B with identical store showing bm25ResultCount 0→1 is convincing. The live deployment note confirms no behavioral change for already-initialized stores.

### Minor Notes
- The cast to anonymous shape is pragmatic but slightly ugly. If the store is going to expose `ensureInitialized` as a public method (which it clearly will given this usage), consider adding it to the type definition to avoid the cast in the future.
- Consider adding a test case: first retrieve() in fresh process uses hybrid route.

### Conclusion
Clean, minimal, well-tested fix for a real correctness bug. No blocking issues. APPROVE.

rwmjhb (approved):
Approved. This is a small, high-value fix: `retrieve()` now initializes the store before reading `hasFtsSupport`, which addresses the fresh-process vector-only routing bug for the primary retrieval path.

Verification checked:

- `npm run build --if-present` passed on head `e911dae339d36a6631e922132ff57182e7857105`
- `node --test test/retrieval-trace.test.mjs` passed, 16/16
- `node test/functional-e2e.mjs` passed

- `retrieveWithTrace()` still reads `store.hasFtsSupport` before awaiting store initialization, so a first debug/trace retrieval can still route vector-only.
- The new initialization await happens before retrieve diagnostics are created/caught, so initialization failures can bypass `lastDiagnostics` population.
- The private `ensureInitialized()` access is hidden behind an `unknown` cast; a typed public initialization interface would make this harder to regress silently.

PR #887, opened 2026-06-12
fix: fold FTS index after write threshold and on startup backlog

### LINKED ISSUE #886: FTS index never folds new rows on a default install: index maintenance is coupled to the opt-in storageMaintenance.autoCleanup (numIndexedRows stays at creation-time count)
**Version:** master (16c33a0) and 1.1.0-beta.11 lineage, @lancedb/lancedb 0.26.2, OpenClaw 2026.6.1, Linux, Node 24

### Summary

LanceDB indices are point-in-time snapshots: rows written after index creation accumulate in an unindexed tail until `table.optimize()` folds them in. In this plugin the only automatic `optimize()` call is the storage maintenance scheduler in index.ts, and it is gated on `storageMaintenance.autoCleanup.enabled === true` (default `false`). On a default install the FTS index is created once at first init, frequently over an empty or near-empty table, and never folds another row for the lifetime of the store.

### Observed

Live store with 205 entries accumulated over weeks of auto-capture: `indexStats` on the text index reported `numIndexedRows: 0, numUnindexedRows: 205` (the index was created at first boot over an empty table and never updated). Search correctness is preserved because LanceDB brute-force scans the unindexed tail, but that scan grows linearly with store size, which quietly defeats the point of the index for BM25 and for any future vector index.

This surfaced while debugging [#884](https://github.com/CortexReach/memory-lancedb-pro/issues/884) and was mentioned there as a related observation; filing separately since it is an independent defect with an independent fix.

### Why the existing knob does not cover it

1. `storageMaintenance.autoCleanup` couples two different concerns: version-retention pruning (deletes table history, reasonable to keep opt-in) and index folding plus compaction (pure performance maintenance, safe, arguably should always be on).
2. The config schema describes the knob only as "pruning old version snapshots with table.optimize(). Disabled by default." An operator reading the schema has no signal that FTS index coverage depends on enabling it.
3. Even when enabled, folding is interval-based (default 24h plus a 5 minute post-boot run). The LanceDB SDK docstring for `optimize()` recommends running it after roughly 20 data modification operations, which an interval cannot track.

### Suggested direction

1. Decouple index folding from retention pruning: in `MemoryStore`, count data modification operations (the batched `table.add` in `doFlush`, plus the delete-and-re-add paths in `upsert`/`update`) and run `table.optimize({ cleanupOlderThan: new Date(0) })` after ~20 ops. The epoch cutoff folds indices and compacts fragments while pruning no version history, so retention stays fully under the opt-in `storageMaintenance.autoCleanup` path.
2. On initialization, if the FTS index reports an unindexed backlog above the same threshold, schedule a one-time background catch-up fold (repairs long-lived stores like the one above on the first boot after upgrade).
3. Mention index folding in the `autoCleanup` schema description either way.

A PR implementing 1 and 2 follows shortly.

gorkem2020:
Fixes [#886](https://github.com/CortexReach/memory-lancedb-pro/issues/886)

### What

- `MemoryStore` now counts data modification operations (the batched `table.add` in `doFlush`, plus the delete-and-re-add paths in `upsert`, `update`, and bulk update) and runs `table.optimize({ cleanupOlderThan: new Date(0) })` after 20 ops, following the LanceDB SDK guidance, so indices fold in newly written rows instead of leaving them in the brute-force-scanned unindexed tail. The epoch cutoff makes the prune step a no-op, so version retention remains exclusively under the existing opt-in `storageMaintenance.autoCleanup` path, which is unchanged.
- On initialization, if the FTS index reports an unindexed backlog of 20 or more rows, a one-time background catch-up fold is scheduled. This repairs long-lived stores on the first boot after upgrade (the store in the issue reported `numIndexedRows: 0, numUnindexedRows: 205`).
- Folds are guarded by an in-flight flag, run under the existing cross-process file lock, never block writes or initialization, and re-arm the op counter on failure so a transient error retries on later writes.

### Tests

- New `test/fts-index-fold.test.mjs` (registered in the CI manifest under storage-and-schema): asserts the write-threshold fold and the startup catch-up fold via `indexStats` on a live store. 2/2 pass.
- Neighboring suites pass: storage-maintenance 5/5, update-consistency-lancedb 6/6, cross-process-lock 7/7, store-empty-scope-filter 1/1.

### Notes

`dist/src/store.js` was rebuilt with the repo's `tsc` build. A fresh build also wants to modify `dist/src/retriever.js`, because `src/retriever.ts` already contains a rerank `top_n` cap that the committed dist predates; that file is intentionally left untouched here to keep this PR scoped to the issue. You may want to refresh that file separately.

### DIFF
--- a/dist/src/store.js
+++ b/dist/src/store.js
@@ -30,6 +30,21 @@ export function __setLockfileModuleForTests(module) {
 }
 export const loadLanceDB = async () => {
     if (!lancedbImportPromise) {
+        // @lancedb/lancedb's napi-rs loader (dist/native.js) detects musl via
+        // process.report.getReport(). The report's network section performs
+        // reverse-DNS lookups for every network interface, synchronously on the
+        // calling thread; on hosts with several interfaces and a slow or flaky
+        // DNS path (measured on WSL2 + Tailscale) this blocks the event loop for
+        // 110-250s on the FIRST LanceDB load — the host app's HTTP server and
+        // pollers freeze with CPU at 0% (main thread stuck in poll()). Excluding
+        // the network section turns the measured 235s hang into ~3ms and loses
+        // nothing: isMusl() only reads report.header.glibcVersionRuntime.
+        try {
+            process.report.excludeNetwork = true;
+        }
+        catch {
+            /* Node < 22 without the flag — keep the previous behavior */
+        }
         // Use a createRequire-built require() so LanceDB's CommonJS native bindings
         // keep Windows-safe CJS semantics while still working in pure ESM runtimes.
         // Do not name this binding "require": bundlers may rewrite bare require()
@@ -361,6 +376,8 @@ export class MemoryStore {
     _lastFtsError = null;
     updateQueue = Promise.resolve();
     nativeCosineFallbackLogged = false;
+    dataModsSinceIndexFold = 0;
+    indexFoldInFlight = false;
     // Cross-call batch accumulator（Issue #690）
     // 多個 concurrent bulkStore() 會先累積在這裡，每 100ms flush 一次，
     // 合併成一個 lock acquisition，大幅降低 lock contention。
@@ -381,6 +398,10 @@ export class MemoryStore {
     // 【MR2 fix】pendingBatch 上限，防止高生產率時無限增長。
     // 當 pending callers 超過此值時，block 並同步 flush，確保 pendingBatch 不會無限膨胀。
     static MAX_PENDING_BATCH_SIZE = 1000;
+    // LanceDB indices are point-in-time snapshots; the SDK recommends running
+    // optimize() after ~20 data modification operations so indices fold in
+    // newly written rows instead of leaving them in the brute-force-scanned tail.
+    static INDEX_FOLD_OP_THRESHOLD = 20;
     config;
     disableNativeCosine;
     constructor(config) {
@@ -538,6 +559,63 @@ export class MemoryStore {
             };
         });
     }
+    /**
+     * Fold newly written rows into existing indices without touching version
+     * history. LanceDB indices are point-in-time snapshots: rows added after
+     * index creation sit in an unindexed tail that searches must brute-force
+     * scan, so without periodic optimize() the FTS index never covers rows
+     * written after first init. The epoch cleanupOlderThan cutoff makes the
+     * prune step a no-op; version retention stays under the opt-in
+     * storageMaintenance.autoCleanup path (see runStorageMaintenance).
+     */
+    async foldIndices(reason) {
+        if (this.indexFoldInFlight || !this.table || typeof this.table.optimize !== "function") {
+            return;
+        }
+        this.indexFoldInFlight = true;
+        const mods = this.dataModsSinceIndexFold;
+        this.dataModsSinceIndexFold = 0;
+        try {
+            await this.runWithFileLock(async () => {
+                await this.table.optimize({ cleanupOlderThan: new Date(0) });
+            });
+            console.log(`[memory-lancedb-pro] index fold completed (reason=${reason}, modsSinceLast=${mods})`);
+        }
+        catch (err) {
+            // Re-arm the counter so a transient failure retries on later writes.
+            this.dataModsSinceIndexFold += mods;
+            console.warn(`[memory-lancedb-pro] index fold failed (reason=${reason}): ${err instanceof Error ? err.message : String(err)}`);
+        }
+        finally {
+            this.indexFoldInFlight = false;
+        }
+    }
+    noteDataModification() {
+        this.dataModsSinceIndexFold += 1;
+        if (this.dataModsSinceIndexFold >= MemoryStore.INDEX_FOLD_OP_THRESHOLD) {
+            void this.foldIndices("write-threshold");
+        }
+    }
+    async scheduleStartupIndexCatchUp() {
+        try {
+            const table = this.table;
+            if (!table || typeof table.indexStats !== "function")
+                return;
+            const indices = await table.listIndices();
+            const fts = indices.find((idx) => idx.indexType === "FTS" || idx.columns?.includes("text"));
+            if (!fts)
+                return;
+            const stats = await table.indexStats(fts.name ?? "text_idx");
+            const backlog = stats?.numUnindexedRows ?? 0;
+            if (backlog >= MemoryStore.INDEX_FOLD_OP_THRESHOLD) {
+                console.log(`[memory-lancedb-pro] FTS index has ${backlog} unindexed rows; scheduling catch-up fold`);
+                void this.foldIndices("startup-backlog");
+            }
+        }
+        catch {
+            // Index stats are best-effort; failures must never affect initialization.
+        }
+    }
     async ensureInitialized() {
         if (this.table) {
             return;
@@ -658,6 +736,9 @@ export class MemoryStore {
         }
         this.db = db;
         this.table = table;
+        // Fold any unindexed backlog accumulated while no maintenance ran
+        // (best-effort, runs in the background, never blocks initialization).
+        void this.scheduleStartupIndexCatchUp();
     }
     async backfillLegacySecondTimestamps(table) {
         try {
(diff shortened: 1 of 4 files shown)


rickthomasjr:
## Substantive Code Review — PR #887

**Reviewer:** Harrison Forge (agent) | **Date:** 2026-06-12

---

### TL;DR

**APPROVE.** This is a well-designed, correctly implemented fix for LanceDB FTS index staleness. The implementation follows LanceDB SDK best practices and the test suite is thorough.

---

### Change-by-change analysis

#### Core mechanism: `noteDataModification()` + `foldIndices()`

**`noteDataModification()`** increments a counter after each data-modifying operation (upsert, update, bulk add, bulk update). When the counter reaches 20 (`INDEX_FOLD_OP_THRESHOLD`), it triggers `foldIndices("write-threshold")`.

**`foldIndices()`** runs `table.optimize({ cleanupOlderThan: new Date(0) })` under the existing file lock:
- Guards against concurrent folds with `indexFoldInFlight` flag
- Resets the counter on success
- Re-arms the counter on failure (so a transient error retries on later writes)
- Uses `console.log`/`console.warn` for status (consistent with existing patterns in store.js)

**`scheduleStartupIndexCatchUp()`** runs on init:
- Queries `table.listIndices()` for an FTS index
- Calls `table.indexStats()` to get `numUnindexedRows`
- If backlog >= 20, triggers a background catch-up fold
- Never blocks initialization (uses `void` fire-and-forget)

**Correctness:** Excellent. The implementation correctly handles:
- **Concurrency:** File lock + in-flight flag prevents double-folds
- **Failure recovery:** Counter re-arming ensures transient failures don't cause permanent index staleness
- **Non-blocking:** Startup catch-up is fire-and-forget; folds don't block writes
- **Version retention:** `cleanupOlderThan: new Date(0)` is a no-op for pruning (all rows have timestamps after epoch), so version history is untouched

#### `ensureInitialized()` — startup catch-up

Called at the end of `ensureInitialized()`, after the table is opened. This ensures the catch-up runs against a real table (not undefined).

**Correctness:** Sound. The `void` prefix ensures init never waits for the catch-up.

#### `noteDataModification()` call sites

Added to:
1. `upsert()` — after the write
2. `doFlush()` — after each chunk's `table.add()`
3. `update()` — after the delete-and-re-add
4. `updateMany()` (bulk update) — after the updated entries are added

**Correctness:** Complete coverage of all mutation paths. Each call is after the write completes (not before), ensuring the counter reflects actual committed modifications.

#### `src/store.ts` — `isLegacyStableMemoryId` regex

Added: `LEGACY_STABLE_MEMORY_ID_REGEX = /^[A-Za-z0-9][A-Za-z0-9._:-]{2,127}$/`

This replaces the old `mem-md-\d+` regex with a broader pattern that accepts stable IDs from older stores (e.g., `data-pointer-...`).

**Correctness:** The regex is reasonable — requires alphanumeric start, allows common ID separators, enforces 3-128 char length. However, it's a subtle behavior change (more IDs accepted as "legacy exact match" instead of prefix match). Worth noting but not a blocker.

#### `test/fts-index-fold.test.mjs`

Two tests:
1. **Write-threshold fold:** Creates 20 entries, waits for the background fold, asserts `numUnindexedRows === 0`
2. **Startup catch-up:** Seeds one entry, then grows a backlog of 25 entries directly (bypassing the store's modification counter), opens a fresh store, waits for the catch-up fold

**Test quality:** Good. Both tests use `waitFor()` to handle the async nature of the background operations. The second test cleverly bypasses the counter to simulate a real-world upgrade scenario.

#### `dist/src/store.js` — correctly rebuilt

Matches the source changes.

#### CI manifest

New test registered under `storage-and-schema`. Clean.

---

### Assessment

| Dimension | Value |
|-----------|-------|
| Bug severity | **HIGH** — unindexed FTS tail degrades search performance over time; long-lived stores can accumulate hundreds of unindexed rows |
| Fix size | MEDIUM — ~100 lines of new code |
| Risk | **LOW** — folds are fire-and-forget, under existing file lock, non-blocking |
| Test coverage | **STRONG** — 2 targeted tests + verified neighbor suites pass |
| Correctness | **SOLID** — handles concurrency, failures, and version retention correctly |

### Minor notes (non-blocking)

1. **Threshold tuning:** `INDEX_FOLD_OP_THRESHOLD = 20` follows LanceDB SDK guidance but is hardcoded. Consider making it configurable via `storageMaintenance.autoCleanup.indexFoldThreshold` for power users who want to tune it based on their workload.
2. **Logging:** The `console.log`/`console.warn` calls in `foldIndices()` bypass the OpenClaw API logger. This is fine for now (consistent with existing patterns) but consider using `api.logger` if this function ever gets called from a context where the API is available.
3. **The `isLegacyStableMemoryId` change** is technically unrelated to the FTS fold scope. Consider splitting it into a separate commit or PR to keep the diff focused.

### Verdict

**APPROVE.** Well-designed, well-tested, correctly scoped.

rickthomasjr (approved):
## Agent Code Review: PR #887 — fix: fold FTS index after write threshold and on startup backlog

**Verdict: APPROVE**

### Summary
Adds FTS index maintenance to MemoryStore: (1) counts data modifications and runs `table.optimize()` after 20 ops per LanceDB SDK guidance, (2) on startup detects unindexed backlog and schedules a background catch-up fold, (3) folds are guarded by an in-flight flag, run under the existing cross-process file lock, and re-arm on failure. Also patches `process.report.excludeNetwork = true` to avoid 235s DNS hang on first LanceDB load.

### P1 — Index fold correctness: APPROVED
The `foldIndices` method is well-designed: in-flight guard prevents concurrent optimize calls, the file lock serializes with other write operations, the counter is reset atomically, and transient failures re-arm the counter for retry. The `cleanupOlderThan: new Date(0)` makes the prune step a no-op, preserving version retention under the existing opt-in maintenance path. Correct.

### P2 — Startup catch-up: APPROVED
`scheduleStartupIndexCatchUp` is fired-and-forget (`void`) from `ensureInitialized()` — never blocks initialization. It checks for FTS indices via `listIndices()` and queries `indexStats` for unindexed row count. Failure paths are all caught with no-op. Correct.

### P3 — Write path integration: APPROVED
`noteDataModification()` is called after: `upsert`, `store()` (batch), `update()`, and `delete()` paths. The counter increments and triggers `foldIndices` at threshold. No blocking — `foldIndices` is called with `void`. Correct.

### P4 — `process.report.excludeNetwork = true`: APPROVED (with note)
Fixes a measured 235s event loop block on hosts with slow DNS (WSL2 + Tailscale). The `isMusl()` check only reads `report.header.glibcVersionRuntime`, so excluding the network section loses nothing for this purpose. Wrapped in try/catch for Node < 22. Minor concern: this is a global process-level flag. If multiple OpenClaw plugins or the gateway itself calls `process.report.getReport()`, they won't see network data. The risk is low — `process.report` is rarely used at runtime — but worth documenting.

### P5 — `isLegacyStableMemoryId` regex broadening: APPROVED
The original `^mem-md-\d+$` was overly narrow. The new regex `/^[A-Za-z0-9][A-Za-z0-9._:-]{2,127}$/` accepts stable identifiers from data exports (e.g. `data-pointer-...`, `mem-md-...`). The 3-128 char length constraint is reasonable. No SQL injection risk since the ID is escaped via `escapeSqlLiteral` in the query path. Correct.

### P6 — CI manifest reordering: APPROVED
`test/fts-index-fold.test.mjs` added to the manifest under `storage-and-schema`. Neighboring test file order preserved. Correct.

### Minor Notes
- The `dist/src/store.js` rebuild may cause unnecessary noise on the next CI run if the `tsc` build produces different output (e.g., whitespace). The PR note about `retriever.js` being intentionally left untouched is good hygiene.
- The `isLegacyStableMemoryId` regex is duplicated in `dist/src/store.js` and `src/store.ts`. If the dist is auto-built, this should be consistent. If not, watch for divergence.
- Consider a log message when `noteDataModification()` triggers a fold (currently only logged on completion, not on trigger). Not blocking.

### Conclusion
Solid maintenance feature that addresses a real performance issue (unindexed FTS tail causing brute-force scans). The global `excludeNetwork` flag is the one item worth a note but is justified by the measured impact. APPROVE.

rwmjhb (commented):

GitHub currently reports this branch as `mergeable=CONFLICTING` / `merge_state_status=DIRTY`, so the orchestrator stopped at the conflict gate after R0/R1. The FTS index-folding problem is worth reviewing, but the implementation should be checked after the branch applies cleanly to the current base.

rwmjhb (approved):
Approved. The conflict is resolved and this head passes the focused storage maintenance checks. I verified independently:
- `node --test test/fts-index-fold.test.mjs`
- build left `src/store.ts` and `dist/src/store.js` clean

Non-blocking follow-ups: `importEntry` still bypasses the new fold counter, so high-volume imports may rely on restart catch-up or later counted writes; the `process.report.excludeNetwork` dist/source concern did not reproduce after build, but please keep unrelated runtime-loader changes out of future focused storage PRs when possible.

PR #893, opened 2026-06-13
fix: verify fresh dist output

### LINKED ISSUE #890: Committed dist is stale relative to src: shipped builds miss the process.report.excludeNetwork hang guard and the rerank top_n cap
**Version:** master (16c33a0), @lancedb/lancedb 0.26.2, Node 22/24, Linux

### Summary

Running the repo's own build (`npm run build`, `tsc -p tsconfig.json`) on a clean master checkout modifies two committed dist files. The committed compiled output is stale relative to `src/`, so installs that run the shipped `dist/` are missing fixes the source already contains:

1. `dist/src/store.js`: missing the `process.report.excludeNetwork = true` guard in `loadLanceDB()`. Per the comment in `src/store.ts`, without it the napi-rs musl detection can synchronously block the event loop for 110-250s on the first LanceDB load on hosts with several network interfaces and a slow DNS path (WSL2 + Tailscale measured). Shipped builds still have that hang exposure.
2. `dist/src/retriever.js`: missing the rerank `top_n` cap (`Math.min(results.length, Math.max(1, candidatePoolSize))`); the shipped build passes `results.length` straight through to the rerank request.

Fresh `tsc` output diff against committed dist: `dist/src/store.js | 32 +++++++++---` and `dist/src/retriever.js | 3 ++-` (plus whatever change one is building, which is how we noticed: every scoped PR build wants to drag these two files in).

### Suggested direction

1. Rebuild and commit `dist/` from current `src/` so the shipped files match the source.
2. Add a freshness check to the packaging workflow (the repo already has a packaging-and-workflow CI group and `scripts/verify-package-runtime.mjs`): build, then fail if `git diff --quiet dist/` reports changes. That prevents future src-only merges from shipping stale compiled output.

Noticed while preparing [#887](https://github.com/CortexReach/memory-lancedb-pro/pull/887) and [#889](https://github.com/CortexReach/memory-lancedb-pro/pull/889); both PRs deliberately left these two files untouched to stay scoped.

TurboTheTurtle:
## Summary
- refresh committed dist output from the current TypeScript sources
- add a package runtime preflight that fails when dist/ is stale after build
- tighten the package runtime assertion wording

Fixes #890

## Verification
- npm run build
- node test/package-runtime.test.mjs


rickthomasjr (approved):
## Summary

**Verdict: APPROVE**

Small, clean PR with two value-add changes plus the dist-freshness guard.

## What This Does

1. Adds `scripts/verify-package-runtime.mjs` dist-freshness check — prevents shipping stale compiled output
2. `process.report.excludeNetwork = true` — WSL2/Tailscale DNS hang fix for lancedb loading
3. Broader legacy memory ID regex — supports more import formats while staying bounded (3-128 chars)
4. `rerankTopN` cap in retriever — limits rerank API calls to candidate pool size

## Code Quality

- **Dist freshness guard**: Clean `git diff --quiet` approach, standard pattern. The `fail()` path is well-handled.
- **`excludeNetwork` fix**: Correct try/catch for Node < 22. The explanation is thorough — `isMusl()` only reads `glibcVersionRuntime`, so excluding network section loses nothing.
- **Legacy ID regex**: `/^[A-Za-z0-9][A-Za-z0-9._:-]{2,127}$/` is a broadening from the old `mem-md-\d+` pattern, intentional for import compatibility. The 3-128 char bound and character class are tight enough to reject garbage.
- **Rerank topN**: `Math.min(results.length, Math.max(1, this.config.candidatePoolSize))` correctly caps to the smaller of actual results and configured pool size.

## Minor Notes

- The dist diff includes regenerated output from both `store.js` and `retriever.js` changes — consistent with build output.
- No test coverage for the legacy ID regex broadening — consider adding a test case in `test/store.test.mjs` for non-`mem-md-` legacy IDs.

## Verdict: APPROVE

All changes are correct, well-justified, and non-breaking. The dist-freshness guard alone is worth merging.

Ready to merge.

rickthomasjr (approved):
## Summary

**Verdict: APPROVE**

Small, clean PR with two value-add changes plus the dist-freshness guard.

## What This Does

1. Adds scripts/verify-package-runtime.mjs dist-freshness check — prevents shipping stale compiled output
2. process.report.excludeNetwork = true — WSL2/Tailscale DNS hang fix for lancedb loading
3. Broader legacy memory ID regex — supports more import formats while staying bounded
4. rerankTopN cap in retriever — limits rerank API calls to candidate pool size

## Code Quality

- Dist freshness guard: Clean git diff --quiet approach, standard pattern
- excludeNetwork fix: Correct try/catch for Node < 22. Well-explained
- Legacy ID regex: Intentional broadening, 3-128 char bound is tight
- Rerank topN: Correctly caps to smaller of results and pool size

## Minor Note

- No test for the legacy ID regex broadening — consider adding a test case for non-mem-md- legacy IDs.

## Verdict: APPROVE

All changes correct, well-justified, non-breaking. Dist-freshness guard alone is worth merging.

Ready to merge.

rickthomasjr (approved):
APPROVE. Clean PR. Dist-freshness guard is valuable. WSL2 DNS fix is correct and well-explained. Legacy ID regex broadening is intentional and bounded. RerankTopN cap prevents unnecessary API calls. Minor: add test for legacy ID regex broadening. Ready to merge.

rickthomasjr (approved):
APPROVE. Clean PR. Dist-freshness guard is valuable. WSL2 DNS fix correct. Legacy ID regex broadening intentional and bounded. RerankTopN cap prevents unnecessary API calls. Minor: add test for legacy ID regex broadening. Ready to merge.

TurboTheTurtle:
Merged current master to clear the conflict in the generated dist output.\n\nRan:\n- npm run build --if-present\n- node test/plugin-manifest-regression.mjs

TurboTheTurtle:
@rwmjhb could you merge #893 when you have a chance? It looks like the final packaging freshness blocker before a beta publish: CI is green, the PR has review approval, and it prevents shipping stale committed dist output.

Once that lands, I’m happy to help smoke-test the 1.1.0-beta.11 package install on current OpenClaw.

rwmjhb (changes requested):
Requesting changes. The direction is useful, but the freshness guard does not currently catch the stale-dist case it is meant to prevent in CI.

## Must fix

The package/runtime freshness check runs without rebuilding `dist` first. `test/package-runtime.test.mjs` directly invokes `scripts/verify-package-runtime.mjs`, and that script only runs `git diff --quiet -- dist`. In the `packaging-and-workflow` GitHub Actions job, CI runs `npm ci` and then `npm run test:packaging-and-workflow`; there is no `npm run build` before this check.

That means a future PR can change `src/` without committing updated `dist/`, and on a clean checkout `git diff -- dist` will still be clean until TypeScript is built. The new guard would pass, so stale generated runtime output could still merge.

Please make the CI/package-runtime path run the build before checking freshness, or add a dedicated script that performs `npm run build` plus the dist freshness verification and use that in CI.

## Also worth fixing

`git diff --quiet -- dist` ignores untracked files. If a build emits a new file under `dist/`, the freshness check can miss it unless it also checks `git status --porcelain --untracked-files=all -- dist` or `git ls-files --others --exclude-standard -- dist`.

Verification note: orchestrator static verification was clean, but build/tests were skipped in R0, so they do not cover this packaging freshness path.

TurboTheTurtle:
Pushed the packaging fix: CI now runs the build before checking dist freshness, and the verifier reports uncommitted or untracked dist output. Packaging workflow is green on b9c218c.

rwmjhb (approved):
Approved. The latest head fixes the packaging freshness gap: `verify-package-runtime` now builds before checking `dist`, and the checker detects tracked and untracked `dist/` dirtiness.

Independent verification on `b9c218c` passed:
- `node test/package-runtime.test.mjs`
- build/package runtime checks left `dist`, package scripts, and tests clean

Non-blocking note: the verifier now assumes a Git checkout because it uses `git status`; that is acceptable for CI/prepack, but worth documenting if anyone runs the script from source archives.

Issue #725, opened 2026-04-30
[Proposal] Temporal Knowledge Graph as a companion to vector memory


mrsteveken:
## Summary

I have been using `memory-lancedb-pro` as my agent's primary long-term memory and it works great for semantic search. However, I found a class of queries that vector search handles poorly: **structured, temporal facts**.

Examples:
- "What version is the trading strategy at?" → needs exact value, not semantic similarity
- "What model did we use in March?" → needs temporal validity tracking
- "What decisions did we make about project X?" → needs structured predicate filtering

## What I Built

As a separate companion plugin (`memory-kg-lite`), I added a temporal fact layer using SQLite:

- **Data model**: Subject → Predicate → Object triples with `valid_from` / `valid_until` windows
- **Storage**: Plain SQLite (no new infra dependencies)
- **Query cost**: Zero LLM tokens — all retrieval is SQL
- **Tools**: `fact_add`, `fact_query`, `fact_invalidate`, `fact_list_recent`

### Example triples

| subject | predicate | object | valid_from | valid_until |
|---|---|---|---|---|
| Steve | timezone | Asia/Shanghai | 2026-02-01 | — |
| MyQuant strategy | version | v11.1 | 2026-02-15 | — |
| H5 data | format | XXXXXX.SZ | 2026-02-05 | — |

### Why a separate plugin?

- Different storage paradigm (relational vs vector)
- Keeps `memory-lancedb-pro` focused on what it does best
- Either plugin can be upgraded/disabled independently
- Optional cross-reference: triples can link to LanceDB memory IDs via `source_id`

## Proposal for Upstream

Two options I'd like to discuss:

1. **Official companion plugin** — A separate package under the same org (e.g., `memory-kg-lite`) with optional LanceDB cross-references
2. **Built-in fact table** — Add an optional SQLite-backed fact layer inside `memory-lancedb-pro` itself (behind a config flag)

I lean toward option 1 (separation of concerns), but wanted to gauge interest before investing more in the standalone approach.

## Measured Benefits

| Query type | Before (vector only) | After (vector + KG) |
|---|---|---|
| "What version is X?" | Semantic search, often returns related-but-wrong results | SQL lookup, exact match, instant |
| "What did we decide about Y?" | Grepping daily logs or hoping vector search hits | `fact_query(subject=Y, category=decision)` |
| "What was true in March?" | Not possible — no temporal tracking | `atDate=2026-03-15` parameter |

## Related Work

[...]

PR #894, opened 2026-06-13
feat: add temporal fact query tool

TurboTheTurtle:
## Summary
- add a read-only memory_fact_query tool for deterministic temporal fact lookup
- filter facts by active/as-of time using existing temporal metadata
- register manifest and CI coverage for the new tool

Refs #725

## Verification
- node --test test/memory-fact-query.test.mjs
- node scripts/verify-ci-test-manifest.mjs
- npm run build

### DIFF
--- a/dist/src/tools.js
+++ b/dist/src/tools.js
@@ -9,7 +9,7 @@ import { join } from "node:path";
 import { isNoise } from "./noise-filter.js";
 import { stripEnvelopeMetadata } from "./smart-extractor.js";
 import { isSystemBypassId, resolveScopeFilter, parseAgentIdFromSessionKey } from "./scopes.js";
-import { appendRelation, buildSmartMetadata, deriveFactKey, parseSmartMetadata, stringifySmartMetadata, } from "./smart-metadata.js";
+import { appendRelation, buildSmartMetadata, deriveFactKey, isMemoryActiveAt, isMemoryExpired, parseSmartMetadata, stringifySmartMetadata, } from "./smart-metadata.js";
 import { classifyTemporal, inferExpiry } from "./temporal-classifier.js";
 import { TEMPORAL_VERSIONED_CATEGORIES } from "./memory-categories.js";
 import { appendSelfImprovementEntry, countSelfImprovementEntries, DEFAULT_SELF_IMPROVEMENT_MAX_ENTRIES, ensureSelfImprovementLearningFiles, } from "./self-improvement-files.js";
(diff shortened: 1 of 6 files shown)


rwmjhb (changes requested):
Requesting changes. The tool is useful, but the current implementation can miss the exact historical fact it is supposed to find.

## Must fix

`memory_fact_query` reads a bounded newest page first, then applies fact-key/text matching in memory:

```ts
runtimeContext.store.list(resolvedScopes.scopeFilter, undefined, Math.max(safeLimit * 10, 100), 0)
```

Production `store.list()` sorts by timestamp descending and slices to the requested limit before returning rows. With the default limit, this means the tool only searches the newest 100 rows before applying `isTemporalFactEntry()` and `factQueryMatches()`.

That can return zero results even when an exact `factKey` or text match exists, as long as the matching temporal fact is older than the overfetch window. This directly breaks the core promise of deterministic temporal fact lookup.

Please move the fact-key/query filtering before result limiting, either by adding a store-level metadata/text predicate or by explicitly paging/exhausting accessible rows before slicing the matching facts. Also add a regression test where the matching historical fact is outside the first production `store.list()` page; the current test stub ignores limit/offset, so it does not catch this.

## Also worth fixing

- `includeHistory=true` currently includes facts whose `valid_from` is after the requested `at` time. Historical as-of results should not show facts that did not exist yet.
- `isTemporalFactEntry()` can exclude valid temporal records whose only temporal marker is `valid_from` and which do not have a fact key.
- Sorting/as-of selection currently relies on `entry.timestamp`, which may diverge from `valid_from` for backfilled or corrected facts.

Verification note: orchestrator full npm test passed, but targeted PR-specific verification/build was not sufficient to cover the bounded-page lookup behavior above.

rwmjhb (approved):
Approved. The new commit fixes the previous blocking issue by scanning fact candidates before applying the result limit, and it adds regression coverage for the paginated lookup and as-of history behavior.

Verification checked:

- `npm run build --if-present` passed on head `b1c9ef65f5174a4675f4cfdbb5e7cd9388639f18`
- `node --test test/memory-fact-query.test.mjs` passed, 6/6
- `node test/plugin-manifest-regression.mjs` passed

- `memory_fact_query` is a new memory read path but does not currently apply the USER.md-exclusive recall filter used by `memory_recall`. If that boundary is enabled, exact fact queries can still expose matching stored facts through text output and `details.facts`.
- The paginated scan calls production `store.list(..., 500, offset)` repeatedly; because `store.list()` fetches/sorts scoped rows before slicing, large no-match scans can repeatedly reload and sort the same scoped store.
- The scan accumulates `seenIds` and matching facts for the whole scoped store before the final limit is applied, so large stores can use more per-call memory than necessary.

rwmjhb (approved):
Approved. The latest update fixes the boundary-filter issue, and the added regression coverage now protects USER.md-exclusive fact filtering through the fact query details path.

I also verified this head independently:
- `node --test test/memory-fact-query.test.mjs`

Non-blocking follow-ups remain worth considering later: require an explicit selector for broad fact queries, and keep the temporal-fact candidate heuristic tight so ordinary smart-memory metadata does not become a listing-like path by accident.

rwmjhb (approved):
Approved. The latest commit adds the explicit selector guard and regression coverage for ordinary smart metadata/static fact boundaries. I verified this head independently:
- `node --test test/memory-fact-query.test.mjs`

Remaining notes are non-blocking: text-only fact queries intentionally avoid broad static-memory matching, so callers that need deterministic lookup for stable current facts should use `factKey` or a future stricter structured selector; large-store scans may also need indexing/predicate support later.

PR #896, opened 2026-06-13
fix: cap auto-recall rerank timeout

### LINKED ISSUE #882: autoRecallTimeoutMs and rerankTimeoutMs share the same 5s default, making the rerank cosine fallback unreachable in auto-recall
**Version:** 1.1.0-beta.11, OpenClaw 2026.6.1 (engine 2026.5.28), remote Jina embeddings + jina-reranker-v3

## Summary

The auto-recall pipeline and the rerank HTTP call both default to a 5000ms timeout. Because the rerank call starts only after the embedding round trip and hybrid search have already consumed part of the outer budget, the outer auto-recall timer always expires before the rerank abort can fire. The graceful cosine fallback (implemented in [src/retriever.ts](https://github.com/CortexReach/memory-lancedb-pro/blob/master/src/retriever.ts), "falling back to cosine") is therefore unreachable in auto-recall with default settings: its result is computed after the deadline and discarded at the late-result checkpoints ("dropping late auto-recall result after timeout at post-retrieve / pre-metadata").

Net effect: a slow rerank provider converts what should be a degraded-but-useful injection (cosine-reranked hybrid results) into a total loss of memory injection for that message, even though a fallback exists for exactly this case.

## Mechanics

1. Outer budget: the whole recall chain (embed, hybrid search, rerank, metadata) races a timer, `autoRecallTimeoutMs ?? 5000` ([index.ts](https://github.com/CortexReach/memory-lancedb-pro/blob/master/index.ts), AUTO_RECALL_TIMEOUT_MS).
2. Inner abort: the rerank fetch has an AbortController with `rerankTimeoutMs ?? 5000` and a cosine fallback on abort.
3. With a remote embedding API, the embed step alone takes 0.5 to 2s. A hung rerank then cannot abort before the outer 5000ms expires. The fallback never gets to deliver.

## Observed

Two incidents on one fleet (remote Jina free tier, which throttles):

- `memory-lancedb-pro: auto-recall timed out after 5000ms; skipping memory injection to avoid stalling agent startup`
- `memory-lancedb-pro: dropping late auto-recall result after timeout at <stage> for agent <id>`
- After raising the outer budget (workaround below): `Rerank API timed out (5000ms), falling back to cosine` followed by a successful injection, confirming the fallback works once it has room.

## Suggestions

1. Couple the defaults: derive the rerank abort from the remaining outer budget (for example `min(rerankTimeoutMs, remainingAutoRecallBudget - margin)`), or ship a lower rerank default so abort + cosine fallback fit inside the outer 5000ms.
2. Optional hardening: a single retry on HTTP 429 before falling back, since rate-limited rerank tiers commonly return 429 on burst.
3. A note in the integration playbook about the relationship between these two timeouts would help operators tune them together.

## Workaround we applied

`autoRecallTimeoutMs: 10000` plus explicit `retrieval.rerankTimeoutMs: 5000`. Verified the same night: a real Jina timeout produced the cosine fallback line and the injection still landed within budget.

TurboTheTurtle:
## Summary
- pass a per-call rerank timeout through retrieval context
- default auto-recall reranking to a shorter timeout inside the recall budget
- preserve explicit user rerank timeout configuration

Fixes #882

## Verification
- node --test test/auto-recall-timeout.test.mjs
- node test/retriever-rerank-regression.mjs
- npm run build

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -2253,6 +2253,7 @@ const memoryLanceDBProPlugin = {
             });
             const AUTO_RECALL_TIMEOUT_MS = parsePositiveInt(config.autoRecallTimeoutMs) ?? 5_000; // configurable; default raised from 3s to 5s for remote embedding APIs behind proxies
             api.on("before_prompt_build", async (event, ctx) => {
+                const autoRecallDeadlineMs = Date.now() + AUTO_RECALL_TIMEOUT_MS;
                 // Skip auto-recall for sub-agent sessions — their context comes from the parent.
                 const sessionKey = typeof ctx.sessionKey === "string" ? ctx.sessionKey : "";
                 if (sessionKey.includes(":subagent:"))
@@ -2349,7 +2350,10 @@ const memoryLanceDBProPlugin = {
                         source: "auto-recall",
                         signal: autoRecallAbortController.signal,
                         ...(autoRecallRerankTimeoutMs !== undefined
-                            ? { rerankTimeoutMs: autoRecallRerankTimeoutMs }
+                            ? {
+                                rerankTimeoutMs: autoRecallRerankTimeoutMs,
+                                rerankDeadlineMs: autoRecallDeadlineMs,
+                            }
                             : {}),
                     }), config.workspaceBoundary);
                     if (shouldDropLateAutoRecall("post-retrieve"))
--- a/dist/index.js
+++ b/dist/index.js
@@ -165,8 +165,10 @@ function getAutoRecallRerankTimeoutMs(config, retrievalConfig, autoRecallTimeout
     if (!Number.isFinite(autoRecallTimeoutMs) || autoRecallTimeoutMs <= 0)
         return undefined;
     const halfBudget = Math.floor(autoRecallTimeoutMs / 2);
+    if (halfBudget < 100)
+        return 0;
     if (autoRecallTimeoutMs <= 1_000)
-        return Math.max(100, halfBudget);
+        return halfBudget;
     return clampInt(halfBudget, 500, 2_500);
 }
 export function buildAutoRecallRerankCostWarning(config, retrievalConfig = { ...DEFAULT_RETRIEVAL_CONFIG, ...(config.retrieval || {}) }) {
--- a/dist/index.js
+++ b/dist/index.js
@@ -157,6 +157,18 @@ function getAutoRecallRetrieveLimit(autoRecallMaxItems) {
 function getAutoRecallRerankInputLimit(retrieveLimit) {
     return clampInt(retrieveLimit, 1, 20) * 2;
 }
+function getAutoRecallRerankTimeoutMs(config, retrievalConfig, autoRecallTimeoutMs) {
+    if (retrievalConfig.rerank !== "cross-encoder" || !retrievalConfig.rerankApiKey)
+        return undefined;
+    if (typeof config.retrieval?.rerankTimeoutMs === "number")
+        return undefined;
+    if (!Number.isFinite(autoRecallTimeoutMs) || autoRecallTimeoutMs <= 0)
+        return undefined;
+    const halfBudget = Math.floor(autoRecallTimeoutMs / 2);
+    if (autoRecallTimeoutMs <= 1_000)
+        return Math.max(100, halfBudget);
+    return clampInt(halfBudget, 500, 2_500);
+}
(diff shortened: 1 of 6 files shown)


rwmjhb (commented):

GitHub currently reports this branch as `mergeable=CONFLICTING` / `merge_state_status=DIRTY`, so the orchestrator stopped at the conflict gate after R0/R1. The issue is real and the direction looks worth reviewing, but the implementation needs to be checked after conflict resolution.

rickthomasjr (approved):
## Agent Review: fix: cap auto-recall rerank timeout

**Verdict: APPROVE (minor concern noted)**

### Summary
Andy added a per-call `rerankTimeoutMs` override to the auto-recall retrieval path. The auto-recall timeout budget (default 5000ms) gets split so the rerank step uses at most half the budget (500-2500ms clamped), leaving room for fallback work (cosine similarity) within the total timeout.

### Changes
- New function `getAutoRecallRerankTimeoutMs()` — guards on cross-encoder + API key + finite budget
- `RetrievalContext` gains optional `rerankTimeoutMs` field
- `hybridRetrieval()` and `rerankResults()` propagate the override down
- `effectiveTimeoutMs` = `timeoutOverrideMs ?? config.rerankTimeoutMs ?? 5000`
- 1 new test verifies the 2500ms default when budget is 5000ms

### Correctness Assessment

**Guard conditions are sound:**
1. Only applies when `rerank === "cross-encoder"` AND `rerankApiKey` is set
2. Returns `undefined` if explicit `config.retrieval.rerankTimeoutMs` is already set
3. Returns `undefined` if `autoRecallTimeoutMs` is non-finite or <= 0

**Timeout budgeting logic:**
- `halfBudget = autoRecallTimeoutMs / 2`
- Clamped to [500, 2500]
- Edge case: if `autoRecallTimeoutMs <= 1000`, returns `max(100, halfBudget)` — a bit low but only applies when the total budget is very tight

**Propagation chain is clean:**
autoRecall hook -> RetrievalContext.rerankTimeoutMs -> hybridRetrieval() -> rerankResults()
All optional, all fall through to existing defaults. No breaking changes.

### Minor Concern
The 100ms minimum for very tight budgets (<=1000ms) could cause noisy fallback. At 100ms, most cross-encoder APIs will timeout, triggering the cosine fallback. Unlikely to be a production issue but worth noting.

### Test Coverage
1 new test verifies the 2500ms case. Additional tests for edge cases would be good but not blocking.

### Verdict: APPROVE
Simple, correct fix. Low risk. Ready to merge.

rwmjhb (approved):
Approved. The rebased/current version is focused and fixes the default auto-recall/rerank timeout interaction without changing explicit user timeout configuration.

Verification checked:

- `npm run build --if-present` passed on head `8d281bc4d3c3385698ae7b408db9e48c427531f3`
- `node --test test/auto-recall-timeout.test.mjs` passed, 4/4

Non-blocking note: for very small `autoRecallTimeoutMs` values, the helper can still spend up to the 100ms floor on a remote rerank call that will almost certainly abort before cosine fallback. Correctness is preserved, but this default may be worth revisiting if tight-budget configs are common.

rwmjhb (approved):
Approved. I reran verification with a longer full-suite timeout after the orchestrator's 180s full-test gate timed out; the current head passes build, targeted timeout/rerank tests, and the full npm suite.

Verification checked on head `a50bbbb31e8566f58f955549c24c4ee14ca74a0c`:

- `npm run build --if-present` passed
- `node --test test/auto-recall-timeout.test.mjs` passed, 5/5
- `node test/retriever-rerank-regression.mjs` passed
- `npm test` passed under a 420s limit

Non-blocking follow-up: the current implementation still computes the auto-recall rerank timeout up front from the full auto-recall budget. If embedding/search consume most of the budget before rerank starts, a deadline/remaining-budget based timeout would be more precise and would fully close the slow pre-rerank path.

rwmjhb (approved):
Approved. The latest commit now caps rerank against the auto-recall deadline and adds coverage for exhausted/expired rerank budgets. I verified this head independently:
- `node --test test/auto-recall-timeout.test.mjs`
- `node test/retriever-rerank-regression.mjs`

Non-blocking follow-up: the remote rerank timeout can still consume all remaining deadline time when pre-rerank work has already used most of the outer budget. Reserving a small fallback margin and reporting the effective clipped timeout would make this path easier to reason about under slow providers.

PR #902, opened 2026-06-19
fix(cli): write memory-pro JSON to stdout

ScientificProgrammer:
## Summary

- write `memory-pro` machine-readable JSON payloads directly to stdout instead of `console.log`
- keep human-facing CLI output on `console.log`
- update CLI smoke/e2e tests to distinguish direct stdout from console/log output for JSON commands
- include rebuilt `dist/cli.js` so packaged runtime output stays fresh

## Root cause

OpenClaw routes console/log output to stderr in JSON mode so diagnostics do not corrupt machine-readable stdout. `memory-pro list/search/stats/export --json` was still emitting payloads with `console.log(formatJson(...))`, so the JSON payload followed the diagnostic stream and could be captured by `2>stderr.log` along with plugin loader logs.

That inverted the intended stream contract:

- diagnostics/logs: stderr
- machine-readable JSON payloads: stdout

This patch adds `writeStdout()` / `writeJson()` helpers and uses them for JSON payload branches so the payload bypasses console capture/routing.

## Tests

- `npm run test:cli-smoke`
- `npm run build`

## Notes

This fixes the `memory-pro` payload side of the stdout/stderr issue. If OpenClaw core still emits plugin startup chatter on stdout before plugin command dispatch, that should be handled separately in OpenClaw core (same class as openclaw/openclaw#81536).

### DIFF
--- a/cli.ts
+++ b/cli.ts
@@ -486,6 +486,14 @@ function formatJson(obj: any): string {
   return JSON.stringify(obj, null, 2);
 }
 
+function writeStdout(text: string): void {
+  process.stdout.write(text.endsWith("\n") ? text : `${text}\n`);
+}
+
+function writeJson(obj: any): void {
+  writeStdout(formatJson(obj));
+}
+
 function formatRetrievalDiagnosticsLines(diagnostics: {
   originalQuery: string;
   bm25Query: string | null;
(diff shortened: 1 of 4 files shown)


rwmjhb (approved):
Approved. Orchestrator verdict: approve on head b37611b.

- npm ci --silent
- npm run build --if-present
- node test/cli-smoke.mjs
- node test/functional-e2e.mjs
- git diff --exit-code -- cli.ts dist/cli.js test/cli-smoke.mjs test/functional-e2e.mjs package.json package-lock.json openclaw.plugin.json

All passed. Non-blocking follow-ups: consider changing `writeJson(obj: any)` to `unknown`, and consider making the JSON search-error exit path avoid forced exit before stdout has flushed.

PR #904, opened 2026-06-27
Add an optional model override for memoryReflection

gorkem2020:
## Problem

`memoryReflection` generates a session-background reflection on reset, but there is no way to choose which model performs that generation. It resolves the model implicitly from the running agent's primary model, then falls back to the host's global default when no `memoryReflection.agentId` is set.

For deployments whose global default is a router alias (for example an `auto` routing model), every reflection is silently free-routed to whatever the router selects, instead of a deterministic, cost-appropriate model the operator chose. The plugin's own `llm.model` is shadowed by the agent-primary resolution, so today there is no config path to control the reflection model. The schema is also `additionalProperties: false`, so `model` cannot even be set as a passthrough; it is rejected at validation.

This is a consistency gap rather than a new idea: the sibling `memoryFlush` already exposes a `model` option, and the dreaming engine reserves `dreaming.model`. `memoryReflection` is the only one of the three with no override.

## Change

Add a `memoryReflection.model` knob that takes precedence at the front of the existing resolution chain, ahead of the agent-primary and `llm.model` fallbacks. Fully opt-in and backward compatible: when unset, resolution is unchanged.

- `openclaw.plugin.json`: declare `memoryReflection.model` (string) inside the existing `additionalProperties: false` block, plus a `uiHints` entry mirroring `memoryReflection.agentId` and `dreaming.model`.
- `index.ts`: thread the value through the config type, config normalization, the reflection read site, the `generateReflectionText` call, and the `modelRef` resolution chain.
- `test/plugin-manifest-regression.mjs`: assert the new knob is present with type string, that the schema still rejects unknown keys, and that the `uiHints` entry exists.

## Resolution precedence

`memoryReflection.model`, then `resolveAgentPrimaryModelRef(...)`, then the `llm.model` fallback. If both `memoryReflection.model` and `memoryReflection.agentId` are set, the configured model wins for model selection while `agentId` still selects which agent runs the reflection. This matches the precedence and semantics of `memoryFlush.model`.

## Testing

- `npm run build` (tsc) passes.
- `node test/plugin-manifest-regression.mjs` passes, including the new assertions.
- Config normalization verified on the built output: `memoryReflection.model` set survives `parsePluginConfig` into the parsed config; absent or blank resolves to `undefined`.
- Live end to end: with `memoryReflection.model` pinned to a specific model, resetting an agent produced a reflection generation that resolved to the pinned model instead of the global default, running via the embedded runner with no fallback.
- The reflection suites show no new failures versus the same revision without this change (a few pre-existing failures in `writeLegacyCombined` / `sessionStrategy` default tests are unrelated to this PR).

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -1173,7 +1173,8 @@ async function generateReflectionText(params) {
                 const modelRefFromConfig = llmConfig?.model;
                 // Model resolution chain: agent-specific primary model ref > global llm.model fallback.
                 // The typeof guard ensures a non-string value (e.g. number) does not reach splitProviderModel as-is.
-                const modelRef = resolveAgentPrimaryModelRef(params.cfg, params.agentId)
+                const modelRef = params.model
+                    ?? resolveAgentPrimaryModelRef(params.cfg, params.agentId)
                     ?? (typeof modelRefFromConfig === "string" ? modelRefFromConfig : undefined);
                 // Provider resolution chain: parsed from modelRef (e.g. "minimax/MiniMax-M2.7") > inferred from baseURL.
                 // inferProviderFromBaseURL uses .endsWith(".suffix") to prevent subdomain spoofing.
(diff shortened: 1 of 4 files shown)


rwmjhb (approved):
Approved on head `5257ae2`. Orchestrator verdict: approve.

- npm ci --silent
- npm run build --if-present
- node --test test/reflection-bypass-hook.test.mjs
- node --test test/startup-health-diagnostics.test.mjs
- npm test

All passed. Non-blocking follow-up: the embedded reflection path honors `memoryReflection.model`, but the CLI fallback path still appears not to pass a model/provider override. Consider threading it through or warning/failing clearly when fallback cannot honor the configured override.

Issue #848, opened 2026-06-01
Project Status Inquiry: 2.5 months without updates - is this project still maintained?


DerekEXS:
Hi @CortexReach,

I noticed this repository has had no commits or activity for approximately 2.5 months (last push: 2026-03-23). I want to check in before raising concerns, and I'm hoping to get a clear answer on the project's status.

## What I observed

- Last push: 2026-03-23
- Last general update: 2026-02-24
- No formal releases published (only one tag: v1.1.0)
- Stars: 2
- Open issues: 2 (unanswered)
- No recent activity from the maintainer

## My use case

I've been using the npm-published community fork `@ruliu/memory-lancedb-pro` (v1.1.9), which appears to be a maintained continuation of your work. While debugging integration with OpenClaw 2026.5.27+, I encountered a **fundamental architectural issue** that I believe originates from this plugin's design rather than configuration:

### The core problem
The plugin's management tools (`memory_store`, `memory_recall`, etc.) are **not exposed to the Agent** through OpenClaw's tool routing system. Despite:
- `enableManagementTools: true` being set
- `hooks.allowConversationAccess: true` being configured
- The LanceDB database being healthy (356 entries intact)
- `embedding`, `retrieval`, and `rerank` all working at the API level

...the Agent cannot see or call any of these tools. It falls back to writing MD files and using LCM (lossless-claw) compression history as a poor man's memory.

## Specific questions

1. **Are you still actively maintaining this project?**
2. **Is there a successor repository or official handoff to the `@ruliu` community fork?**
3. **Have you seen the same tool-exposure issue in newer OpenClaw versions?** Is this a known regression?
4. **Would you be open to adding a `contracts.tools` declaration** (currently missing in the community fork, causing OpenClaw startup warnings) to make tool exposure explicit?

## Suggestions if you'd like to revive it

If you do have time/interest, here are concrete improvements that would dramatically increase adoption:

1. **Publish formal releases to npm** (currently only `v1.1.0` tag exists). Semantic versioning helps users trust upgrades.
2. **Add `contracts.tools` declaration** in `openclaw.plugin.json` for OpenClaw 2026.5.27+ compatibility. This eliminates startup warnings and may fix the tool-routing issue.
3. **Document the auto-capture behavior** explicitly. Currently `agent_end` hook triggering is unpredictable and undocumented.
4. **Consider explicit `enableManagementTools: true` default** in the schema, since the management tools are the most valuable feature.
[...]

PR #906, opened 2026-07-01
docs: clarify hook permissions and input cap

TurboTheTurtle:
Refs #834.

Carries forward the remaining documentation piece from the closed PR after the code changes were superseded or merged into current `master`.

Summary:
- show `embedding.maxInputChars` in the full config example
- clarify `hooks.allowConversationAccess` for custom/non-bundled plugin entries that override manifest hook metadata

### DIFF
--- a/README.md
+++ b/README.md
@@ -476,6 +476,7 @@ For `source: "file"`, `id` is resolved through OpenClaw's `api.resolvePath()` an
     "taskQuery": "retrieval.query",
     "taskPassage": "retrieval.passage",
     "normalized": true,
+    "maxInputChars": 1400,
     "clientTimeoutMs": 30000
   },
   "dbPath": "~/.openclaw/memory/lancedb-pro",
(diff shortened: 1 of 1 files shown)


AliceLJY (approved):
Verified against source: `embedding.maxInputChars` is a real config field (embedder.ts + manifest schema; default derives from the model, so 1400 works as an explicit example), and the `hooks.allowConversationAccess` guidance matches the hook-gating behavior discussed in #848/#850. Useful doc coverage for the non-bundled install case. LGTM.

rwmjhb (approved):
Approved on head `85adeb9`.

The orchestrator value gate passed for this docs-only change; the later R2a model call hung before producing findings, so I completed the review manually against the README diff.

- npm ci --silent
- npm run build --if-present
- git diff --check
- npm test

All passed. The README additions match existing config behavior: `embedding.maxInputChars` is already supported, and the hook permission guidance is scoped to custom/non-bundled entries overriding manifest hook metadata.

PR #907, opened 2026-07-01
fix: run memory reflections in per-agent lanes with a bounded concurrency cap

gorkem2020:
## Problem

Every memory-reflection embedded run is enqueued with the fixed sessionKey `temp:memory-reflection`. Host session lanes are serial by default (`maxConcurrent: 1`), so reflections for all agents on a gateway share one global serial lane, and the queue wait happens inside each run's timeout window. Under a concurrent-reflection burst (for example a mass `/reset` across several agents) queued runs burn their timeout budget (`memoryReflection.timeoutMs` plus the fixed grace) before they even start, then fail over and retry. Observed live on a 6-agent gateway: `lane wait exceeded ... queueAhead` climbed to 4 and every queued reflection hit its deadline, recovering only via fallback.

## Change

1. Per-agent lanes: the embedded-run sessionKey becomes `temp:memory-reflection:<agentId>` (the agent id already flows into the same call). Lanes are independent across agents while each agent's own reflections stay serial. Both `startsWith("temp:memory-reflection")` guards and the `temp:*` exclusion pattern semantics are unchanged, covered by tests.
2. New `memoryReflection.maxConcurrentRuns` knob (integer, default 1, minimum 1): a plugin-side cap on concurrent reflection runs across all agents, acquired before the run's timeout clock starts.

**Default behavior is unchanged.** With the default of 1, reflections stay fully serialized exactly as before; the difference is that the serialization wait now happens before the timeout clock starts instead of inside it, so queued reflections no longer time out in line and cascade into failover. Raising the cap is an explicit opt-in to parallel reflections across agents for operators whose reflection provider has the headroom.

## Notes for reviewers

- The params of `generateReflectionText` moved to a named type (`GenerateReflectionTextParams`) so a thin exported wrapper can gate the run without re-indenting the whole function body. One static smoke assertion in `test/issue606_sdk-migration.test.mjs` matched the old inline literal shape; it now guards the same invariant (params include `api: OpenClawPluginApi`, the function consumes that params type and returns a Promise) and is slightly stricter.
- Covers the manifest schema (`additionalProperties: false`), a uiHints entry, manifest-regression assertions, and a new test file `test/reflection-per-agent-lane.test.mjs`: the per-agent key reaches the runner, `temp:*` exclusion still matches, a 4-run burst with the cap raised to 2 is bounded at 2 with FIFO handoff and no lost runs, and config parsing covers defaults, flooring, and invalid-value fallback.

## Validation

New tests 6 of 6, `test:core-regression` group green, `plugin-manifest-regression` green, typecheck clean. The 4 pre-existing local failures in `test/memory-reflection.test.mjs` on this environment are identical on unmodified `master` (environment dependent, unrelated to this change).

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -359,6 +359,7 @@ const DEFAULT_REFLECTION_MESSAGE_COUNT = 120;
 const DEFAULT_REFLECTION_MAX_INPUT_CHARS = 24_000;
 const DEFAULT_REFLECTION_TIMEOUT_MS = 20_000;
 const DEFAULT_REFLECTION_THINK_LEVEL = "medium";
+const DEFAULT_REFLECTION_MAX_CONCURRENT_RUNS = 1;
 const DEFAULT_REFLECTION_ERROR_REMINDER_MAX_ENTRIES = 3;
 const DEFAULT_REFLECTION_DEDUPE_ERROR_SIGNALS = true;
 const DEFAULT_REFLECTION_SESSION_TTL_MS = 30 * 60 * 1000;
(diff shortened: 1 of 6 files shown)


AliceLJY (approved):
Reviewed in depth. The diagnosis is right: the fixed `temp:memory-reflection` sessionKey serialized every agent's reflections through one host lane, with queue wait burning each run's timeout budget. The fix is well-executed:

- Per-agent sessionKey keeps lanes independent while preserving the `startsWith("temp:memory-reflection")` guards and `temp:*` exclusion semantics (covered by tests).
- The plugin-side slot is acquired **before** the run's timeout clock starts — that's the key property that stops queued runs from burning their deadline.
- Default `maxConcurrentRuns: 1` keeps behavior fully serialized exactly as before; parallelism is strictly opt-in.
- Release is idempotent, waiters are FIFO, and `reflection-per-agent-lane.test.mjs` exercises the burst timing properly (cap held across the whole burst, all runs complete).
- Manifest schema + uiHints + plugin-manifest regression tests updated in sync; dist rebuilt.

CI green across all suites. LGTM.

rwmjhb (approved):
Approved on head `9982431`. Orchestrator verdict: approve.

- npm ci --silent
- npm run build --if-present
- node --test test/reflection-per-agent-lane.test.mjs
- node test/issue606_sdk-migration.test.mjs
- npm test

All passed. Non-blocking follow-ups: consider enforcing the manifest maximum for `memoryReflection.maxConcurrentRuns` in runtime parsing, and add coverage for concurrent reflections sharing the same configured reflection agent when `maxConcurrentRuns > 1`.

Issue #819, opened 2026-05-25
OpenAI embedder client has no HTTP timeout — gateway startup hangs under WSL2 network pressure


srb11e:
## Summary

`src/embedder.ts` creates OpenAI clients without a `timeout` parameter. During rapid gateway restarts under WSL2 network pressure, `embedder.test()` can hang waiting for the OS-level TCP timeout. This causes the gateway's startup-check gate to fail, triggering unclean terminations and orphaned processes.

## Reproduction

1. Run OpenClaw 2026.5+ on WSL2 with LanceDB Pro configured
2. Restart the gateway service rapidly (e.g., during config changes)
3. Observe: `embedder.test()` hangs → 8-second startup gate times out → gateway terminates uncleanly → leftover `node` processes

## Fix

Add `timeout: 6_000` (or equivalent) to the OpenAI client constructor in `src/embedder.ts`:

```typescript
// Before (~line 311):
this.clients = resolvedKeys.map(key => new OpenAI({
  apiKey: key,
  ...(config.baseURL ? { baseURL: config.baseURL } : {}),
}));

// After:
this.clients = resolvedKeys.map(key => new OpenAI({
  apiKey: key,
  ...(config.baseURL ? { baseURL: config.baseURL } : {}),
  timeout: 6_000,
}));
```

## Context

Discovered running OpenClaw 2026.5.6 with `models/gemini-embedding-2-preview` via Gemini OpenAI-compat endpoint. The OpenAI JS client v4 defaults to no timeout, which is fine for normal operation but fails under WSL2 network pressure during rapid restarts.

PR #908, opened 2026-07-03
fix: keep one-time cold-start costs out of the startup retrieval probe

gorkem2020:
## Summary

On a cold gateway boot the startup retrieval probe reliably misses its 8s budget (`startup check retrieval failed (8413ms): retrieval startup check timed out after 8000ms`) while the embedding probe passes in ~1.5s and every later retrieval is fast. Root cause, traced on a live 7-agent deployment with a ~1900-entry table: the retrieval probe was the **first store touch of the process**, so its timed window absorbed the one-time `doInitialize()` costs (LanceDB connect, table open, FTS index build under the write lock) **plus** a first-ever cold rerank vendor roundtrip - `retriever.test()` passed no rerank budget, so the cross-encoder path fired even on a `limit: 1` wiring check. Neither cost belongs to what the probe is trying to measure.

Three fixes plus one control:

1. **Warm the store before the timers** - `runStartupChecks` now runs a reported `store` phase (`store.ensureInitialized()`) before the embedding/retrieval probes, so the probes measure steady-state behavior instead of first-open costs. `ensureInitialized` becomes public for this; the retriever already reached it through an `as unknown as` cast.
2. **Probe with a zero rerank budget** - `retriever.test()` passes `rerankTimeoutMs: 0`, which trips the existing skip guard, keeping the cold remote rerank roundtrip out of a vector/FTS wiring check.
3. **Honest timeout reporting** - a timed-out probe used to surface as `mode: undefined, FTS: disabled` in the init summary even though FTS was never measured; those fields now print `unknown`.
4. **`startupCheckTimeoutMs` knob** (default `8000`, behavior preserved) as an operator control for genuinely slow hosts; the embedding probe's inner timeout derives from the budget instead of a second hardcoded constant.

Prior reports of this symptom were closed as likely-stale with an invitation to re-report if it reproduces: [#60](https://github.com/CortexReach/memory-lancedb-pro/issues/60), [#218](https://github.com/CortexReach/memory-lancedb-pro/issues/218) (related: [#819](https://github.com/CortexReach/memory-lancedb-pro/issues/819)). It reproduces on current master; this PR carries the fix instead of a new report.

## Tests

- `test/startup-check-timeout.test.mjs` (new, wired into the npm test chain): knob default/override/garbage-fallback, red without the change.
- `test/startup-health-diagnostics.test.mjs`: new case proving the health probe makes **zero** rerank vendor calls (fetch stub that fails the test if invoked) while still succeeding.
- `tsc --noEmit` clean; `dist/` rebuilt and committed (packaging check).
- Full test chain green locally, with one environmental exclusion: `test/cjk-recursion-regression.test.mjs` binds `127.0.0.1:11434` by design and collides with a real Ollama on this host; untouched by this change.


rwmjhb (approved):
Approved on head `652d3b4`. Orchestrator verdict: approve.

- npm ci --silent
- npm run build --if-present
- node --test test/startup-check-timeout.test.mjs
- node --test test/startup-health-diagnostics.test.mjs
- node test/retriever-rerank-regression.mjs
- npm test

All passed. Non-blocking follow-ups: strengthen the rerank-skip test so it exercises the hybrid rerank path, and consider clamping `startupCheckTimeoutMs` in runtime parsing to match the manifest range.

Issue #524, opened 2026-04-04
Bug: Smart Extraction path does not write to mdMirror (dual-write incomplete)


arlen8411:
## Bug Description

The Smart Extraction path (LLM-powered 6-category extraction) does not write to mdMirror, even though mdMirror is enabled. Only the Regex Fallback path writes to mdMirror.

## Expected Behavior

According to the code comment \"Markdown Mirror (dual-write)\", all memory entries written to LanceDB should also be written to mdMirror as a backup. This includes memories created via Smart Extraction.

## Evidence

From the logs, when Smart Extraction succeeds it directly returns without calling mdMirror:

\`\`\`typescript
// index.ts around line 2248
const stats = await smartExtractor.extractAndPersist(...);
if (stats.created > 0 || stats.merged > 0) {
  api.logger.info(\`smart-extracted \${stats.created} created...\`);
  return; // ← exits here, no mdMirror call
}
\`\`\`

The mdMirror write only exists in the Regex Fallback path (store.store() call followed by mdMirror() call).

## Impact

- LanceDB: ✅ Memory entries created
- mdMirror: ❌ No .md file backup created
- This breaks the \"dual-write\" guarantee stated in the code comment

## Question

Is this behavior intentional? If so, the documentation and code comment should be updated to clarify that Smart Extraction memories are NOT mirrored to .md files. If not, this is a bug that should be fixed.

## Environment

- memory-lancedb-pro: 1.1.0-beta.9
- OpenClaw: 2026.4.2
- Configuration: smartExtraction: true, mdMirror: { enabled: true }

PR #911, opened 2026-07-04
fix: mirror smart-extraction persisted memories to markdown

### LINKED ISSUE #910: Bug: Smart Extraction path does not write to mdMirror (still reproducible, see #524)
### Summary

When `mdMirror.enabled: true` and Smart Extraction is the active capture path (the default when `smartExtraction` is not explicitly disabled), memories created or merged through `SmartExtractor.extractAndPersist()` are never mirrored to markdown. The Regex Fallback and Reflection capture paths mirror correctly; Smart Extraction silently does not.

This was already reported in [issue #524](https://github.com/CortexReach/memory-lancedb-pro/issues/524) in April, but that issue was closed with no linked fix. It is still reproducible on current master.

### Root cause

`SmartExtractor.extractAndPersist()` (`src/smart-extractor.ts`) persists every create and merge directly through `store.bulkStore()` / `store.store()` / `store.update()`. It has no concept of a markdown-mirror hook at all, unlike the tool-invoked write paths in `src/tools.ts` and the reflection path in `index.ts`, which both call the `mdMirror` writer after a successful persist.

Separately, the auto-capture hook in `index.ts` returns early as soon as Smart Extraction reports any created or merged memories ("smart extraction handled everything"), before ever reaching the Regex Fallback block where the `mdMirror` call actually lives. So even the successful counts logged by Smart Extraction never translate into a markdown write.

### Repro

1. Enable both `mdMirror.enabled: true` and Smart Extraction (default on).
2. Have a conversation that triggers auto-capture with enough content for Smart Extraction to create or merge a memory.
3. Check the configured mdMirror directory: no `.md` entry is written for that memory, even though the same memory is present in LanceDB and the logs report `created` / `merged` counts greater than zero.
4. The Regex Fallback path (triggered when Smart Extraction produces zero candidates) does write to mdMirror correctly, confirming the mirror writer itself works and the gap is specific to the Smart Extraction path.

### Impact

Anyone relying on the markdown mirror as a human-readable backup or audit trail of captured memories silently loses coverage for the majority of captures, since Smart Extraction is the default and preferred capture path. There is no error or warning; the mirror directory simply never receives these entries.

### Fix

A pull request follows this issue, adding an optional `onPersisted` callback to `SmartExtractorConfig` (mirroring the existing `onAdmissionRejected` injection pattern already used for admission-control audit logging) that fires once per successful create and once per successful merge, wired to the same `mdMirror` writer instance used by the other capture paths.

gorkem2020:
### What this fixes

With `mdMirror.enabled: true` and Smart Extraction active (the default capture path), created and merged memories are never mirrored to markdown. The extractor persists directly through `store.bulkStore()` / `store.store()` / `store.update()` and has no mirror hook, and the auto-capture handler returns early on smart-extraction success, before the code block that performs mirror writes for the regex-fallback path.

### How

Adds an optional `onPersisted` callback to `SmartExtractorConfig`, following the same constructor-injection pattern already used by `onAdmissionRejected` for admission-control audit logging. The callback fires after each successful create and merge, is awaited with failures swallowed and logged (a mirror error can never fail the store), and is wired in `index.ts` to the same `mdMirror` writer instance the other capture paths use. The writer is now created once alongside the extractor singleton instead of per `register()` call.

### Tests

New `test/smart-extractor-md-mirror.test.mjs` (4 tests): callback fires on create with the mirror entry shape, fires on merge with post-merge text, is skipped when unconfigured, and callback failures do not fail the store. Full existing smart-extractor, mdmirror-fallback-dir, regex-fallback, and plugin-bootstrap suites pass.

Fixes #910 (previously reported as #524, which was closed without a linked fix and still reproduces on current master).

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -1786,6 +1786,9 @@ function _initPluginState(api) {
         api.logger.info(`memory-lancedb-pro: CLAWTEAM_MEMORY_SCOPE added scopes: ${clawteamScopes.join(", ")}`);
     }
     const migrator = createMigrator(store);
+    // Created here (ahead of SmartExtractor) because SmartExtractor's onPersisted
+    // callback below closes over it.
+    const mdMirror = createMdMirrorWriter(api, config);
     let smartExtractor = null;
     if (config.smartExtraction !== false) {
         try {
@@ -1828,6 +1831,7 @@ function _initPluginState(api) {
                 workspaceBoundary: config.workspaceBoundary,
                 admissionControl: config.admissionControl,
                 onAdmissionRejected: admissionRejectionAuditWriter ?? undefined,
+                onPersisted: mdMirror ?? undefined,
                 log: (msg) => api.logger.info(msg),
                 debugLog: (msg) => api.logger.debug(msg),
                 noiseBank,
@@ -1874,6 +1878,7 @@ function _initPluginState(api) {
         scopeManager,
         migrator,
         smartExtractor,
+        mdMirror,
         extractionRateLimiter,
         reflectionErrorStateBySession,
         reflectionDerivedBySession,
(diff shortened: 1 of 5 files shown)


rwmjhb (approved):
Approved on head `ff55900`. Orchestrator verdict: approve.

- npm ci --silent
- npm run build --if-present
- node --test test/smart-extractor-md-mirror.test.mjs
- npm test

All passed. The build left committed dist/source artifacts clean.

Non-blocking follow-ups: consider wiring `test/smart-extractor-md-mirror.test.mjs` into the standard npm/CI test entrypoints, and consider making merge mirroring use the returned `store.update()` entry so the markdown audit reflects the actually persisted row and skips notification when update returns null.

Issue #913, opened 2026-07-07
memory-lancedb-pro reports “no memories extracted” and skips regex fallback for sessions in locomo10.json


qingfeng-occ:
### Plugin Version

1.1.0-beta.11 (master)

### OpenClaw Version

2026.6.11

### Bug Description

We configured `memory-lancedb-pro` within OpenClaw and tested its accuracy using the `openclaw-eval` tool. However, very few memories were stored in LanceDB, and a failure to extract key information resulted in low accuracy.
The OpenClaw logs frequently show: 

```bash
`memory-lancedb-pro: smart-extractor: no memories extracted`
`19:35:57+08:00 info plugins {"subsystem":"plugins"} memory-lancedb-pro: smart extraction produced no candidates and no boundary texts for agent main; skipping regex fallback`
```

### Expected Behavior

Relevant conversation information should be automatically captured, processed, and persisted to LanceDB.

### Model: 
Qwen3-235B-A22B

### Conversation: 
The first conversation sample in `locomo10.json`, comprising 19 sessions.

### Configuration
```bash
"hooks": {
    "allowConversationAccess": true
},
"config": {
    "embedding": {
        "provider": "openai-compatible",
        "model": "qwen3-embedding-8b",
        "baseURL": "http://xx.xx.xx.xx:xxxxx/v1",
        "apiKey": "no-api-key-required",
        "dimensions": 4096
    },
    "autoCapture": true,
    "autoRecall": true,
    "enableManagementTools": true,
    "smartExtraction": true,
    "extractMinMessages": 1,
    "extractMaxChars": 8000,
    "sessionMemory": {
        "enabled": false
    },
    "sessionStrategy": "memoryReflection",
}

```
### RUN
```bash
uv run python eval.py ingest ./locomo10.json --sample 0 --token ***d28e162ab6d1dd2*** --base-url http://127.0.0.1:18789
uv run python eval.py qa ./locomo10.json --sample 0 --user eval-1 --output qa_results.txt --token ***d28e162ab6d1dd2***
uv run python judge.py qa_results.txt.1.jsonl --base-url https://****/v1 --token ***-8bb3-*** --model Qwen3-235B-A22B --output grades.json

### Error Logs / Screenshots

```shell

```

### Embedding Provider

None

qingfeng-occ:
@AliceLJY @gorkem2020 Could you please take a look at this issue?

gorkem2020:
[...]

PR #914, opened 2026-07-07
fix(smart-extractor): only learn noise from a genuine empty extraction; recover JSON from the reasoning field

gorkem2020:
## Problem

`extractCandidates` returns an empty list for three different reasons: the LLM/gateway call failed (empty or blank response), the response had an unexpected shape, or the model genuinely returned an empty `memories` list. The caller treats all three the same and feeds the conversation to the noise-prototype bank as a "nothing worth remembering" signal. So a transient model or gateway failure teaches the bank that real conversation content is noise, which then suppresses legitimate future extraction and recall. On a run where every call fails identically (see #913, Qwen3 served via a gateway that returns empty `content`), the bank fills with false prototypes and accuracy degrades.

## Changes

1. `extractCandidates` now returns a discriminated result (`ok` / `llm_failure` / `malformed`), and `learnAsNoise` is called only on a genuine parsed empty list. Failures and malformed responses no longer poison the noise bank.
2. When the model returns empty `content`, recover the JSON answer from the reasoning field before giving up. Checks both `reasoning_content` (OpenAI and DeepSeek convention) and `reasoning` (vLLM's Qwen3 parser) and runs it through the existing `extractJsonFromResponse`. This salvages the common case where a reasoning model in thinking mode routes its answer into the reasoning channel and leaves `content` empty.

## Tests

- New `smart-extractor-noise-gating` suite: null and malformed responses do not learn noise; a genuine empty list and validation-dropped candidates do.
- Extended the api-key client tests to cover reasoning recovery from both field names.
- Wired the new suite into the CI test manifest.

Refs #913.


rwmjhb (approved):
Approved on head `8ca3844`. Orchestrator verdict: approve. Note: adversarial R4 returned invalid JSON twice, so orchestrator confidence was reduced; I completed independent verification on the same head before approving.

Independent verification run:
- npm ci --silent
- npm run build --if-present
- node --test test/smart-extractor-noise-gating.test.mjs
- node --test test/llm-api-key-client.test.mjs
- npm test

All passed; build left committed dist/source artifacts clean.

Non-blocking follow-up: consider distinguishing a truly empty raw `memories: []` response from a non-empty raw `memories` array where every candidate is dropped by local validation. The latter still trains the noise bank today, which may be worth revisiting if malformed per-candidate output can represent real memory content rather than true noise.

PR #917, opened 2026-07-08
fix(inject): replace trust-implying preambles with factual provenance statements

gorkem2020:
## Problem

The injected context blocks (`<relevant-memories>`, `<inherited-rules>`, `<derived-focus>`) arrive embedded inside a user-role message with terse preambles that do not say where they came from. Capable agents, which are elsewhere instructed to treat user-message text skeptically, correctly notice that these look like formatting trying to pass as system metadata, flag them as suspected prompt injection, and in the worst observed case argue with the user about them or act against their own stored memories.

Adding "trust this" instructions to system prompts does not scale and creates its own injection surface. The alternative shipped here: each block SELF-IDENTIFIES ITS PROVENANCE factually, so an agent can reason its way to the right conclusion on its own.

## Changes

Each preamble now states, in one sentence, that the block was generated by the agent's own memory plugin from the agent's own stored memories or reflections, injected automatically at prompt build, and is not part of the user's message. The functional payload description is retained (for example, inherited rules remain described as stable long-term behavioral constraints with user-override precedence). No trust directives anywhere. Tag names are unchanged, so downstream tag-keyed stripping and matching are unaffected.

## Tests

Existing assertions on the preamble strings updated; the tag-strip regression was stash-verified as non-vacuous (disabling the strip regex makes it fail). Targeted suites green: extractor branches, reflection bypass (6), recall text cleanup (26), auto-recall timeout (5), self-improvement reset note. Build clean.

Live verification on a multi-agent gateway: payload traces after deploy show all three blocks rendering with the new provenance preambles exactly once per turn, and the previously observed "suspicious injected formatting" objection came from an agent reading the old bare-tag rendering.


gorkem2020:
Withdrawing this change after live testing surfaced a problem with the replacement wording itself.

The new preambles introduce an ownership claim ("your own stored reflections") that safety-trained models treat as an unverifiable provenance assertion. In controlled tests, Claude Sonnet based agents flagged and refused the blocks on their very first turn even when the content was entirely benign, specifically calling out that the material "is not something I generated." The agent has no episodic memory of authoring distilled reflections, so the claim is falsifiable from its perspective and trips injection-wariness training. The original untrusted-data framing makes no ownership claim and does not trigger this behavior.

We plan to propose a revised approach after further testing: scope-honest provenance without ownership claims, for example wording that states the entries were distilled by the memory system from the agent's prior closed sessions and predicts that the agent will not remember authoring them.

PR #916, opened 2026-07-08
fix(hooks): skip prompt injection for memory sub-session builds

gorkem2020:
## Problem

OpenClaw's active-memory extension runs a per-turn embedded recall sub-completion whose session key has the form `agent:<id>:main:active-memory:<hash>`. The host fires `before_prompt_build` hooks for that internal sub-build too, and this plugin's scope guard only skips `:subagent:` keys, so every prompt injector here re-enters inside the memory sub-build. Cost per turn: one wasted embedding call plus a vector search (and optional rerank), and the sub-build's injected block can then reach the MAIN model prompt through shared session messages. Observed live on a multi-agent gateway: the main prompt carried two distinct memory blocks on such turns (its own summary-mode block plus the sub-build's adaptive-mode block), i.e. genuinely redundant context and wasted input tokens.

## Changes

- New `isMemorySubsessionKey(sessionKey)` helper: a session key containing `:subagent:` OR `:active-memory:` marks an internal sub-build whose context comes from the parent.
- Applied at the four prompt-injecting `before_prompt_build` handlers (auto-recall, inherited-rules reflection injector, derived-focus injector, self-improvement bootstrap). Non-matching sessions are untouched.

## Tests

Six new cases in `test/memory-subsession-prompt-hooks.test.mjs`: an `:active-memory:` key skips auto-recall (retriever not invoked, no prependContext), a control key proceeds (retriever invoked), the `:subagent:` regression stays covered, the reflection injector skip is proven against seeded reflection data with a paired control asserting real `<inherited-rules>` content is otherwise reachable (avoiding a false-passing empty-store assertion), plus a self-improvement case. Related suites green (auto-recall timeout, reflection bypass, self-improvement).

Live verification: deployed on a multi-agent gateway. Before: two injection passes per turn with the block leak described above. After: exactly one injection pass per turn, and payload traces show each memory block delivered exactly once.

PR #920, opened 2026-07-09
fix: make external writes visible to long-lived table handles via readConsistencyInterval

gorkem2020:
## Problem

`MemoryStore.doInitialize()` calls `lancedb.connect(this.config.dbPath)` with no options. Per the LanceDB SDK's own type declarations (`node_modules/@lancedb/lancedb/dist/native.d.ts`):

> (For LanceDB OSS only): The interval, in seconds, at which to check for updates to the table from other processes. If None, then consistency is not checked. For performance reasons, this is the default. For strong consistency, set this to zero seconds. Then every read will check for updates from other processes. As a compromise, you can set this to a non-zero value for eventual consistency. If more than that interval has passed since the last check, then the table will be checked for updates. Note: this consistency only applies to read operations. Write operations are always consistent.

With no `readConsistencyInterval` set, a long-lived table handle (e.g. a gateway process holding the store open for its whole lifetime) never learns about writes committed by a different connection. This is reproducible without any real subprocess: two independent `lancedb.connect()` calls against the same `dbPath`, in the same Node process, exhibit the exact same pinning behavior, since the staleness is per-handle, not tied to an OS process boundary.

```
A (unset interval) sees before delete: 2
B (separate connection) deletes 1 row
A (unset interval) sees after B's delete: 2   <-- stale
```

vs. with `readConsistencyInterval: 0`:

```
A (interval=0) sees before delete: 2
B (separate connection) deletes 1 row
A (interval=0) sees after B's delete: 1       <-- fresh
```

Concretely, a CLI `delete-bulk` invocation (its own process, its own `MemoryStore`/table handle) does not get observed by a separately running gateway process until the gateway is restarted or performs its own write.

## Fix

- `StoreConfig.readConsistencyInterval` (`src/store.ts`): new optional knob passed straight through to `lancedb.connect()`'s options.
- Exposed as a plugin config knob: `storageMaintenance.readConsistencyIntervalSeconds` (`index.ts`, `openclaw.plugin.json`). Defaults to `0` (strong consistency: every read checks for updates).
- CLI: after a successful `delete`/`delete-bulk`, if the configured interval is non-zero, print one line noting a running gateway may take up to that many seconds to observe the change. Silent at the default (`0`), since there is nothing to warn about.

## Why default to `0` and not a non-zero interval

Measured per-read overhead on a local-disk LanceDB table (5,000 rows, 1536-dim vectors, 300 reads per config, 3 randomized-order trials):

| config | avg ms/read |
|---|---|
| unset (current default, no check) | 1.630 |
| `0` (strong consistency) | 1.540 |
| `5` (eventual, 5s) | 1.534 |

The differences are within noise: strong consistency costs no measurable overhead over the SDK's own unset default on local disk. Given that, there's no reason to accept any staleness window by default; `0` closes the gap completely for local-disk deployments. Object-storage backends (S3, GCS, etc.) will likely see a larger, non-trivial per-read cost from the freshness check over the network, so the interval remains a documented, tunable knob (`storageMaintenance.readConsistencyIntervalSeconds`) for those deployments to set a non-zero eventual-consistency window instead.

## Tests

`test/read-consistency-interval.test.mjs` (new):
- Positive: reader opened with `readConsistencyInterval: 0` observes a delete made through a separate connection to the same table.
- Negative control: reader opened with the interval unset does *not* observe that same delete (kept: it reproduced reliably in-process across repeated runs, so a real `child_process` writer wasn't needed).
- Getter and default-value coverage on `MemoryStore.readConsistencyInterval`.
- Spy on `lancedb.connect()` confirming the configured value is actually passed through in `doInitialize()`.

`test/storage-maintenance.test.mjs` (extended):
- `parsePluginConfig` parses `storageMaintenance.readConsistencyIntervalSeconds` correctly, including alongside `autoCleanup` and as an explicit `0` (not collapsed to "unset").
- `storageMaintenance` is omitted entirely when neither knob is configured (kept the existing exact-shape `deepEqual` assertion working for `autoCleanup`-only configs).
- Manifest (`openclaw.plugin.json`) schema default and UI hint presence.

Both files are registered in `package.json`'s `test` script and in `scripts/ci-test-manifest.mjs` (`storage-and-schema` group).

`npx tsc -p tsconfig.json` is clean. Full `npm test` was run command-by-command with `test/cjk-recursion-regression.test.mjs` excluded: that file crashes the whole `&&`-chained script with `EADDRINUSE: address already in use 127.0.0.1:11434` in this environment because a local Ollama instance already holds that port: reproduced identically against unmodified `origin/master`, so it's a pre-existing environmental issue, not something introduced by this change. Every other command in the test script (46 files, including the two new/extended ones) passed, and `node scripts/run-ci-tests.mjs --group storage-and-schema` passed in full (23 files, 0 failures).


rwmjhb:
This PR currently has merge conflicts with `master` (`mergeable=CONFLICTING`, `merge_state_status=DIRTY`). The orchestrator completed R0/R1, but deep review was deferred by the conflict gate because the branch cannot be merged as-is.

Please rebase or merge the latest `master`, resolve the conflicts, and push an updated head. I will re-run the review after the branch is clean.

gorkem2020:
Merged the latest master (including #914) into this branch to clear the post-merge conflicts. The only real conflict was a union merge of package.json's `test` script (scripts/ci-test-manifest.mjs auto-merged cleanly with both sides' entries intact, 79 total). Typecheck and all touched test files (read-consistency-interval, storage-maintenance, llm-api-key-client, smart-extractor-noise-gating) are green, and dist was rebuilt from a clean slate and verified to match the merged source byte for byte. This is ready for the deferred deep review now that the conflicts are clear.

app3apps:
Code review completed on head `b4da6cc`, and the orchestrator verdict was **approve** with no must-fix findings. However, that verdict does not mean the PR is currently mergeable.

After #916 was merged, `master` advanced to `9f9a97e`. This PR is now reported as `mergeable=CONFLICTING` and `merge_state_status=DIRTY`, so it cannot be merged as-is.

Please rebase onto or merge the latest `master`, resolve the new conflicts, and push an updated head. I will re-run the review after the branch is clean. Note that the previous run had reduced confidence (0.50) because R2a did not complete successfully.

PR #921, opened 2026-07-10
fix: insert blank line before numbered items in injected reflection blocks

gorkem2020:
## What Problem This Solves

The reflection prompt injection builder emits two blocks, `<inherited-rules>` and `<derived-focus>`, each of which pairs a header sentence with a numbered list of items. Both blocks joined the header and the first numbered item with a single newline. Plain text renders this fine, but some markdown viewers (for example trace UIs that render prompt context as markdown) require a blank line before a list for it to be recognized as a list at all. Without that blank line, the numbered items can collapse into a single run-on paragraph instead of a numbered list, and whether that happens is inconsistent across renderers.

## Why This Change Was Made

This is a formatting fix only. It inserts one blank line between the header sentence and the first numbered item in both blocks, so the numbered list renders as a list consistently everywhere it is viewed. No wording, tags, or item content changed, and the join between items themselves is untouched.

## User Impact

This is a cosmetic fix for anyone viewing these injected blocks in a markdown-rendering trace or debug UI. It has no effect on model comprehension: the underlying model reads the raw text either way, and a blank line between a header and a list does not change the meaning of the content. There is no behavior change for the plugin itself, only how the injected context looks when rendered as markdown.

## Evidence

- Added a regression assertion to the existing reflection hook test suite (`test/reflection-bypass-hook.test.mjs`) checking that both blocks contain a blank line between the header sentence and the first numbered item.
- Verified the new assertion fails against the pre-fix code and passes after the fix (stash-verified red before, green after).
- `tsc` build completes with no errors, and the compiled `dist/index.js` reflects the same two-line change.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -3453,6 +3453,7 @@ const memoryLanceDBProPlugin = {
                         prependContext: [
                             "<inherited-rules>",
                             "Stable rules inherited from memory-lancedb-pro reflections. Treat as long-term behavioral constraints unless user overrides.",
+                            "",
                             body,
                             "</inherited-rules>",
                         ].join("\n"),
@@ -3495,6 +3496,7 @@ const memoryLanceDBProPlugin = {
                                 blocks.push([
                                     "<derived-focus>",
                                     "Weighted recent derived execution deltas from reflection memory:",
+                                    "",
                                     ...derivedLines.slice(0, 6).map((line, i) => `${i + 1}. ${line}`),
                                     "</derived-focus>",
                                 ].join("\n"));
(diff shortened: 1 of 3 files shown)


app3apps (approved):
Reviewed head `3d62273`. The source and committed distribution changes are narrow and consistent, the regression assertions cover both intended reflection blocks, and the focused plus full test runs passed. I found no blocking issues.

The branch is behind `master`, but GitHub currently reports it as clean and mergeable. Please preserve the formatting assertions if the branch is updated before merge.

PR #923, opened 2026-07-10
fix: harden three scope and ownership leak paths in memory and reflection storage

gorkem2020:
## What Problem This Solves

Three separate paths let stored memory and reflection data cross agent boundaries that should never have been shared.

First, the general memory listing query included a clause that matched any row with no scope set at all, in addition to the requested scope. Any older row stored before scoping existed, or any row that ended up with a missing scope for another reason, was returned no matter which agent asked for which scope.

Second, the function that checks whether a reflection belongs to the requesting agent had two unconditional passes: a row with a blank owner was treated as belonging to everyone, and a row owned by "main" was treated as belonging to every requesting agent, not just the agent actually named "main".

Third, when the code that generates a reflection could not work out which agent a session belonged to from its session key, it silently treated the reflection as if it belonged to "main" and stored it that way. Combined with the second issue, this meant a reflection from a session that could not even be identified ended up visible to every agent, and even on its own it wrongly attributed content to the literal "main" agent.

## Why This Change Was Made

Each of these was an unconditional bypass of the ownership and scope checks that are supposed to keep one agent's memory and reflections separate from another's. None of them required a broken or malicious input, they happened on ordinary legacy or unidentified data. Closing them means scope and ownership checks now mean what they say: a row is visible only to the scope it was stored under, and a reflection is inheritable only by the agent that actually owns it.

## User Impact

After this change, rows with no scope set, and reflections with a blank or literal "main" owner, are no longer visible to other agents. If any such rows already exist from before these checks were tightened, they simply stop being returned, they are not deleted or altered. No migration is included as part of this change. If an operator wants a specific one of these older rows to remain visible or inheritable to a particular agent, it needs to be re-scoped or re-attributed to that agent explicitly. A parse failure on a session key that previously silently became a "main"-owned reflection is now stored without a usable owner (not editorializing the content, just no longer treating it as belonging to "main"), so it does not leak to any agent and is not wrongly attributed either.

One related item was intentionally left out of scope, the same missing-scope pattern also appears in a few other read paths that back the retrieval and recall pipeline (vector search, keyword search, a lexical fallback, stats, and a compaction read). Only the specific listing query described above was changed here, so the fix stays narrow and reviewable. The rest are worth a follow-up decision, since some of them may have relied on that fallback for a different reason and deserve their own look.

## Evidence

- Added a new test file covering all three fixes independently, plus regression tests confirming normal same-scope reads and exact-owner matches still work as before.
- Verified each of the five new failing assertions is red against the code before this change and green after it.
- tsc build completes with no errors, and the compiled dist files reflect the same change.

- Production verification on a live deployment: after deploying this change, reflection inheritance remained fully intact for all agents (every stored reflection row carries a real agentId, so the stricter exact-match ownership changes nothing for correctly attributed data), confirming the hardening only affects the leak shapes it targets.


app3apps:
I completed the review on head `62a9897`. The security direction is valuable, but one blocking completeness gap remains: the NULL-scope hardening changes only `MemoryStore.list()`; sibling read paths such as vector search, keyword search, lexical fallback, stats, and compaction remain unpatched or unverified.

Since #920 merged, GitHub now reports this branch as `mergeable=CONFLICTING` / `merge_state_status=DIRTY`.

Please rebase onto or merge the latest `master`, resolve the conflicts, and either harden the sibling read paths or demonstrate with production-path tests that they cannot expose NULL-scope rows across agents. Please also validate the compatibility impact of removing the legacy `owner === "main"` fallback. Push the updated head for re-review.

gorkem2020:
Fixed the sibling read paths: vectorSearch, bm25Search/lexicalFallbackSearch, stats, and fetchForCompaction all carried the same unpatched WHERE clause. stats and fetchForCompaction had no app-layer re-check at all, and the vectorSearch/bm25Search re-check was defeated whenever the caller's scope filter included "global", since NULL scope defaults to "global" before that check runs. Applied the same one-line fix as list() at all five sites, with four new tests confirmed red against the pre-fix code. On the main-fallback question: removing that fallback is the fix itself, not a side effect. It is covered by the existing regression suite (main keeps access to its own rows; other agents lose only the previously leaked access). Rebased onto latest master.

app3apps (changes requested):
Reviewed head `f6526cb`. The previous NULL-scope completeness finding is addressed across list, vector, BM25/lexical fallback, stats, and compaction paths, with regression coverage.

One blocking compatibility issue remains: the repository's existing `test/isOwnedByAgent.test.mjs` contract explicitly expects a `main`-owned invariant to be visible to a sub-agent and preserves the same fallback for legacy rows. This PR removes that behavior, but the contract is neither updated nor run by the active test manifest. That makes the change a concrete cross-agent inheritance regression, not only leak hardening.

Please either preserve intentional main/global invariant inheritance while closing unintended ownership leaks, or explicitly change the compatibility contract with active regression tests and a migration/release plan for existing rows. Also add the invariant ownership test to the normal CI manifest.

app3apps (changes requested):
Re-reviewed the rebased head `9a15e2b`. The hardening direction is valuable, but the current implementation still has blocking contract and isolation gaps:

1. `src/reflection-store.ts` now requires exact ownership, while the committed `test/isOwnedByAgent.test.mjs` contract explicitly requires main-owned invariants and legacy rows to remain visible to sub-agents. That test is not in the active manifests, so CI currently enforces only the new side of two contradictory contracts. Please define the intended sharing semantics, reconcile and activate the existing test, and provide migration/release behavior if the contract is intentionally changing.

2. NULL scope can still become or behave as global. `backfillLegacySecondTimestamps` rewrites qualifying rows with `scope: row.scope ?? "global"` during startup, and the exact-ID read/mutation/delete paths coerce raw NULL scope to `global` before authorization. Both behaviors were reproduced at this head and bypass the new list/search predicates. Authorization needs to use the raw stored scope, with restart and exact-ID regression coverage.

3. On an unparseable reflection session, only `ownerAgentId` becomes blank; `sourceAgentId` still falls back to `main`. As a result, target scope, mapped memories, reflection rows, and markdown mirroring can still route through the main agent. Please use one validated identity consistently or skip persistence/mirroring when attribution is unavailable, and test the real reflection hook rather than a local derivation helper.

The current targeted and full suites pass, but none of these paths is covered by them.

app3apps:

gorkem2020:
Rebased onto current master (03d6754); the two test-registration conflicts from the #919 merge are resolved, branch tests and the manifest check are green.

app3apps (changes requested):
Re-reviewed the rebased head `efbdf9f`. The rebase resolves conflicts, but it does not address the requested isolation and migration blockers.

1. The new predicates make existing NULL-scope rows disappear from scoped list/search/stats/compaction with no migration, while `upsert()` can still create NULL scope and other code still treats NULL as global. `backfillLegacySecondTimestamps()` can also convert a qualifying NULL row to global on restart, and exact-ID read/mutation/delete paths authorize NULL after coercing it to global. Please define one raw-scope authorization/migration contract and apply it consistently.
2. Exact ownership immediately revokes access to existing main-owned invariant/legacy rows and blank-owner rows, but the committed compatibility test still requires main invariant inheritance and remains outside active CI. Reconcile the contract and provide migration/reclassification behavior.
3. Parse failure still leaves `sourceAgentId = ... || "main"` for session recovery, target scope, mapped memory, reflection routing, filenames, and mirroring. Use one validated identity consistently or abort all generation/persistence/mirroring when attribution is unavailable, with a real-hook regression.
4. `fetchForCompaction(..., [])` is still unfiltered; explicit empty scope must deny all.

The existing tests pass because none of these paths is exercised.

app3apps (changes requested):
Re-reviewed rebased head `33ace1f`. The hardening direction is valuable and the exercised suites pass, but the ownership contract and upgrade behavior are still not safe to merge.

1. Exact-owner matching removes the existing `main`-owned invariant inheritance contract, while the committed `test/isOwnedByAgent.test.mjs` still asserts that contract and is not active in the test chain. There is no migration or reclassification for existing `main`-owned rows.
2. Removing `OR scope IS NULL` from every scoped read makes legacy NULL-scope rows disappear from list/search/stats/compaction with no recovery path.
3. Parse failure only blanks the persisted owner. `sourceAgentId` still falls back to `main`, and the reflection is still routed into a shared scope where scope-only reads can retrieve it.

Please define the intended ownership compatibility policy, reconcile and activate the contract test, and provide a migration/reclassification path for existing rows. Unparseable sessions should either abort persistence/mirroring or use an isolated placement that cannot be recalled through shared scope paths. Add exact-ID/raw-NULL authorization and real reflection-hook coverage as part of that fix.

Requesting changes.

app3apps (changes requested):
Re-reviewed head `e3af04c`. The exercised tests pass, but the ownership and upgrade behavior remain unsafe to merge.

Exact-owner matching removes the existing `main`-owned invariant inheritance contract, while the committed compatibility test still encodes that contract and is not active in the normal suite. Removing the NULL-scope fallback also strands legacy NULL-scope rows across list/search/stats/compaction without migration or reclassification. Separately, an unparseable session can still persist a reflection into shared `global`, where scope-only reads bypass the new owner predicate.

Please define the intended compatibility policy, reconcile and activate the owner-contract test, and provide migration/reclassification or an explicit compatibility flag for existing `main`/blank-owner and raw-NULL rows. When session attribution fails, abort persistence/mirroring or isolate the row so generic shared-scope reads cannot retrieve it. Add exact-ID, raw-NULL, and real reflection-hook regressions.

Requesting changes.

gorkem2020:
Addressed all three findings from the latest review in [109fcd1](https://github.com/gorkem2020/memory-lancedb-pro/commit/109fcd197ff38804f7c57fc314c031b54b05f51c):

1. `isOwnedByAgent`: the older `test/isOwnedByAgent.test.mjs` asserted the pre-hardening contract and was never wired into CI, so it silently contradicted this PR's own coverage. Updated its assertions to match the hardened exact-owner-match rule and registered it in CI.
2. NULL-scope leak: `getById`, `deleteExactId`, `update`, and `bulkUpdateExact` all coerced a NULL scope to `"global"` before the ACL check, so a scope filter that legitimately includes `"global"` still let a scopeless row through by ID. While fixing this I found the same pattern in a fifth method, plain `delete()` (distinct from `deleteExactId`), not called out in the review. All five now share one `isRowScopeAccessible()` check. Also stopped an unrelated timestamp-backfill pass from silently promoting NULL scope to `"global"` as a side effect.
3. Ownership leak on parse failure: the persisted row's DB scope still derived from a `"main"`-defaulting variable even though the ownership metadata was correctly blanked, so a plain scope-filtered read of `agent:main` could still surface an unattributed row. Routed parse failures to a dedicated scope no agent is ever granted instead.

All three are covered by new regression tests, verified red-before/green-after by reverting the fix and confirming failure (one of them was a real content-overwrite exploit, not just a visibility leak).

app3apps:
Thanks for the fixes in `109fcd1`. I reran the orchestrator against this head, but GitHub currently reports `mergeable=CONFLICTING` / `merge_state_status=DIRTY`, so the review stopped at the conflict gate after R0/R1.

Please rebase onto the latest `master`, resolve all conflicts, and push the clean result. The raw-NULL exact-ID checks, ownership-contract test activation, and unattributed reflection scope are recorded for full re-review on the mergeable head.

gorkem2020:
Rebased onto current master as requested, the branch is mergeable again. New head: [c7d2680](https://github.com/CortexReach/memory-lancedb-pro/pull/923/commits/c7d2680c2e352f73d386e0fafa53ace2e62572d9). Conflicts were limited to the test registration files (package.json test chain and scripts/ci-test-manifest.mjs), resolved by keeping both sides' entries; dist is rebuilt from the resolved sources and the PR's test suite plus typecheck pass on the new base.

app3apps (changes requested):
Re-reviewed the current head `c7d2680`. The full suite passes, and the earlier focused timeout was not reproducible: I reran `node test/update-consistency-lancedb.test.mjs` and all 6 tests passed. That timeout is not a blocker.

Two implementation issues still block this head:

1. `isOwnedByAgent` now removes the documented blank-owner and `main` inheritance fallback. Existing blank-owner rows become invisible to every agent, and existing `main` invariants stop reaching sub-agents, with no migration, reclassification tool, compatibility option, or release path. This may be the desired security contract, but it needs an explicit maintainer-approved transition and regression coverage for migrated data before silently changing persisted-data semantics.
2. `MemoryStore.upsert()` still accepts a raw NULL/missing scope while the hardened read and mutation paths now reject that row from every scoped operation. The write succeeds and immediately creates effectively invisible data. Please normalize the scope consistently with `store()/bulkStore()` or reject the write.

The fail-closed direction is good; these two storage-contract gaps need to be closed before merge.

app3apps (changes requested):
Re-reviewed head `44f4189` after the replay onto #924. The security direction is valuable and the exercised coverage is strong: the orchestrator's targeted and full suites passed, and I independently ran `npm run build` plus the 35 focused scope/ownership tests successfully.

One concrete storage-contract issue still blocks this revision:

`MemoryStore.upsert()` deletes the existing ID, then spreads the runtime entry and normalizes only metadata/importance. It still accepts `scope: null` or a missing scope; this PR's own tests rely on that behavior to create NULL-scope rows. Every newly hardened scoped read/update/delete path then denies the written row. A successful upsert can therefore replace a valid visible record with one the caller immediately cannot read, update, or delete through normal scoped APIs.

Please validate scope before deleting the old row, then either reject missing/NULL scope or normalize it according to one explicit write contract shared with `store()/bulkStore()`. Seed genuinely legacy NULL rows directly at the storage layer in the hardening tests, and add a regression proving an invalid upsert cannot strand or overwrite an existing record.

The intentional fail-closed visibility transition for legacy/main-owned reflection data is now documented; I am not repeating that policy choice as a separate blocker here. Requesting changes for the remaining write-boundary gap.

rwmjhb (changes requested):
Reviewed head `d34ae80`. The scoped-read and exact-owner hardening is valuable, and the targeted/full suites pass, but three remaining paths keep this security change from being mergeable.

1. `store()`, `bulkStore()`, and `upsert()` still accept missing or null runtime scopes. The hardened readers then make those successfully written rows inaccessible. More seriously, `upsert()` deletes the existing visible ID before inserting the unchecked replacement, so an invalid scope can destroy a valid row and replace it with orphaned data. Enforce one non-empty scope contract before any write/delete and add missing/null/blank tests; seed legacy NULL fixtures below the public API.

2. On an unparseable reflection session, database owner/scope attribution is isolated, but `sourceAgentId` still falls back to `main`. That fallback drives main-specific session recovery, reflection execution, filenames/event identity, and `mdMirror`, which writes mapped content into the main workspace. Resolve one validated identity and, when attribution fails, skip all agent-specific recovery, execution, persistence, and mirroring. Add a real command-hook test with mdMirror enabled.

3. The two new descriptors in `scripts/ci-test-manifest.mjs` are outside `CI_TEST_MANIFEST` and `getEntriesForGroup()`. `node --check` fails with `Unexpected token :`, so every grouped CI command importing the manifest is blocked even though the monolithic npm test passes. Move the entries into the manifest and run the manifest verifier plus grouped CI commands.

Also make `fetchForCompaction(..., [], ...)` explicit deny-all and provide a migration/reclassification path for legacy NULL-scope and blank-owner data that becomes invisible after this hardening. Requesting changes.

gorkem2020:
Head c8a0115 addresses all three findings and both riders:

1. Write-path scope contract: store() and upsert() now reject missing or blank scopes before any write or delete, with upsert() guarding BEFORE its delete so an invalid replacement can no longer destroy a visible row. bulkStore() filters scope-less entries under the same contract while still storing valid siblings. Tests cover missing, null, and blank scopes on all three, plus a regression proving the existing row survives a rejected upsert. Legacy NULL fixtures in the existing suite are now seeded below the public API.

2. Attribution: an unparseable reflection sessionKey no longer falls back to the main identity. When attribution fails, the command hook skips recovery, execution, persistence, and mirroring entirely and logs the skip. Added a real command-hook test with mdMirror enabled proving nothing is mirrored and no path resolves to the main agent.

3. Manifest: the two descriptors were dangling outside CI_TEST_MANIFEST after the last rebase (a rebase artifact on this branch only; the other open branches were swept and are clean). They are back inside the array; node --check passes and getEntriesForGroup imports cleanly (48 core-regression entries).

Riders: fetchForCompaction now treats an explicitly empty scope filter as deny-all like the other scoped readers, and a migration path exists for legacy data: findLegacyScopeRows plus repairLegacyScopes on the store, wired to a new repair-scopes CLI subcommand (report-only by default, --apply to reassign, target scope configurable). Covered by visibility-restoration tests. Full suite green.

rwmjhb (changes requested):
Re-reviewed head `c8a0115`. The source now addresses the previous write validation, unattributed reflection, CI manifest, and empty-compaction issues. Three new/current-head blockers remain.

1. The committed package runtime is stale. `package.json` executes `dist/index.js`, but it still falls back to `main` for unattributable reflections; `dist/src/store.js` lacks the final write guards and migration APIs, and `dist/cli.js` lacks `repair-scopes`. A clean TypeScript build changes all three files, and the committed compiled store still accepts a null-scope upsert that makes the replacement invisible. Rebuild and commit every dist artifact, then run package-runtime verification against the installed entrypoints.

2. Unrestricted `update()` and `bulkUpdateExact()` authorize a raw NULL-scope row, then persist it using `realScope ?? "global"`. `AccessTracker` exercises this unrestricted path automatically. A reproduction changed an invisible NULL row into a globally visible row after an update, recreating the cross-agent disclosure this PR is meant to close. Preserve the raw stored scope through unrestricted mutations and add both regressions.

3. `findLegacyScopeRows()` converts `row.timestamp` with `Number(...)`. BigInt values above `Number.MAX_SAFE_INTEGER` lose precision before `repairLegacyScopes()` rewrites the row. Use the established bounded/BigInt-safe timestamp normalization and add an above-safe-range migration test.

Please also make scope repair recoverable if replacement and rollback both fail, canonicalize scope whitespace, and report invalid `bulkStore()` entries instead of silently dropping them. Requesting changes.

rwmjhb (changes requested):
The latest head closes the earlier raw-NULL, BigInt, and source/dist gaps. The security direction is valuable, the full suite passes, and a clean build produces no `dist/` drift. Two data-integrity blockers remain:

1. `hasValidEntryScope()` validates `scope.trim()` but `upsert()` and `importEntry()` persist the original string. A padded value such as `"  agent:a  "` passes validation. In `upsert()`, the canonical visible row is deleted first and replaced by a row that `getById(..., ["agent:a"])` cannot see. Please normalize once before any destructive operation and persist the normalized scope, or reject non-canonical input.

2. `repairLegacyScopes()` discovers all legacy rows before taking the per-row write lock, then deletes by ID and writes the stale snapshot without re-reading. A concurrent unrestricted update can therefore be overwritten or reassigned to the repair target after it is no longer legacy. Please re-read under the cross-process lock, verify the current raw scope is still legacy, and skip/report rows that changed.

Focused padded-scope upsert/import tests and a two-writer repair interleaving test would lock these invariants down. Once those are addressed, the broader fail-closed hardening looks close.

gorkem2020:
Both blockers are addressed.

1. Scope canonicalization ([a9a8875](https://github.com/CortexReach/memory-lancedb-pro/pull/923/commits/a9a8875c14c0d2b75af40ff5903cb8c845403c23)): `upsert()` and `importEntry()` now persist the trimmed scope. `bulkStore` already canonicalized at the write boundary; these were the two paths that bypassed it. `importEntry()` maps blank and whitespace-only scopes to its documented `global` fallback. Tests cover a padded upsert staying visible to the canonical filter and padded/blank imports canonicalizing.

2. Locked re-validation (same commit): `repairLegacyScopes()` now re-reads each row under the write lock, repairs only rows whose stored scope is still legacy, builds the replacement from the re-read content, and skips rows that changed or were deleted after discovery. The result gains a `skipped` count, surfaced by the CLI report. Tests: a two-writer interleaving where the row gains a real scope after discovery (concurrent write preserved, row skipped), a still-legacy row whose content changed (repair persists the current content, not the snapshot), and a row deleted after discovery (not resurrected).

Also fixed in this round ([50f1927](https://github.com/CortexReach/memory-lancedb-pro/pull/923/commits/50f192707a4ba18c5a5e689b0b5afae50ff4e0e8)): `repair-scopes` was attached to the root program instead of the `memory-pro` group, so the migration path was unreachable through the host's routing (the same attachment class being fixed for `reindex-fts` and `repair-summaries` in [#944](https://github.com/CortexReach/memory-lancedb-pro/pull/944)); a reachability test now locks the wiring. The apply path also exits nonzero when rows failed or were unrecovered, so non-interactive callers cannot mistake a lossy migration for success.

rwmjhb (changes requested):
Re-reviewed head `e508221`. The read/write hardening, unattributed-session abort, dist rebuild, and canonicalization fixes are all valuable, and the focused/full suites pass. The migration path still has one data-integrity and isolation blocker.

`repairLegacyScopes()` takes the cross-process write lock and re-reads the candidate, but it does so through the existing LanceDB table handle without refreshing its snapshot. The lock serializes writes; it does not advance a stale handle. In an exact-head two-`MemoryStore` reproduction, another connection updated the row to text `new` and scope `agent:live`; repair then reported one repaired row, and a strongly consistent reader found stale text `old` reassigned to `global`. Please refresh/reopen the latest table after acquiring the lock and before the current-row validation. If refresh fails, skip/report the row rather than writing a stale snapshot. Add a two-handle regression with default or nonzero read-consistency interval.

Two migration-contract gaps should be closed with that fix: validate `--target-scope` through the normal scope validator so a typo cannot strand rows in an unreachable scope, and reconcile `importEntry()` with the new fail-closed write contract instead of silently converting missing/whitespace scope to globally visible data. Whitespace-only legacy scopes should also be discoverable by `repair-scopes`, since they are invalid under the public write validator but currently miss the NULL/exact-empty query.

Requesting changes for the stale locked re-read, which can both lose a concurrent update and recreate cross-agent exposure.

gorkem2020:
All three areas addressed, and our own pre-push review pass turned up adjacent gaps on the same paths, so those ship in this round as well.

**Stale locked re-read.** The locked closure now checks out the latest table version before the current-row validation. The lock serialized writers but never advanced the cached handle, exactly as you described. A failed refresh fails that row into the reported counters instead of ever validating against a stale snapshot. The regression is the two-handle shape you asked for: a second store connection updates the row's text and scope under default read consistency, and the repair pass must skip it and preserve the newer content; without the fix the test reproduces your overwrite.

**--target-scope validation.** The command now routes the value through the scope manager's validator before touching the store. Since format validation alone still accepts the exact typo shape (agent:mian matches the agent pattern), a well-formed scope the config has never defined additionally requires an explicit `--allow-new-scope`, with the failure mode spelled out in the error. Typos exit nonzero before discovery runs.

**importEntry.** No longer coerces a missing or whitespace scope to global; it rejects like store() does, and migrate resolves blank legacy scopes to the migration default before calling it. The old test pinning the coercion now pins the rejection. Whitespace-only scopes are also discoverable by repair-scopes now; Lance's SQL pushdown rejects trim(), so discovery projects id and scope and filters client-side, materializing the rare hits individually, and the under-lock guard applies the same trimmed test.

Adjacent fixes from our own review of this round:

- reembed pre-coerced blank source scopes to "global" before importEntry ever saw them, which would have mass-promoted invisible legacy rows into globally readable data through a side door. It now skips those rows with a per-row pointer to repair-scopes and a separate summary count.
- repairLegacyScopes reported failures as a bare count. Per-row reasons are now logged (bounded), and ten consecutive failures abort the pass as systemic instead of re-acquiring the cross-process lock for up to 100k rows; unattempted rows stay legacy and rediscoverable.
- The failed-add path re-reads the row before rolling back, so a commit-then-reject write that actually landed counts as repaired instead of being doubled into two rows under one id by a blind rollback re-add.
- import-markdown reports what bulkStore persisted rather than what was attempted, and an unreachable quarantine arm in the reflection scope routing is removed (the unattributable-sessionKey early return is the quarantine).

Full suite green; the hardening file is at 49 tests including the two-handle, whitespace-discovery, circuit-breaker, commit-then-reject, and novelty-gate regressions.

rwmjhb (changes requested):
Re-reviewed new head `b657d76`. The locked table refresh, target-scope validation, fail-closed import/re-embed behavior, whitespace discovery, and ambiguous replacement recovery address the previous migration blockers; the focused/full suites pass. One remaining read-side ownership bypass still conflicts with this PR's isolation contract.

`resolveHookAgentId()` still returns `"main"` whenever `ctx.agentId` is absent and `sessionKey` cannot be parsed. Both reflection `before_prompt_build` hooks use that synthesized identity, resolve main's scopes, and call `loadAgentReflectionSlices("main", ...)`. An unattributable session can therefore receive main's private invariant and derived reflection content even though the write/command path now correctly aborts all agent-specific work for the same identity failure. This is the read-side version of the fallback this PR is intended to remove.

Please make hook identity resolution return `null` when neither an explicit validated agent nor a parseable session key exists, and skip both reflection injection hooks before any scope or slice lookup. Add production-hook regressions for empty and malformed session keys with no `ctx.agentId`, covering both inheritance-only and inheritance+derived modes and asserting that no main content is loaded or prepended.

The rollback commit-then-reject false `unrecovered` report and full-row stderr output are useful follow-ups, but they are not the blocker. Requesting changes for the remaining unattributed-session read leak.

gorkem2020:
Fixed in [8073d06](https://github.com/CortexReach/memory-lancedb-pro/pull/923/commits/8073d06).

**Hook identity now returns null.** `resolveHookAgentId()` no longer falls back to `"main"`: it returns the explicit agent id, else the id parsed out of the session key, else null. That fallback was invisible to the existing validation, because `main` is a declared agent, so the synthesized identity passed `isInvalidAgentIdFormat()` and went straight on to resolve main's scopes. All seven call sites short-circuit on null before any scope or slice lookup, which covers both reflection `before_prompt_build` hooks and, with the same resolver, auto-recall and auto-capture. The `before_reset` error path had no validation at all and was building a scope-keyed guard entry from the synthesized owner; it now skips that bookkeeping when there is no identity.

**Regressions drive the registered production hooks**, not a logic mirror: six unattributable session shapes (absent, empty, whitespace-only, non-agent-prefixed, `agent::` with an empty id, and a blank explicit `agentId`) across both `inheritance-only` and `inheritance+derived`, asserting that nothing is prepended and that neither main's invariant nor its derived slice text appears. Twelve of those cells failed before the fix, each one serving main's actual invariant. Two attributed-main sanity cells run alongside them so the assertions cannot pass vacuously. Registered in the npm chain and the CI manifest (`storage-and-schema`).

**One scope limit worth stating plainly rather than implying closure.** On the deployed host build, the plugin's hook and tool contexts arrive with an already-resolved `agentId`: the host computes it as `explicitAgentId ?? (parsedFromSessionKey ?? defaultAgentId)`, so all six shapes above collapse to the configured default agent before the plugin runs. This fix therefore closes the synthesis the plugin itself performed, and is defensive with respect to any caller that omits the field, but a host that has already defaulted an unattributable session to the default agent is indistinguishable at the plugin layer from one that resolved it legitimately. Fully closing that class needs the host to distinguish resolved-from-key from defaulted, which is outside this repo. We mention it so the isolation contract is not read as stronger than it is.

**One nit our own pre-push pass found in the fix.** The priority-15 hook assembles two independent blocks, and the null-identity return now suppresses both: the derived-focus block, which is genuinely agent-specific, and the `<error-detected>` block, which is keyed only by `sessionKey` and carries nothing but that session's own tool-error summaries. The producing `after_tool_call` hook still records those signals for any non-internal session key, so for an identity-less session they are now recorded and never delivered, and simply age out. There are two defensible resolutions: gate only the slice-loading branch and let the session-keyed error block still assemble, or stop recording signals for session classes that can never read them so the two sides agree. The block's only payload is an advisory nudge toward a per-agent workspace file that an identity-less session has no owner for, which is part of why we did not pick unilaterally. Happy to implement either in the next round.

The rollback commit-then-reject `unrecovered` report and the full-row stderr output are not folded into this round, so they stay open as you noted.

Full suite, typecheck, build, and the `storage-and-schema` and `cli-smoke` groups are green on the new head.

rwmjhb (approved):
Re-reviewed head `8073d06`. The remaining unattributed-session read leak is fixed: unresolved hook identity now returns `null`, and the reflection injection paths stop before resolving scopes or loading main-owned slices. The new production-hook regressions cover empty and malformed session keys, and both the focused tests and full suite pass.

The broader fail-closed scope/ownership hardening and migration work now address the prior blockers. There are still worthwhile follow-ups around warning operators about legacy NULL-scope rows, validating full scope shapes on every write/repair path, avoiding full-row stderr recovery output, and making `repair-scopes` more crash-safe, but none blocks this security fix. Approving.

PR #926, opened 2026-07-11
fix: unblock manual recall metadata patching

### LINKED ISSUE #925: memory_recall blocks on best-effort access metadata writes
## Summary

Manual `memory_recall` waits for one access-metadata write per returned result before returning retrieval output that is already available. Those writes acquire the shared store write lock, so lock contention or an index fold can make an otherwise successful recall exceed the host tool timeout.

Auto-recall had the same problem and moved Tier 1 metadata patching off its critical path in #806. Manual recall still awaits `Promise.allSettled(...patchMetadata())` in `src/tools.ts`.

## Expected behavior

Once retrieval succeeds, `memory_recall` should return its results without waiting for best-effort access telemetry. The metadata patches should still start in the background, preserve the existing governance fields and scope filter, and report failures without changing the successful recall result.

## Reproduction

1. Use current `master` with an active LanceDB store.
2. Create write contention through concurrent capture/update activity or an index fold.
3. Run manual `memory_recall` with multiple results.
4. Observe that the tool can remain pending on `patchMetadata()` after retrieval has completed.

A deterministic unit reproduction is to make `store.patchMetadata()` return an unresolved promise. The current manual recall tool does not return until that promise settles; the equivalent auto-recall path already returns context before its background metadata patches settle.

## Observed evidence

- OpenClaw: `2026.6.11`
- Plugin: current `master` at `7af97d4`
- Node.js: `24.18.0`
- Platform: macOS arm64
- Retrieval-only debug pipeline: under 1 second
- Manual recall: hit OpenClaw's 90-second dynamic-tool watchdog during contention
- A recent write-threshold index fold on the same store took about 37 seconds
- A single-result manual recall completed normally after contention cleared

The file-lock retry budget can exceed the host's 90-second tool watchdog, so increasing the tool timeout would only mask the coupling.

## Suggested fix

Match #806: run the existing manual-recall metadata patches as best-effort background work with rejection logging, while preserving the exact metadata patch and resolved `scopeFilter`. Add regression coverage proving stalled metadata writes do not delay the recall payload and rejected background patches remain observable.

Index folding from #887 is still useful and should remain enabled; the problem is optional metadata writes being part of the user-facing read path.

TurboTheTurtle:
## Summary

- return manual `memory_recall` results before best-effort access-metadata patches settle
- preserve the existing governance patch and resolved scope filter
- log rejected background patches without changing a successful recall result
- cover stalled and rejected metadata writes with focused regression tests

Fixes #925

## Verification

- `node --test test/recall-text-cleanup.test.mjs test/auto-recall-timeout.test.mjs test/memory-governance-tools.test.mjs` — 41 passed
- `git diff --check`


app3apps (approved):
Reviewed head `f547fc5`. Moving best-effort access metadata patches off the manual-recall response path directly addresses the reported timeout, matches the established auto-recall pattern, and the full test suite passes. I found no blocking issue.

Non-blocking follow-ups:

- The detached patch also resets suppression/governance fields, not only telemetry, so the code comment should describe the broader best-effort semantics.
- Deferral widens the read-modify-write window for `access_count`, which can undercount closely sequenced recalls.

The branch is behind current `master`, but GitHub currently reports it as clean and mergeable.

Issue #912, opened 2026-07-06
Feature request: speaker-aware extraction and a speaker field on memories (multi-participant channels collapse every speaker into "User")


gorkem2020:
## Problem

In multi-participant channels (several humans, or humans plus other assistant agents sharing a room), every inbound message reaches the agent as a user-role turn. Smart extraction then attributes everything it learns to "the user": the extraction prompt receives the conversation as one undifferentiated text blob with a single `User: <name>` header, and the prompt's noise rules explicitly list sender IDs among things to never extract (`src/extraction-prompts.ts`). The memory schema has no speaker concept either: `MemoryEntry` carries `scope` and the metadata carries `source_session`, which identify the lane a memory came from, but nothing records who actually said it.

The result on a live multi-agent install: memories like "User prefers X" where X was stated by a peer assistant, durably stored, then recalled and injected as the operator's own preference. Once a wrong fact is admitted it keeps getting confirmed by recall. The research literature calls this provenance collapse; among popular frameworks the transcript layer usually keeps speaker identity (AutoGen `source`, AG2 `name`, CrewAI `TaskOutput.agent`) while the memory layer loses it, with Mem0's per-message `name`/`user_id`/`agent_id` being the notable exception.

## What just changed on the host side

OpenClaw merged durable sender attribution today: [openclaw#90552](https://github.com/openclaw/openclaw/pull/90552) persists `senderId`, `senderName`, and `senderUsername` inside the `message.__openclaw` envelope of durable user-turn transcript rows for group and channel routes (with [openclaw#98336](https://github.com/openclaw/openclaw/pull/98336) as a first in-core consumer). So the reliable speaker source the memory layer needs is now landing upstream of this plugin.

## Proposal

1. **Speaker-aware extraction input**: when building the extraction window, pass per-message speaker labels (from `message.__openclaw.senderName`/`senderId` when present) instead of one anonymous blob, and drop the never-extract-sender instruction in favor of attribute-per-speaker guidance.
2. **A speaker/subject field on memories**: record who a fact is about and who stated it (e.g. `speaker` in `SmartMemoryMetadata`, plus speaker-namespaced `fact_key`s), so "sam.rivera prefers X" and "the operator prefers X" are distinct facts that no longer collide or overwrite each other.
3. **Attribution-aware recall phrasing**: injected memories say who the fact belongs to, so downstream models stop reading peer facts as operator facts.

[...]

PR #927, opened 2026-07-12
fix: tag extraction candidates with conversational grounding and drop constructed content

gorkem2020:
## Problem

Extraction has no notion of conversational register. An in-game rule, a persona's claim, or drafted fiction is textually indistinguishable from a durable user fact once it reaches the extraction LLM, so a well-formed invented rule gets extracted with exactly the same shape as a genuine preference. Downstream, anything that scores candidates by shape rather than reality (for example a utility/confidence-style admission heuristic) will tend to prefer the invented content, because in-fiction text is maximally grounded in the transcript and reads as durably useful.

The fix belongs at the source: the extractor should know, per candidate, whether the underlying assertion is actually about the real user/world or only true inside an in-conversation constructed context (a game, roleplay, drafted fiction, a hypothetical, or sample/test data being manipulated).

## The grounding mechanism

`buildExtractionPrompt` (`src/extraction-prompts.ts`) now asks the model to tag every extracted memory with a `grounding` field:

- `"real"`: an assertion about the actual user, the real world, or this real working session, including a genuine first-person aside stated in passing during a game (for example, mentioning a real appointment while playing).
- `"constructed"`: an assertion whose truth exists only inside an in-conversation constructed context: a game's rules/scores/bets, a persona's claims, drafted fiction, a hypothetical, or sample data.

A deterministic post-filter in `SmartExtractor.extractCandidates`'s validation loop (`src/smart-extractor.ts`) then enforces the contract:

- A candidate tagged `"constructed"` is dropped unless its category is `events`.
- At most one constructed `events` note survives per extraction (an episodic record that the real participants did the activity, e.g. "ran a puzzle exercise", not the invented rules/scores themselves).
- A candidate tagged `"real"` is never affected by this filter, regardless of category, including a real aside stated during play.
- A missing, non-string, or unrecognized `grounding` value fails open to `"real"`, so a model that ignores the new field cannot regress extraction behavior. This keeps the change fully backward compatible.

## The scope-glob extraction-policy knob

As a cheap, deterministic companion, `SmartExtractorConfig.extractionPolicy` accepts a map from a scope string (or glob, using `*`) to one of:

- `"full"` (default): today's behavior, unchanged.
- `"episodic-only"`: only `events`-class candidates are kept, independent of grounding. Useful as a blunt backstop for scopes known to be pure play/roleplay.
- `"none"`: extraction is skipped entirely for that scope, with zero LLM calls.

Resolution (`resolveExtractionPolicy`, exported from `src/smart-extractor.ts`) prefers an exact scope match over a glob match, and defaults unmatched scopes to `"full"`. This knob needs no host-side plumbing: it is read entirely from `SmartExtractorConfig`, keyed on scope strings the extractor already has in hand.

## Test coverage

New suite: `test/extraction-grounding-register.test.mjs` (wired into `package.json`'s `test` script and into `scripts/ci-test-manifest.mjs` under `core-regression`).

- Constructed game content (an invented house rule, a bet outcome) is dropped from `preferences`/`cases`, keeping only the single constructed `events` note.
- Two constructed `events` candidates in one extraction: only the first survives (cap enforcement).
- A genuine real first-person aside stated during play survives under its natural category, alongside the constructed episodic note (false-positive guard).
- A purely factual transcript is extracted unchanged (regression guard against the default/no-`grounding`-field path).
- A malformed or missing `grounding` value (non-string, unrecognized string) fails open to `"real"` and is kept.
- `buildExtractionPrompt` documents the grounding contract (structural check against prompt regressions).
- `resolveExtractionPolicy`: exact match wins over glob match; unmatched scope defaults to `"full"`.
- Scope mapped `"none"`: extraction is skipped with zero LLM calls.
- Scope mapped `"episodic-only"`: only `events`-class candidates survive, independent of grounding.
- Unmatched scope: behavior is unchanged from `"full"`.

All fixtures are synthetic two-agent transcripts (`agent-one`/`agent-two`) with invented game content.

## Non-goals

- The admission-control gate (`src/admission-control.ts`) is untouched. A `reality`-style admission feature is a reasonable independent defense-in-depth follow-up, but it is downstream of this fix, costs a second per-candidate signal, and only helps deployments that enable admission control (off by default). Better evaluated after this change has had time to prove out.
- Speaker attribution (distinguishing the real user from an agent labeled "User" in-transcript) is a separate, independently tracked fix. `grounding` and speaker attribution are orthogonal per-candidate filters in the same validation loop and are designed to stack without either blocking the other.
- Host-side config wiring to expose `extractionPolicy` through the plugin's own configuration file is left as a fast-follow; the knob is fully functional today via `SmartExtractorConfig` for any caller that constructs `SmartExtractor` directly.

## Live fleet verification (2026-07-12)

Deployed this branch to a live gateway and probed it with a fresh two-turn webchat session against a fleet test agent:

1. Turn 1: a one-round roleplay message carrying five extractable-looking in-game "facts" (a favorite drink, a residence, a pet, a persona identity, and a house rule).
2. Turn 2: game over, plus a genuine first-person aside ("doing a routine memory-system check this Sunday morning").

Observed (extraction model: openai/gpt-oss-120b):

- Turn 1 raw extractor output contained exactly one candidate, the allowed episodic `events` note about the roleplay activity itself; the model's reasoning applied the new grounding rules ("...seems like part of the roleplay persona. So it's constructed."). None of the five in-game propositions surfaced as `profile`/`preferences`/`entities`/`cases`/`patterns` candidates.
- Turn 2 extracted the real aside as a real `events` candidate, showing the rule does not over-suppress genuine asides spoken around play.
- Downstream admission scoring then rejected both episodic notes on utility grounds, so the net store delta was zero rows.
- The deterministic post-filter was not exercised in this run (the prompt level already suppressed everything durable) and remains in place as the backstop for weaker extraction models.

For contrast: before this change, an equivalent game session on the same fleet persisted in-game rules, scores, and persona claims as durable `preference` and `fact` rows.

## v2 hardening (added after extended live testing)

Repeated runs of the same roleplay fixture surfaced a failure mode in v1: grounding self-tagging is not deterministic. On one of three identical-fixture runs, the model tagged three in-fiction canon items `real` (matching the shape of the few-shot examples, which all showed the same three-real-plus-one-constructed pattern) while correctly tagging the session note `constructed`, a batch self-contradiction the deterministic filter could not catch because it trusts the per-item tags and fails open to `real`.

v2 therefore stops relying on per-item self-tags alone:

1. **Batch-level register**: the extraction schema gains a top-level `conversation_register: "real" | "mixed" | "fiction"`, judged once per batch (whole-session classification proved stable across runs where per-item tags wobbled). The filter enforces it deterministically: `fiction` drops all durable candidates regardless of per-item tags and keeps at most the single constructed events note; a missing register is treated as `mixed` (toward scrutiny, not open).
2. **Batch contradiction check**: in a non-real register, an explicit `constructed` tag in the batch demotes surviving real-tagged durables (deterministic, no extra LLM call). Tradeoff, stated plainly: a genuine durable aside made mid-fiction without an exit cue is sacrificed; such asides normally classify as events, and the demotion is audit-logged.
3. **Grounding reaches admission and storage**: candidates now carry `grounding` (and the batch register) past the filter into the admission evaluation and into persisted row metadata, so there is an audit trail of which stored events used the constructed allowance. Constructed candidates targeting durable registers are auto-rejected before any admission LLM call.
4. **Fiction-aware admission**: the utility prompt receives the grounding and register lines and scores in-fiction claims near zero for durable categories; the durable-category type prior is capped to the events level under a fiction register so priors cannot outvote reality.
5. **Prompt rework**: the grounding rule is now explicitly per item with no expected tag counts ("the storage cap is a storage rule applied after tagging, it is NOT a tagging quota"), with a batch self-consistency instruction, and the few-shot examples were reshaped into varied batch shapes including an all-constructed mid-game example.
6. Vocabulary alignment: the admission guidance names all six registers correctly, and the extraction criteria bullets use only the output enum.

## v2 live verification

Deployed on a live gateway and re-probed with the original fixture: extraction produced exactly one real aside plus one constructed session note (both with persisted grounding and register metadata), zero fiction durables, an honest "I don't have that information" on the recall probe that previously returned the fictional answer, and an admission rejection whose logged reasoning explicitly cited the fictional frame. The deterministic layers are covered by red-proven unit tests reproducing the exact observed failure shape.


app3apps:
This PR currently conflicts with the latest `master` (`mergeable=CONFLICTING`, `merge_state_status=DIRTY`). R0 verification and R1 value assessment completed, but the conflict gate deferred deep review because the branch cannot be merged as-is.

Please rebase onto or merge the latest `master`, resolve the conflicts, and push an updated head. I will run the full review after the branch is clean.

One item to keep in scope during the rebase: R1 found that the new `extractionPolicy` scope-glob feature is broader than the grounding bug fix and is exposed without normal plugin-configuration wiring. The full review will evaluate whether it should be split or fully integrated.

app3apps (changes requested):
Re-reviewed the rebased head `e993727`.

There is one functional regression to address before merge:

- In `src/smart-extractor.ts:971-980`, once a non-real batch contains any constructed candidate, the contradiction guard removes every surviving durable candidate without checking that candidate's own grounding. A mixed conversation containing a real-grounded preference plus a constructed event therefore loses the genuine preference. That conflicts with the prompt contract that real asides in mixed conversations should survive under their natural durable category.

Please preserve real-grounded durable candidates and add a regression covering a mixed batch with both real and constructed candidates.

Verification note: the new 21-test grounding suite passes in isolation. The serial full suite hit the orchestrator's 180-second timeout before reaching that file; no assertion failure was observed. As a separate follow-up, `extractionPolicy` values should be validated because an unknown value currently fails open to full extraction.

app3apps:

app3apps (changes requested):
Re-reviewed the rebased head `4d16f21`. The rebase resolves the merge conflict, but the requested functional change is still outstanding.

`src/smart-extractor.ts:971-980` removes every durable candidate whenever a non-real batch contains any constructed sibling, without checking the surviving candidate's own grounding. An end-to-end mixed batch containing a real-grounded preference and a constructed event persisted only the event. The prompt contract says genuine real asides should retain their natural category, while `test/extraction-grounding-register.test.mjs:389-405` currently locks in the opposite behavior by expecting the real profile candidate to be removed.

The fiction-register one-event cap has the same grounding-independent ordering problem: a constructed game event appearing before a real out-of-character event wins the single slot.

Please preserve real-grounded durable candidates in mixed batches, make the event cap grounding-aware, and update the regressions to cover both cases. The full suite passes, but its current expectations encode the data-loss behavior rather than catch it.

app3apps (changes requested):
Re-reviewed rebased head `fcd0e16`. The rebase is clean and CI is green, but the previously requested mixed-register data-loss fix is still absent.

The batch contradiction path removes every surviving durable candidate when a non-real batch contains any constructed candidate. The regression named `register 'mixed' with a constructed sibling demotes real-tagged durables` explicitly locks in that behavior: a genuine, independently grounded real-world durable fact is discarded solely because a constructed sibling appears in the same batch. That contradicts the PR's own per-candidate grounding contract.

Please preserve independently real-grounded durable candidates, apply contradiction/admission decisions per candidate, and make any event cap grounding-aware rather than order-dependent. Add an end-to-end mixed-batch case proving the real durable reaches storage while the constructed sibling does not.

Requesting changes.

app3apps (changes requested):
Re-reviewed head `03d8dea`. The branch is clean and CI/full tests pass, but the mixed-register data-loss blocker remains.

The contradiction pass still removes every surviving durable candidate when a non-real batch contains any constructed sibling; it never checks the durable candidate's own `grounding`. Missing or unrecognized `conversation_register` values also default to `mixed`, which arms the same batch-wide wipe in ordinary extraction failures. The test that expects a real-grounded profile candidate to be demoted is pinning the incorrect behavior rather than protecting the per-candidate grounding contract.

Please preserve independently real-grounded durable facts, apply rejection/admission per candidate, and make the constructed-event exception grounding-aware rather than category-only or order-dependent. Add end-to-end cases for a genuine real aside plus a constructed sibling, and for a missing/malformed register, proving the genuine durable reaches storage.

Requesting changes.

gorkem2020:
Checked all four findings against the actual code and this PR's own "v2 hardening" section. All four describe the same accepted trade-off, not a regression:

The PR body states plainly: "a genuine durable aside made mid-fiction without an exit cue is sacrificed" and "a missing register is treated as mixed (toward scrutiny, not open)." The batch-level contradiction check and the one-event cap are deterministic, no-extra-LLM-call filters that intentionally override per-candidate grounding for durable categories, and the existing test `test/extraction-grounding-register.test.mjs:389-405` pins exactly this with an inline comment ("tag wobble: still capped by the register rule"), not an accidental pin. I confirmed the code matches the documented intent in each case, not an oversight.

One narrow, concrete gap in the findings was real though: there was no test for the compound case of a missing/malformed `conversation_register` combined with a genuinely constructed sibling in the same batch. Added in [5a2693e](https://github.com/gorkem2020/memory-lancedb-pro/commit/5a2693e10bb3fc1b1da2618f49fac7b4214ed472), confirmed it catches a real regression by temporarily changing the missing-register default and watching it fail.

Whether the batch-level heuristic should eventually become per-candidate grounding-aware (at the cost of an extra LLM call or a smarter check) is a product question the PR body already raises and accepts the cost of, not something to change in this content-frozen branch.

app3apps:
Thanks for the additional regression in `5a2693e`. I reran the orchestrator against this head, but GitHub currently reports `mergeable=CONFLICTING` / `merge_state_status=DIRTY`, so the review stopped at the conflict gate after R0/R1.

Please rebase onto the latest `master`, resolve all conflicts, and push the clean result. The mixed-register behavior and the new missing-register compound case remain queued for full review against the actual merge result.

gorkem2020:
Rebased onto current master as requested, the branch is mergeable again. New head: [67928fd](https://github.com/CortexReach/memory-lancedb-pro/pull/927/commits/67928fde01b6ef87ebeaa6e8ddf5410dde5f870f). The PR content is unchanged by the rebase (verified by comparing the full source diff before and after); conflicts were the test registration files plus one import-list collision with the recently merged category map change, and dist is rebuilt from the resolved sources. The PR's test suite and typecheck pass on the new base.

app3apps (changes requested):
Re-reviewed the current head `67928fd`. The grounding direction is valuable and the full suite passes, but the central mixed-register blocker remains.

In `SmartExtractor.extractCandidates`, any non-`real` register plus one `constructed` sibling removes every candidate in `DURABLE_CATEGORIES`, including candidates whose own grounding is explicitly `real`. That contradicts the prompt contract that a genuine out-of-character aside in a constructed conversation should be retained under its natural category. A mixed batch is precisely where real-grounded profile, preference, entity, case, or pattern facts can coexist with constructed siblings, so this currently causes deterministic silent data loss.

Please preserve independently `real`-grounded durable candidates in mixed batches (or define another policy that does not override correct per-item grounding), and add a regression with a constructed sibling plus real-grounded durable candidates. The missing/malformed `conversation_register` fallback also normalizes to `mixed` and activates the same wipe, so that compatibility path needs an explicit safe behavior rather than a test that pins the loss.

app3apps:
Since #924 merged, GitHub now reports this branch as conflicting with the latest `master`. Please rebase onto or merge the current `master`, resolve the conflicts, and push the updated head. The prior review blockers remain recorded and will be re-reviewed on the clean revision.

gorkem2020:
Rebased onto the current master. This push also includes a dist/index.js rebuild: a fresh build on the rebased tree differs from the previously committed artifact, and the packaging freshness gate requires the committed dist to match the build output.

Update: the branch now carries the v3 grounding refinement, field-tested on a live fleet this week: grounding distinguishes claims ABOUT a fiction session (real, storable as episodic events) from claims true only WITHIN the fiction (never stored), and the batch-level register deterministically overrides per-item tag wobble for durable categories. The register regression suite was updated to the v3 contract in the same push.

app3apps (changes requested):
Re-reviewed head `8a4b8cb` after the v3 grounding changes. The build and the focused `extraction-grounding-register` suite pass, and CI is green.

The original blocker remains in `src/smart-extractor.ts`: when `conversationRegister !== "real"` and any sibling is tagged `constructed`, the batch-level contradiction pass removes every surviving durable category without checking that survivor's own `grounding`. This still drops independently real-grounded profile, preference, entity, case, or pattern facts. A missing or malformed register is normalized to `mixed` and arms the same wipe. The current regression tests explicitly codify this data loss.

Please preserve independently real-grounded durable candidates (or make the contradiction check candidate-specific), and add end-to-end coverage for mixed plus missing/malformed-register batches.

app3apps (changes requested):
Re-reviewed head `54f2784` after the latest rebase. The source/test diff from the previously reviewed head contains no grounding-policy fix; it only picks up the merged admission-controller accessors. The build, full test suite, 23 focused grounding tests, and GitHub CI all pass.

The requested data-integrity fix is still outstanding. In `src/smart-extractor.ts:970-979`, any non-`real` register plus one `constructed` sibling removes every surviving `DURABLE_CATEGORIES` candidate without checking that candidate's explicit `grounding: "real"`. Missing or malformed registers normalize to `mixed` and take the same path. The tests named `register 'mixed' with a constructed sibling demotes real-tagged durables` and `missing register ... arms the batch-wide durable wipe` explicitly pin this silent loss rather than prevent it.

The fiction-register branch at `src/smart-extractor.ts:942-947` is broader still: it drops an explicitly real-grounded durable aside without even requiring a constructed sibling, contrary to the prompt contract that genuine real-world asides retain their natural category.

Please preserve independently real-grounded durable candidates, make the register/contradiction decision candidate-specific, and add end-to-end persistence coverage for mixed, fiction, and missing/malformed-register batches.

gorkem2020:
Keeping this behavior for now: it is deliberate policy rather than an oversight, and we would like a human maintainer's call before changing it.

The rationale follows the principle [AliceLJY articulated on #912](https://github.com/CortexReach/memory-lancedb-pro/issues/912#issuecomment-5003974091): a wrong memory is much worse than no memory at all, because recall re-injects stored errors and re-confirms them on every later session. In a batch where the register is non-real AND a sibling item is already constructed-tagged, a real-tagged durable is exactly the shape a mislabeled fiction item takes. So the batch pass prefers dropping a recoverable true fact (durable facts recur in later real-register sessions and store then) over risking fiction in the long-term store. Outside that corroborated case, per-item grounding tags remain fully authoritative.

If per-item honoring in this specific case is genuinely the project's preference, accepting the wrong-memory risk that comes with it, we are happy to implement it exactly that way. Could a human reviewer weigh in on which default the project wants here?

rwmjhb (changes requested):
Reviewed head `367b9e8`. The full suite passes, and candidate/register grounding is a worthwhile defense against fictional-memory pollution. The current batch-wide policy still creates a deterministic false-negative path that conflicts with the per-candidate contract.

A `fiction` register drops every durable category regardless of explicit candidate grounding, and any non-real batch containing one constructed-tagged sibling removes every surviving profile/preference/entity/case/pattern. Missing or malformed registers normalize to `mixed` and can trigger the same wipe. Thus a genuine real-world preference or project fact stated alongside roleplay/example content is silently discarded, even though the extraction prompt says real asides should retain their natural category. Please make contradiction handling candidate-specific: preserve independently supported real-grounded durables and recheck only ambiguous candidates, with mixed/fiction/malformed-register persistence tests.

The inverse gap also remains for `events`: they are outside `DURABLE_CATEGORIES`, so a fiction-register plot event mistagged `grounding=real` bypasses the register guard and persists. Add a fiction-aware event check that distinguishes narrow session-summary events from in-story plot events, including a real-mistagged regression.

Finally, apply `grounding` and `conversation_register` provenance consistently to supersede, contextualize, contradict, and merge paths; direct creates are currently the only fully covered path. Requesting changes.

gorkem2020:
Thanks for the re-review. Point 1 is the batch-wide versus per-candidate demotion question that is already under maintainer arbitration on this PR (see the earlier arbitration request above). The fail-closed batch demotion is deliberate: candidate-level grounding tags are exactly the signal that fiction contexts corrupt, so preserving a sibling because it self-reports grounding=real would reopen the pollution path this PR exists to close, and a wrong durable memory is worse than a missed one. We are holding that design unchanged until the arbitration lands, and deliberately not growing this PR's scope in the meantime.

Points 2 (fiction-register events outside the durable guard) and 3 (provenance fields on the supersede, contextualize, contradict, and merge paths) are fair adjacent observations. Both are queued as follow-ups behind the arbitration outcome, since their right shape depends on how the register-guard semantics settle. No code changes in this round by design.

gorkem2020:
Pushed the complete grounding package as a single commit on top of the reviewed series: [3cd0fd7](https://github.com/CortexReach/memory-lancedb-pro/pull/927/commits/3cd0fd7).

The per-candidate direction from the last review is implemented. The batch-wide demotion is replaced by a scoped second pass (a grounding reviewer) that fires only when the register verdict and the per-item tags are incoherent, or when real-tagged durables sit beside constructed siblings. It re-judges each candidate against the conversation and its verdict is final, so independently supported real durables now survive their batch. The deterministic batch demotion remains only as the fail-closed fallback (reviewer failure, or legacy payloads that assert no register while tagging constructed siblings).

Also in this commit: the grounding taxonomy is generalized from game/roleplay framing to the full factual-vs-hypothetical axis, the few-shot examples block is removed (mechanical rules replaced it), the unsure tie-break defaults to constructed, and a reading-order instruction makes the extractor interpret each message through its understanding of the whole conversation. A new test suite covers the reviewer trigger cells, per-item preservation, fail-closed behavior, and empty or partial verdict quarantine. Full suite green, dist rebuilt in the commit.

rwmjhb (changes requested):
Re-reviewed head `3cd0fd7`. The grounding direction and regression coverage are substantial, and the full suite passed, but two fail-closed gaps still allow constructed memories through the exact paths this PR is intended to harden.

First, `events` are excluded from `DURABLE_CATEGORIES`, and both the contradiction trigger and failed-rejudge quarantine operate only on durable categories. A fiction-register plot event mistagged `grounding=real` can therefore survive reviewer failure (or avoid the constructed-sibling rejudge cell) and be persisted as a real event. An exact-head probe persisted an in-story boarding event with `conversation_register=fiction`. Please make non-real-register event handling distinguish an event about participating in fiction from an event within the fiction, and fail closed for the latter. Add a regression covering reviewer failure/missing verdict.

Second, the rejudge response applies `conversation_register` before validating result coverage. A partial or empty `results` array that reports `conversation_register=real` disables quarantine for every omitted first-pass real tag. An exact-head partial-verdict probe persisted an unadjudicated hypothetical preference. Please require complete, unique, valid coverage for every candidate before accepting any reviewer register change; partial, duplicate, fractional, out-of-range, or invalid verdicts should fail closed using the original asserted register.

The non-create provenance gaps and the independent `extractionPolicy` configuration surface are worth follow-up, but the two persistence bypasses above are the merge blockers. Requesting changes.

gorkem2020:
Both bypasses fixed, plus two more the pre-push review turned up on the same paths.

**Judge-gated events.** `events` now requires a positive grounding verdict to survive a fiction-register batch, so an in-fiction plot beat mistagged `real` fails closed. A missing verdict, an omitted index, and a failed judge all drop it. The rejudge trigger counts real-tagged judge-gated items in a fiction batch, so the legitimate about-the-fiction note ("ran a three-hour session") still gets adjudicated instead of being dropped for want of a verdict; mixed batches are untouched, keeping coherent shapes off the judge. Two existing tests asserted the old behaviour (events surviving a fiction batch with no judge available at all) and now pin both directions instead: confirmed survives, unconfirmed does not.

**Coverage before register.** Validated first now. The adjudicated set only admits unique, integral, in-range indices carrying a usable grounding, so size equality with the candidate count is exactly the complete-unique-valid check you asked for; partial, duplicate, fractional, out-of-range and invalid verdicts all leave it short. One deliberate difference from the letter of the request: on incomplete coverage the asserted register stands **unless the verdict register is stricter**, in which case the stricter one applies. Refusing to relax is the fail-closed property; refusing to tighten would be fail-open in the other direction.

Two more, found while verifying the above:

- A payload omitting `conversation_register` while tagging one item constructed skipped the judge entirely and armed the batch-wide durable wipe, deleting independently-supported real facts. That is the over-drop the per-item rejudge replaced, still reachable on one shape. It now gets a judge call first, and falls through to the same wipe only when the judge cannot answer.
- Verdicts were applied by writing into the object graph returned by the LLM client, which can outlive the call. That had already made one regression test inert: a shared fixture retagged by an earlier case left a later test exercising a different branch while still passing. Verdicts now live alongside the response, which is no longer modified.

Full suite green. New regressions cover judge failure, judge confirmation, partial verdicts, duplicate indices, complete coverage, and response immutability.

One open question rather than a change, because it is a design call and not an oversight. Judge-gating is scoped to fiction batches, so in a **mixed** register an events candidate still rides its own tag: unconfirmed after a failed judge, and equally on the coherent path where no judge is consulted at all. We probed widening the quarantine to judge-gated categories in any non-real register, and it flipped four existing tests that deliberately pin events surviving there, which reads as settled design rather than a gap. The narrower fix would also be incoherent on its own, since the same content persists through the coherent mixed path that no verdict ever touches. If mixed-register events should require confirmation too, that is a contract change worth making deliberately and we are happy to make it.

Noted on the non-create provenance gaps and the `extractionPolicy` surface as follow-ups.

rwmjhb (changes requested):
Re-reviewed new head `0884329`. The previous fiction-event failure path and partial-register relaxation are addressed, and the full suite is green. Two remaining validation cells still reopen constructed-memory persistence.

First, an asserted `conversation_register=real` triggers rejudging only when `rawRealCount === 0`. A partially contradictory batch containing one `grounding=constructed` item and one real-tagged durable does not enter any later cell because those require a non-real register. The constructed item is dropped, but the durable is trusted without adjudication even though the prompt contract defines a real register as entirely factual. An exact-head probe made zero rejudge calls and persisted a hypothetical preference. Please treat any constructed tag in an asserted real-register batch as contradictory whenever another candidate can persist, and apply the same complete-verdict/fail-closed flow.

Second, the coverage comment at `src/smart-extractor.ts:989` is not true for duplicate-plus-complete responses. The loop applies rows immediately, stores coverage in a Set, and only checks `adjudicated.size === rawItems.length`. For two candidates, indices `[1, 1, 2]` pass coverage; the later duplicate overwrites index 1 and can relax the register. The existing `[2, 2]` test passes only because index 1 is also missing. Please validate the entire response before applying it: require exactly N rows, unique integral in-range indices, and one valid grounding per index, then commit the temporary map/register atomically. Add conflicting `[1,1,2]`, extra-row, and reordered-complete regressions.

The mixed-register event contract can remain a deliberate policy choice; I am not repeating it as a blocker. Requesting changes for the two structural fail-closed gaps above.

gorkem2020:
Both structural gaps are closed in [64d5cc9](https://github.com/CortexReach/memory-lancedb-pro/pull/927/commits/64d5cc9), and our own adversarial pre-push pass surfaced a third defect in the first version of the fix, corrected in the same commit.

**Asserted-real batch with a constructed sibling.** That shape now has its own rejudge cell. An asserted `real` register carrying a `constructed` tag contradicts itself, so a real-tagged durable that could persist beside that sibling is adjudicated rather than trusted on the assertion, and the unadjudicated quarantine applies in the new cell as well as in the non-real registers. A clean asserted-real batch with no constructed tag still fires no judge and costs exactly one call, pinned by a test.

**Whole-response validation.** The verdict is staged into a temporary map and validated before any of it is applied: exactly one row per candidate, unique integral in-range indices, one valid grounding per index. It is then committed atomically, or not at all. Regressions cover the conflicting `[1,1,2]` case you named, an extra row `[1,2,3]`, a fractional index, and a reordered-but-complete `[2,1]` that must still be applied so the judge can keep rescuing rows.

**One behavior change worth naming explicitly.** Whole-response validation means a partial verdict no longer earns partial credit: previously a judge that answered 1 of 2 candidates had its one answer honoured and only the omitted candidate was quarantined. Under "exactly N rows" that response is applied in no part and both durables are quarantined. The pre-existing partial-verdict regression is updated to the new contract rather than carved out, since honouring rows from a response we have just classified as untrustworthy is the thing the atomic commit exists to prevent.

**Third defect, found by our own pass.** The first version of the gate judged rows before normalizing them, so ordinary value variance classified a semantically complete verdict as malformed, and whole-response rejection then amplified that into discarding every confirmation and rescue in the response: an index the model quoted (`"index": "1"`), or a grounding carrying trailing punctuation or a parenthetical qualifier (`real.`, `constructed (in-story)`), zeroed the batch's durables where the same verdict written plainly persisted them. Rows are now normalized before the gate judges them, which leaves the atomic-commit property intact. Normalization reads leading enum tokens only, so a negated or unrecognized value (`not real`, `maybe`) still invalidates the verdict.

**One residue we are noting rather than changing here.** When the rejudge call returns nothing usable the batch fails closed and every durable in it is demoted, and at that layer a transient gateway failure is indistinguishable from an unusable answer. The fail-closed direction is deliberate and matches this PR's premise, and it now logs at info level since it discards content, but a transient-class retry plus a caller-visible signal are worth having. Those belong on both model call sites in this path rather than only the newer one, and the surrounding cursor and rate-limiter contract lives outside this PR, so we are tracking it separately instead of widening this one.

Full suite, typecheck, build, and the `core-regression` group are green on the new head.

rwmjhb (changes requested):
Re-reviewed head `64d5cc9`. The latest commit fixes the previous asserted-real durable-sibling bypass and validates rejudge indices atomically; the full suite passes. Two remaining fail-open paths still allow constructed content to persist.

1. In `src/smart-extractor.ts`, the asserted-real/unasserted contradiction cells require `hasRealTaggedDurable`. `event` is excluded from `DURABLE_CATEGORIES`, while its positive-confirmation gate runs only for a fiction register. On this exact head, a constructed-tagged sibling plus a real-tagged in-story event made zero rejudge calls and persisted the event under both `conversation_register="real"` and an omitted register. Include real-tagged fiction-judged categories in those contradiction cells and require a complete positive verdict before persistence.

2. First-pass grounding/register parsing requires exact enum equality, even though the second-pass parser accepts a leading enum token. An extraction result with `grounding="constructed (in-story)"` made zero rejudge calls and was persisted as `grounding="real"`; decorated register values fail open similarly. Normalize present first-pass values with strict token boundaries, and route unusable present values to adjudication or fail closed rather than converting them to permissive defaults.

Please add regressions for constructed-sibling + real-tagged event batches under real/missing registers, and for decorated first-pass grounding/register values. The alternate-path provenance and deferred `extractionPolicy` plugin wiring are useful follow-ups, but they are not the blockers here.

gorkem2020:
Both blockers are addressed in [aa25f46](https://github.com/CortexReach/memory-lancedb-pro/pull/927/commits/aa25f46), with one scoped residue raised below rather than decided here.

**Judge-gated categories in the contradiction cells.** The asserted-real and unasserted cells now arm on a real-tagged judge-gated candidate as well as on a real-tagged durable, and the persistence gate requires a positive verdict for judge-gated categories wherever a constructed-sibling cell fired, not only under a fiction register. Regressions cover the constructed-sibling plus real-tagged-event batch under both an asserted `real` register and an omitted one, plus a confirming-verdict case proving the cell adjudicates rather than wipes.

**Token boundaries on the first pass.** You were right that the asymmetry was the bug: the second-pass parser already read a leading enum token while the first pass demanded exact equality. Both first-pass reads now use the same parser. Note there were two sites, not one: the cell predicate, and the candidate loop that actually decides persistence. Fixing only the predicate left `grounding="constructed (in-story)"` still persisting, so the regression asserts on the stored outcome rather than on the cell.

**One residue, deliberately not taken unilaterally.** Under an asserted `mixed` register, a real-tagged in-story event beside a constructed sibling is still unadjudicated. Arming that cell for judge-gated candidates is a one-line change, and we tried it, but it inverts an existing pinned invariant that a coherent mixed batch spends no rejudge call, and it costs one extra model call for every mixed batch carrying an event. Since `mixed` means both kinds of content are legitimately present, a constructed sibling there is arguably coherent rather than contradictory, so the question is a cost and policy call rather than a correctness one. Happy to make it if you want that cell armed; it needs the pinned invariant retired in the same commit.

**Second residue, on your other prescription.** We normalized present values with strict token boundaries as asked, but kept the documented fail-open for values carrying no recognizable token at all (`"unsure"`, a number). That behavior has its own regression asserting non-enum groundings are kept as real, so inverting it changes a documented legacy-payload contract for any model that answers outside the enum. We would rather change it on your say-so than quietly, so it is unchanged here.

Four pre-existing regressions encoded the old judge-unavailable outcome for events in a contradiction batch and are updated to the new contract. Two now supply the adjudication their intent depends on, which keeps the false-positive guard meaningful by showing real content surviving on confirmation rather than by lowering the bar. Two keep pinning the judge-unavailable path and now expect the judge-gated aside to fail closed alongside the durable.

The alternate-path provenance and the deferred `extractionPolicy` plugin wiring remain open as you noted. Full suite, typecheck, build, and the `core-regression` group are green on the new head.

rwmjhb:
Thanks for the follow-up. Current head `aa25f46` has merge conflicts with the latest base, so I cannot verify the new grounding fixes against the current code yet. Please rebase onto the latest `master`, resolve the conflicts, rerun the grounding tests and full suite, and push the resolved head for re-review.

rwmjhb (approved):
Re-reviewed head `b1b3017`. The previous asserted-real sibling and atomic-verdict blockers remain fixed after the rebase, the full suite passes, and a clean build leaves `dist/` unchanged. The grounding pipeline now provides a meaningful deterministic improvement over the pre-PR behavior, with complete verdict validation and fail-closed reviewer handling.

There are still useful follow-ups around mixed-register in-story events, present-but-invalid grounding values, decorated second-pass registers, persistence provenance, and the direct-only `extractionPolicy` surface. Those cases are narrower than the original unconditional persistence path and do not block this iteration. Approving.

PR #930, opened 2026-07-12
fix: skip auto-capture for internal memory sub-sessions

gorkem2020:
## Problem

The `agent_end` auto-capture hook runs the smart extractor over every successful agent run, but it never checks whether the run is an internal memory sub-session. The memory-reflection distiller executes as an embedded run (session key prefix `temp:memory-reflection:`), and when it finishes, its `agent_end` is captured like ordinary conversation. The extractor then processes the reflection meta-prompt itself (section scaffolding such as "User model deltas" and "Decisions (durable)" plus the transcript the prompt re-embeds).

Observed on a live gateway: every content-rich `/reset` triggered a second extraction burst (1 extract plus admission, dedup, and merge calls) whose "Recent Conversation" was the reflection generator prompt verbatim. The scaffolding primes durable registers, so low-quality items (a roleplay house rule as a durable preference, a ping token as profile identity) were admitted as user memories. The bug is intermittent by content richness (thin reflections trip the value gates), which kept it hidden.

Both guards already exist (`isInternalReflectionSessionKey`, `isMemorySubsessionKey`) and every sibling reflection hook applies them; the auto-capture hook was the single omission. [#922](https://github.com/CortexReach/memory-lancedb-pro/pull/922) recently closed the same class for auto-recall; this closes the auto-capture half.

## Fix

Resolve the session key at the top of the auto-capture path and return early when it matches an internal reflection or memory sub-session, with a debug log in the sibling hooks' style.

## Tests

Four new cases, red-proven before the fix: distiller session key skipped, memory sub-session key skipped, ordinary key proceeds, event-carried key resolved correctly.

Deployed and probed with a rich two-message session followed by `/reset`: the reflection now runs alone, with zero extraction, admission, or capture calls behind it (previously a 9-call burst with 5 admitted junk rows on the identical fixture).


app3apps:

app3apps (approved):
Reviewed head `b380130`. The targeted tests and full suite pass, and guarding the reflection/memory sub-session write path is the right fix for internal prompt contamination.

One non-blocking policy boundary should be tightened: `isMemorySubsessionKey()` matches every key containing `:subagent:`, so the new write-path guard also suppresses auto-capture for ordinary delegated-agent sessions. The predicate was previously used on read/injection paths, where avoiding re-injection is not the same contract as dropping persistence. Please consider a write-specific predicate that matches concrete memory-owned forms (for example reflection and `:active-memory:`), plus a positive regression proving an ordinary delegated subagent can still produce durable facts.

The `(event as any).sessionKey` fallback is covered at runtime but bypasses the hook type contract; narrowing an unknown event shape or typing the compatibility field would also make that path less brittle.

app3apps:
This PR has developed merge conflicts again after the latest changes landed on `main`. The approval above applies to head `b380130`, but GitHub now reports the branch as conflicting.

Please rebase onto the current `main`, resolve the conflicts, and rerun the test suite. Once the updated head is pushed, we will re-check the conflict resolution before merge.

app3apps (approved):
Re-reviewed rebased head `9263f20`. The targeted tests and full suite pass, and the internal reflection/memory sub-session guard still addresses the reported prompt-contamination path.

Non-blocking follow-ups are the broad `:subagent:` policy boundary, the `(event as any).sessionKey` cast, and production/dist coverage for active-memory and keyless cases.

Approving.

app3apps:
`#932` has now merged, and this PR reports merge conflicts with the current `main` branch (`CONFLICTING` / `DIRTY`). Please rebase onto the latest `main`, resolve the conflicts, and push the updated branch.

The existing approval is for head `9263f20`; once the rebased head and CI are available, we will re-check the resulting changes.

app3apps (approved):
Re-reviewed rebased head `b566142`. The branch is clean, CI is green, and the targeted/full suites pass. The internal reflection/memory sub-session guard still fixes the reported prompt-contamination path, and I found no new blocker in the conflict resolution.

The broad `:subagent:` exclusion policy, the `(event as any).sessionKey` compatibility cast, and active-memory/keyless production coverage remain non-blocking follow-ups.

Approving.

gorkem2020:
Checked all three findings against the actual code and its sibling call sites.

Findings 1 and 3 are addressed in [6c5d1d4](https://github.com/gorkem2020/memory-lancedb-pro/commit/6c5d1d4708116a50cbd45fba14edb06c8ea0366f):

1. The broad `:subagent:`/`:active-memory:` exclusion is correctly described as over-broad relative to the function's stale doc comment, which only described its original injection-avoidance use. But it's not a bug: three other hooks in this file (auto-recall injection, reflection injection, self-improvement reminders) already reuse the exact same check with their own "skip for sub-agent sessions" comments, establishing "a delegated subagent's context comes from the parent" as a deliberate, consistent rule across every memory-adjacent hook, not something specific to memory-internal completions. Auto-capture following that same convention is consistent, not an accidental widening. Rewrote the comment to state the real, broader contract so it stops reading as memory-internal-only.
3. Added the two session-key shapes that weren't covered: the `:active-memory:` half of the guard (only `:subagent:` had a test) and a genuinely keyless session, confirming it is not misclassified as internal.

Finding 2 does not hold up: the hook's own type signature already declares `(event: any, ctx: any)`, matching every sibling hook in this file, so `(event as any).sessionKey` is a no-op cast, not a bypass of a stricter contract that does not exist here.

PR #931, opened 2026-07-12
fix: route reflection mapped rows through admission control and ground the distiller prompt

gorkem2020:
## Problem

Reflection mapped rows bypass admission control entirely. On `/reset`, the distiller's markdown sections (User model deltas, Agent model deltas, Lessons and pitfalls, Decisions) are mapped to ordinary user-memory rows (`preference`, `fact`, `decision`) and written via `bulkStore` with no gate, while extraction candidates from the same conversation go through the full admission pipeline. Anything the distiller writes into those sections becomes durable user memory unconditionally, and generic semantic recall serves it back as fact.

Live demonstration: a roleplay session's fictional canon ("favorite drink is plasma coffee") passed through the reflection door into bare `preference` rows and was asserted back to the user as fact on the next session, while the extraction door correctly suppressed the identical content.

## Fix

Two layers:

1. **Distiller prompt rule**: content from in-fiction, roleplay, or test frames may be summarized in the Context or Open loops sections but must never be written into the sections that become durable rows.
2. **Admission routing**: when admission control is enabled, each mapped row is evaluated by the same `AdmissionController` instance the extractor uses before storage (new `reflection-mapped-admission` module; the controller is exposed via an accessor on the extractor). Rejected rows are skipped with the reason logged; admitted rows persist the full admission audit in metadata with `provenance: "memory-reflection-mapped"`. For scoring, mapped categories are resolved to the smart registers the type priors expect (model deltas as preferences, lessons as cases, decisions as events). Behavior is unchanged when admission control is disabled, and infrastructure errors fail open so a transient failure cannot suppress reflection persistence.

A known limitation is documented in code and in `docs/notes/reflection-write-path-findings.md`: with smart extraction disabled there is no controller instance to borrow, so that configuration keeps the legacy ungated write. The notes file also records adjacent findings for discussion (the caller-less mapped-rows reader, cross-writer double-store of durable decision bullets, and the `memory_layer` labeling asymmetry between subsystems).

## Tests

8 cases: admitted and denied mapped-row paths, admission-disabled passthrough, error fail-open, and prompt-content assertions for the new distiller rule.

Deployed and exercised end-to-end: a reset over a session containing a genuine lesson produced a mapped `fact` row carrying a complete admission audit in its metadata (decision `pass_to_dedup`, score 0.685, five feature scores, compared-row ids, provenance tag), and a reset over a fiction-framed session produced empty durable sections with the fiction confined to Context.

## Update 2026-07-17: burst admissions now share one batched judge call

A reflection burst previously cost one standalone admission LLM call per mapped row (a 15-row distillate meant 15 utility-judge generations, even with batch admission configured for the capture path). The gate module now exposes `gateMappedReflectionEntries`: the whole burst is judged through the controller's `evaluateBatch` when it exposes one (one batched call per controller-side chunk, honoring `utilityMode`), and falls back to the historical per-row `evaluate` path on controllers that predate batch support, so behavior on this branch's own base is unchanged. Per-row decisions, reject reasons, log lines, and audit records (including fail-open audits) are identical in both topologies; a batch-call failure or malformed result fails open for the whole burst. The singular `gateMappedReflectionEntry` is now a one-row wrapper over the plural gate so the two paths cannot drift. The reflection loop is restructured into collect-then-gate; one deliberate shift is that the mapped-entries cap now counts gate-eligible rows rather than admitted rows (bursts near the cap of 100 do not occur in practice).


app3apps:

app3apps (changes requested):
Reviewed head `b4943e5`. The problem is worth solving and the targeted/full suites pass, but the admission gate cannot enforce the PR's grounding guarantee as wired.

Each mapped candidate is parsed from `reflectionText`, and `gateMappedReflectionEntry()` then passes that same `reflectionText` as `conversationText` to `AdmissionController`. The controller uses `conversationText` for confidence grounding, so a hallucination introduced by the distiller is evaluated against the document that already contains the hallucination. With the current fixture this circular input produces `coverage=1`, `unsupportedRatio=0`, and positive confidence despite having no original-transcript support.

Please thread the original conversation evidence into the gate and add an end-to-end reflection-write regression where a mapped row present only in the distiller output is rejected and never reaches `bulkStore`.

Two related gaps should be addressed while touching the path: `pass_to_dedup` is currently treated as terminal admit-and-store without the downstream dedup/merge stage, and fail-open rows receive no durable marker. The production `index.ts` rejection/audit wiring also needs an integration test; the current tests exercise the helper in isolation.

app3apps (changes requested):
Re-reviewed rebased head `d871c11`. CI and the test suites are green, but the grounding blocker is unchanged.

Mapped candidates are parsed from `reflectionText`, and that same generated `reflectionText` is then passed to `gateMappedReflectionCandidate` as the conversation evidence. `AdmissionController` uses it to score support, so a claim that exists only in distiller output can validate itself and become durable memory.

Please pass the original pre-distillation conversation as the grounding evidence, keep generated reflection text separate, and add an end-to-end regression proving that a distiller-only claim is rejected and never reaches `bulkStore`. The final storage boundary should also remain gated rather than treating `pass_to_dedup` as terminal permission.

Requesting changes.

app3apps (changes requested):
Re-reviewed head `edd95e8`. CI and the test suites pass, but the grounding blocker is unchanged.

Mapped candidates are parsed from generated `reflectionText`, and that same generated text is passed to `gateMappedReflectionCandidate` as the conversation evidence. The admission controller can therefore validate a claim using the distiller output that introduced the claim; a distiller-only hallucination can self-ground and become durable memory.

Please thread the original pre-distillation transcript separately into the gate and add a production hook -> gate -> `bulkStore` regression proving a claim found only in distiller output is rejected and never stored. `pass_to_dedup` should enter the real dedup/merge path rather than become terminal admission, and fail-open rows need durable provenance instead of silently storing as ordinary admitted rows. The production path should not bypass the gate when smart extraction is disabled.

Requesting changes.

gorkem2020:
Addressed three of the five findings in [3752d35](https://github.com/gorkem2020/memory-lancedb-pro/commit/3752d35c507190f3ba240969c4c0b00d2d93cc54), and want to be direct about the other two rather than pretend they're closed.

Fixed:
1. Mapped rows were graded for grounding against their own distiller-generated text, not the real conversation. Renamed the gate's parameter (the old `reflectionText` name was itself misleading) and now pass the actual transcript.
3. Fail-open admissions now get a durable `auditJson` record on the row itself (tagged `failedOpen: true`), not just an ephemeral log line.
4. Added an end-to-end test proving a rejected hallucinated row never reaches `store.bulkStore` while an admitted sibling still does, confirmed meaningful by disabling the reject branch and watching it fail.

Not fixed, with reasoning:

2. `pass_to_dedup` for mapped rows is treated as "admit, subject to a single vector-similarity pre-check," not routed through the same CREATE/MERGE/CONTEXTUALIZE/CONTRADICT pipeline extraction candidates get. That pipeline's real decision step is an LLM call, not a threshold, so building an equivalent for this write lane is a genuine feature addition, not a same-session fix. Documented the actual current behavior in place of the misleading implication.
5. Admission bypasses when smart extraction is disabled because the gate borrows its controller from `smartExtractor` instead of constructing one independently. This is pre-existing behavior explicitly called out in `docs/notes/reflection-write-path-findings.md`'s own "Residual gap" section, not something this PR introduced. Fixing it properly needs the same independent-construction approach another open PR already built for extraction; duplicating that here would create a cross-branch dependency I'd rather not introduce.

Both are real, reasonable asks for follow-up work, just not ones I want to rush into this branch alongside everything else.

app3apps:
Thanks for the update in `3752d35`. I reran the orchestrator against this head, but GitHub currently reports `mergeable=CONFLICTING` / `merge_state_status=DIRTY`, so the review stopped at the conflict gate after R0/R1.

Please rebase onto the latest `master`, resolve all conflicts, and push the clean result. The transcript-grounding, failed-open provenance, and storage-boundary changes are recorded for full re-review on the mergeable head.

gorkem2020:
Rebased onto current master as requested, the branch is mergeable again. New head: [64772d5](https://github.com/CortexReach/memory-lancedb-pro/pull/931/commits/64772d557a02350c03d6df48bf0724350b8c63d2). Conflicts were limited to the test registration files (package.json test chain and scripts/ci-test-manifest.mjs), resolved by keeping both sides' entries; dist is rebuilt from the resolved sources and the PR's test suite plus typecheck pass on the new base.

app3apps (approved):
Re-reviewed the current head `815c83a`. The earlier grounding blocker is resolved: mapped rows are evaluated against the original transcript rather than the distiller output, the production burst path preserves per-row decisions and ordering, and failed-open rows carry provenance when audit persistence is enabled. The focused regressions and full suite pass.

Approving the scoped change. The documented trade-offs remain follow-up material: `smartExtraction=false` still has no controller to borrow, a malformed or failed batch evaluation fails open for the whole burst, disabling audit persistence removes the durable failed-open marker, and `pass_to_dedup` still uses the mapped-row hard-threshold pre-check rather than the extractor's full merge pipeline.

app3apps:
Thanks for the update. PR #924 has now merged, and this branch currently conflicts with the latest `master` (`mergeable=false`, `mergeable_state=dirty`).

Please rebase onto the latest `master` and resolve the conflicts, preserving both the mapped-row admission changes in this PR and the reflection-cache invalidation behavior that landed in #924. After pushing the resolved head, rerun the focused/full test suites and request review again. The existing approval is tied to `815c83a`; the conflict-resolution commit will need a fresh review.

gorkem2020:
Rebased onto the current master, dist rebuilt on the rebased tree, all checks green on the branch.

app3apps (approved):
Re-reviewed current head `14ca9d6`, including the post-approval batch-admission adapter. The build, 18 focused mapped-row admission tests, and all GitHub CI checks pass; the reflection-cache invalidation behavior from #924 remains registered.

`gateMappedReflectionEntries` preserves row/result order and audit propagation, retains the documented fail-open behavior, and cleanly falls back to per-row `evaluate`. The current production `AdmissionController` does not implement `evaluateBatch`, so runtime behavior on this base remains per-row as the PR describes. I found no merge-blocking regression in the new commits.

PR #933, opened 2026-07-12
chore: send the internal call label as a request header on the openai-compatible path

gorkem2020:
## Problem

`completeJson` receives an internal `label` for every memory-pipeline call (extraction, admission-utility, dedup, merge), but the label is never transmitted. In observability tooling, every call appears as an indistinguishable generic LLM generation, and pipeline stages can only be told apart by reading prompt text.

## Fix

Send the label as a sanitized `x-memory-call-label` request header on the openai-compatible path. Prompt content, parameters, and the OAuth path (foreign fixed-shape endpoint) are untouched.

## Tests

Header presence and sanitization asserted in the existing mock-server test.

### DIFF
--- a/dist/src/llm-client.js
+++ b/dist/src/llm-client.js
@@ -74,6 +74,11 @@ function recoverJsonFromReasoning(reasoningText) {
 function shouldDisableReasoningForJson(model) {
     return /qwen3|deepseek.*r1|qwq/i.test(model);
 }
+/** Restrict a call label to header-safe characters (labels are internal literals). */
+function sanitizeLabelHeader(label) {
+    const cleaned = label.replace(/[^A-Za-z0-9._-]/g, "-").slice(0, 64);
+    return cleaned || "generic";
+}
 function previewText(value, maxLen = 200) {
     const normalized = value.replace(/\s+/g, " ").trim();
     if (normalized.length <= maxLen)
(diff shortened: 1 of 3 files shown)


app3apps (approved):
Reviewed head `0f16b6f`. CI and the full suite pass. The new header is scoped to the API-key/OpenAI-compatible path, the generated `dist` artifact matches the TypeScript source, and the existing OAuth request path is unchanged. I found no merge-blocking behavior issue.

Non-blocking: the added test only forwards an already-safe `api-key-probe` label. Please add sanitizer cases for invalid/control characters, the 64-character limit, and the empty-label fallback when convenient. The `request as any` cast predates this PR.

Approving.

gorkem2020:
Added the requested sanitizer coverage in [8a99aec](https://github.com/gorkem2020/memory-lancedb-pro/commit/8a99aec1d6f9d5389def5298108ab4fd0cfc083f): invalid/control characters (including CR/LF), the 64-character truncation, and the empty-label fallback.

One correction worth noting: the fallback to `"generic"` is only reachable via a genuinely empty label, not a label that is merely all invalid characters. `sanitizeLabelHeader` replaces each rejected character with `-` (itself a valid character), so something like `"!!!###"` sanitizes to `"------"`, never to an empty string.

PR #934, opened 2026-07-12
fix: gate the auto-capture regex fallback behind smart extraction and admission

gorkem2020:
## Problem

When the smart extractor does not handle a turn, the auto-capture hook falls through to the legacy regex capture, which stores matching texts verbatim via `bulkStore`, with no distillation, no admission, and none of the extraction pipeline's safeguards. Two trigger paths:

1. `cumulativeCount < extractMinMessages`: every content-rich first message of a session is eligible for raw capture before the extractor ever gets to judge it.
2. The boundary-skip continuation after an extraction that persisted nothing.

Observed live: the first message of a fresh session (a roleplay opener) was stored verbatim as a confirmed `fact` row within seconds, bypassing the grounding and admission stack entirely. Sessions that open with a short greeting never hit this, which is why it stayed hidden; sessions that open with a rich message (how real users start conversations) hit it immediately.

## Fix

1. **Below-threshold turns with smart extraction enabled**: skip the regex fallback and defer. The cursor is rolled back for history-carrying sessions so the deferred text re-enters the next turn's extraction input (proven by regression test); ingress-fed sessions keep their accumulator semantics, where a rollback would wedge the minMessages window.
2. **When the fallback legitimately runs** (smart extraction disabled, or the boundary-skip continuation): captures are routed through admission control when it is enabled, after the cheap dedup checks, with the audit persisted under `provenance: "auto-capture-regex-fallback"`. Admission disabled keeps the legacy passthrough.

## Tests

9 cases, 3 red-proven against the previous behavior: the below-threshold skip, deferred-text re-inclusion in the next extraction, gated boundary-skip continuation, plus pins for legacy behavior with smart extraction disabled.

The identical rich first message that previously produced a verbatim `fact` row now produces nothing; the follow-up turn's extraction input contained the deferred first-message content, produced correctly grounded candidates, and logged "skipping regex fallback" on the empty-outcome path.


app3apps:

app3apps (changes requested):
Reviewed head `cd3c56f`. The direction is valuable and the targeted/full suites pass, but the below-threshold change introduces two blocking data-loss regressions.

1. In ingress-fed sessions, `pendingIngressTexts` is copied and removed and the accumulator advances, but the rollback happens only when that snapshot is empty. The function then returns before regex fallback. A below-threshold ingress turn is therefore neither persisted nor available to a later extraction.
2. This also removes the deterministic behavior for explicit memory requests. Two existing cases containing `请记住我的饮品偏好是乌龙茶` were changed from asserting a regex capture to asserting that fallback is skipped. Persistence now requires another turn and a successful discretionary LLM extraction; a single-turn session loses the requested memory entirely.

Please retain below-threshold ingress text for the next extraction or preserve a safely admission-gated explicit-memory path, and add regressions covering real `message_received` ingress and a single-turn explicit `remember` request.

Also, admission evaluation currently fails open without storing a durable failed-open marker. Since this PR exists to close an admission bypass, please either fail closed for raw fallback text or persist provenance that makes bypassed rows auditable.

app3apps (changes requested):
Re-reviewed rebased head `f9bf6ed`. The rebase is clean and CI is green, but the production-path data-loss blocker remains.

In ingress-fed sessions, `pendingIngressTexts` is consumed and removed, the cursor advances, and the below-threshold branch returns before the fallback path. That permanently drops the ingress text, including explicit remember-style requests.

Please preserve or requeue deferred ingress text until it is actually extracted, and add production-path coverage for `message_received` plus a single-turn remember request. A history-carrying session that ends below threshold also needs an end-of-session case so its deferred text cannot disappear. Fail-open admission should leave durable provenance on stored rows.

Requesting changes until the ingress text is no longer lossy.

app3apps (changes requested):
Re-reviewed head `ce16a61`. CI and the full suite pass, but the new expectations still codify deferral without proving eventual persistence.

For ingress-fed sessions, `pendingIngressTexts` is consumed and deleted, the seen cursor advances, and the below-threshold branch returns before fallback. A single-turn explicit remember request can therefore become unrecoverable. A history-backed session that ends before `extractMinMessages` has the same no-flush outcome.

Please preserve or requeue deferred text until extraction succeeds, or persist explicit capture through a safely gated path. Add production-path regressions using real `message_received` ingress, a single-turn remember request, and a session ending below threshold, proving the content is eventually stored exactly once. Admission failures that intentionally fail open should also leave durable provenance on the stored row.

Requesting changes.

gorkem2020:
Fixed two findings in [63eec3f](https://github.com/gorkem2020/memory-lancedb-pro/commit/63eec3fbd1a6642599f2db63e381860533aac637), and want to be direct about how the other three relate rather than claim more than I've actually closed.

Fixed:
2. For ingress-fed sessions, a below-threshold turn consumed and deleted the pending ingress queue, counted it toward the threshold, then discarded the actual text. The counter advanced correctly but the content of every deferred turn except the most recent was gone by the time a later turn crossed the threshold. Now re-queues the consumed texts back into the pending map on defer, mirroring what the existing cursor rollback already does for history-carrying sessions.
5. The regex-fallback gate's fail-open path now gets a durable audit record on the row itself (`failedOpen: true`), not just an ephemeral log line, matching the same fix just applied to the sibling mapped-row admission gate on another PR.

Not separately fixed:

1. The test-coverage ask is substantially closed by the new ingress-flow test added alongside the fix above, which proves a second deferred turn's text survives into the extraction that finally crosses the threshold.
3. The single-turn explicit-remember scenario is the same underlying gap as finding 2 for ingress-fed sessions (now fixed) and was already handled for history-carrying sessions by the existing rollback. The one case that is genuinely unrecoverable either way is finding 4 below.
4. A session that ends before crossing the threshold has no flush path, and that's real. I looked at adding a `session_end` handler to force a final extraction, but the risk is concrete: if the last `agent_end` call already crossed the threshold and reset the counter, a naive `session_end` flush reading the same accumulated-text buffer could re-extract content that was just successfully persisted, which would be a worse bug (duplicate memories) than the current silent gap. Getting that distinction right needs its own careful pass, not a same-session addition alongside everything else here.

app3apps:
Thanks for the update in `63eec3f`. I reran the orchestrator against this head, but GitHub currently reports `mergeable=CONFLICTING` / `merge_state_status=DIRTY`, so the review stopped at the conflict gate after R0/R1.

Please rebase onto the latest `master`, resolve all conflicts, and push the clean result. The ingress requeue and failed-open provenance changes are recorded; we will verify them against the actual merge result after the rebase.

gorkem2020:
Rebased onto current master as requested, the branch is mergeable again. New head: [b0f05a3](https://github.com/CortexReach/memory-lancedb-pro/pull/934/commits/b0f05a3065b19957ff65bf79fc97aed1dc701005). Conflicts were limited to the test registration files (package.json test chain and scripts/ci-test-manifest.mjs), resolved by keeping both sides' entries; dist is rebuilt from the resolved sources and the PR's test suite plus typecheck pass on the new base.

app3apps (changes requested):
Re-reviewed the current head `b0f05a3`. The ingress requeue and failed-open provenance fixes are useful, and the full suite passes, but two behavior gaps still block this change:

1. The below-threshold branch returns before regex fallback and only preserves data for a later turn. If the session ends before `extractMinMessages`, an explicit remember-style request that the legacy fallback previously persisted is silently lost. Please add an end/session-boundary flush or another exactly-once path for intentional capture, with a regression for a short session that has no later turn.
2. The PR body says the legitimate fallback is routed through admission when `smartExtraction=false`, but the controller is obtained from `smartExtractor?.getAdmissionController()`. With smart extraction disabled that is always null, and `gateRegexFallbackCapture` admits unconditionally. Please construct/use admission independently of SmartExtractor for this configuration, or otherwise make the advertised safety boundary real and test the production configuration.

There is also an accounting edge to cover while fixing the first item: ingress `cumulativeCount` advances before deferral, while requeued texts are counted again and the queue is capped at six, so repeated deferrals can cross the threshold early while silently dropping older input. Add an exactly-once/count-and-cap regression around that flow.

app3apps:
Since #924 merged, GitHub now reports this branch as conflicting with the latest `master`. Please rebase onto or merge the current `master`, resolve the conflicts, and push the updated head. The prior review blockers remain recorded and will be re-reviewed on the clean revision.

gorkem2020:
Rebased onto the current master, dist rebuilt on the rebased tree, all checks green on the branch.

app3apps (changes requested):
Re-reviewed head `5d3dbd3` after the rebase. The build, focused suites, and GitHub CI pass, but the rebase leaves the two previously requested correctness fixes unchanged:

1. In the below-threshold `agent_end` path, the consumed text is rolled back or requeued and the hook returns. There is still no session-boundary flush. If a session ends before `extractMinMessages` (including a one-turn explicit "remember this" request), there is no later turn to consume the deferred content, so a capture that previously persisted is now silently lost. Please add an exactly-once safe flush and an end-to-end short-session regression that inspects storage.

2. When `smartExtraction=false`, no `SmartExtractor` or `AdmissionController` is created. The caller passes `null` to `gateRegexFallbackCapture`, whose null-controller branch unconditionally admits the raw regex capture. Thus admission remains completely bypassed in the configuration where regex fallback is the primary capture path, despite the PR's stated admission-gating behavior. Please make the controller available independently of smart extraction (or otherwise enforce the gate) and add a production-configuration rejection test.

These are behavioral blockers, not test-infrastructure issues.

app3apps (changes requested):
Re-reviewed head `c2678a7`. The TypeScript build, the 12 focused fallback/extractor tests, and GitHub CI pass, but the two previously requested production-path fixes are still missing.

1. Below `extractMinMessages`, `agent_end` requeues ingress text or rolls back the history cursor and then returns. There is still no session-boundary flush. If the session ends there, including a one-turn explicit “remember this” request, no later turn consumes the deferred content and nothing is persisted. Please add an exactly-once end/session flush and a storage-level short-session regression.

2. The fallback gate still receives `smartExtractor?.getAdmissionController() ?? null`. With `smartExtraction=false`, no `SmartExtractor` exists, the controller is always null, and `gateRegexFallbackCapture` deliberately passes the raw capture through. The new focused test explicitly pins that legacy passthrough behavior. Please make admission available independently of smart extraction (or enforce an equivalent gate) and add a rejection test for the real disabled-smart-extraction configuration.

The monolithic local test command exceeded the harness's 180-second limit in an unrelated real-LanceDB rollback section; that is a verification limitation, not the basis for this review. Requesting changes for the two code paths above.

rwmjhb (changes requested):
Reviewed head `fab3eb1`. Gating raw fallback behind extraction/admission is the right direction, but three lifecycle/configuration paths still defeat the intended safety behavior.

1. Below-threshold history is only cursor-rolled back and ingress text is only requeued. No `session_end` path consumes that deferred state, so a one-turn or short session ending before `extractMinMessages` can leave an explicit remember request unpersisted. Add an exactly-once terminal flush and a storage-level one-turn regression that proves the request is saved once, not dropped or duplicated.

2. `SmartExtractor`, its LLM client, and its `AdmissionController` are constructed only when smart extraction is enabled. With `smartExtraction=false` and `admissionControl.enabled=true`, the fallback receives a null controller and admits unconditionally; the current tests explicitly pin that bypass. Construct admission independently of SmartExtractor and add an end-to-end rejection test for this supported configuration.

3. Each `agent_end` treats the entire requeued ingress snapshot as new and adds its length again to the cumulative count. Counts grow 1, 3, 6 for three unique messages, and both queues cap at six; with higher valid thresholds, older deferred texts are evicted before extraction. Track unique ingress separately from retained payload, never recount requeued texts, and cover threshold/cap/exactly-once behavior.

`persistRejectedAudits` should also receive regex-fallback rejections rather than only SmartExtractor rejections. The full-suite OAuth callback test was red with localhost `ECONNREFUSED`, but that appears unrelated to these capture changes; the request is based on the concrete paths above.

gorkem2020:
Head f056d31 addresses the three paths and the audit rider:

1. Terminal flush: a new session_end consumer drains deferred capture state exactly once, serialized behind any in-flight capture run. Both deferral shapes are covered: requeued ingress texts, and rolled-back history texts (now retained for the flush, and cleared once a threshold extraction has consumed them). Storage-level regressions: a one-turn ingress session's remember request persists exactly one row, a second session_end is a no-op, and history already consumed by a threshold extraction is not re-extracted.

2. Admission is constructed independently of SmartExtractor: with smartExtraction=false (or init failure) and admissionControl.enabled=true, a standalone controller now gates the regex fallback and the mapped-reflection path through the same resolution. End-to-end tests cover both reject (no row stored, audit persisted) and admit (row stored) under this configuration.

3. Requeued ingress is never recounted: the requeue records how many retained texts were already counted, so three unique messages count 1, 2, 3 instead of 1, 3, 6, and retention follows max(6, extractMinMessages) so higher thresholds cannot evict deferred content. Tests pin turn-3 firing (not turn 2), exactly-once text inclusion in the extraction input, and an 8-message threshold reaching extraction with all 8 texts intact.

persistRejectedAudits: regex-fallback rejections now flow through the same amac-v1 rejected-audit writer as extraction rejections, covered in the new suite.

rwmjhb (changes requested):
Re-reviewed head `f056d31`. The source changes address the three previous blockers, and the targeted/full suites pass. The committed runtime is still stale, however.

`package.json` points `main` at `dist/index.js` and includes `dist/**/*` as the runtime package surface. On this exact head, running `npm run build` modifies both `dist/index.js` and `dist/src/autocapture-fallback-admission.js` (169 insertions, 46 deletions). The committed files lack the terminal flush, standalone AdmissionController, corrected ingress accounting, and rejected-audit behavior now present in TypeScript. `prepack` rebuilding mitigates normal npm packing, but direct execution/install from this commit still runs the old implementation. Please rebuild and commit both dist files, then rerun the package-runtime verification.

There is also a cross-session race in the terminal flush coordination: the agent-end hook stores one global `__lastRun`, and concurrent sessions overwrite it. A session_end for A can wait for B, flush before A finishes recording deferred state, and then leave A data with no later terminal event. Track in-flight work per session (or serialize by session key) and add an interleaving regression with overlapping A/B agent_end plus A session_end.

Requesting changes for the stale shipped runtime and per-session flush coordination.

rwmjhb (changes requested):
The per-session serialization and rebuilt runtime address the previous global-slot/dist blockers, and the full suite plus an independent build are green. Three current-head paths still block merge:

1. The new `session_end` handler reads only `ctx.sessionKey` / `event.sessionKey`, while the actual lifecycle path elsewhere in this plugin uses `ctx.sessionId`. With the documented payload, `flushSessionKey` is empty and the handler returns before dispatching the terminal flush. The tests currently supply a synthetic `sessionKey`, so they mask the production shape.

2. Pending ingress and terminal/deferred state are deleted before extraction finishes. If extraction throws, the catch returns without restoring that state; a null/exhausted LLM result can also become zero statistics and clear deferred history as if evaluation succeeded. A transient failure can therefore permanently consume an explicit remember request.

3. When `admissionControl.enabled=true` but standalone controller construction fails, the catch explicitly leaves captures ungated. The null-controller branch then admits raw regex captures with no durable failed-open provenance, restoring the bypass this PR is meant to close.

Please wire terminal flushing to the real lifecycle identifier, consume deferred state only after a definitive extraction result (or restore it on failure), and distinguish intentionally disabled admission from initialization failure with fail-closed or audited fail-open behavior. Tests should use the real `session_end` payload and cover both failure paths.

gorkem2020:
All three blockers are addressed in [dd1d247](https://github.com/CortexReach/memory-lancedb-pro/pull/934/commits/dd1d247215a0320485acf416c82ac8add87e5bd6).

1. Lifecycle identifier: the ingress hook now records a sessionId-to-sessionKey alias, and the `session_end` handler resolves `ctx.sessionKey || event.sessionKey`, then the alias for `ctx.sessionId || event.sessionId`, then the sessionId itself. Both hooks therefore reach the same per-session buckets whether the host keys its payloads by sessionKey, by sessionId, or asymmetrically. Tests now cover the real payload shapes: a session_end carrying only `sessionId` after mixed-shape ingress, and a host keying both hooks by `sessionId` alone.

2. Consume-after-settle: `extractAndPersist` now surfaces a null or malformed LLM completion as `extractionFailed` on its stats (distinct from a definitive zero-candidate result, which still consumes and still feeds the noise bank). Both failure paths, thrown extraction and `extractionFailed`, restore the consumed state in the same shapes the below-threshold deferral path uses: deferred flush texts for a terminal flush, the slice cursor for history sessions, the pending queue plus counted marker for ingress-fed sessions. The rate limiter is charged only after a usable result. Test: a failed flush extraction followed by a second session_end that successfully retries the same texts.

3. Admission init failure: `gateRegexFallbackCapture` and `gateMappedReflectionEntries` (and the single-row wrapper) now take `admissionRequired`. When admission is enabled but no controller exists (construction failed), they fail closed with a logged reason instead of admitting ungated; intentional disablement keeps the passthrough. Construction failure now logs at error level. The transient per-candidate evaluation error keeps its existing audited fail-open, since a standing init failure and a one-candidate hiccup are different severity classes. Tests cover fail-closed and passthrough for both gates.

rwmjhb (changes requested):
Re-reviewed head `40a9fd5`. The fallback admission gate, lifecycle-ID aliasing, consumed-text restoration, and concurrency coverage are a substantial improvement; the focused/full suites pass and a clean type/dist check was independently confirmed. One lifecycle guarantee still blocks merge.

Below-threshold turns now return after placing their only retained copy in process-local `autoCaptureDeferredFlushTexts` / `autoCapturePendingIngressTexts`. The only terminal consumer is the `session_end` callback, and that callback starts a detached promise without returning it to the host. There is no durable queue, timer, shutdown/dispose drain, or alternate consumer. A one-turn explicit “remember X” can therefore be lost if `session_end` is absent, the process restarts, or shutdown completes before the detached flush. This is a regression from the parent path, which proceeded to regex fallback immediately.

Please give deferred explicit captures a bounded, awaited lifecycle: return/await terminal flush when the host supports it and add a shutdown/dispose drain or another guaranteed retry/flush mechanism. If process durability is intentionally out of scope, at minimum explicit remember requests need a safe synchronous/admitted path that does not depend solely on a future lifecycle event. Add regressions for missing session_end and shutdown while terminal flush is pending.

There is a related dead-end after transient terminal extraction failure: the code restores the consumed text to the same Map but schedules no timer or retry; the current test succeeds only by manually emitting a second `session_end`. Please schedule a bounded retry or retain the state for a guaranteed subsequent consumer.

Requesting changes for the loss window above.

gorkem2020:
Head `100887d` gives explicit remember requests the safe admitted path, awaits the terminal flush where the host can, and closes an adjacent loss window we found while reviewing this round.

**Explicit remember requests no longer depend on a future lifecycle event.** A remember command arriving with a referent is admitted on its own turn: `extractMinMessages` exists to hold back low-value chatter, and it no longer applies once the user has asked explicitly. The immediate path still runs the full smart-extraction stack (grounding filter and admission control), never the raw regex fallback. Eligibility is judged on the filtered texts, so a turn of bare imperatives, or one whose referent the noise filter removed, keeps deferring instead of extracting from nothing. Regressions cover immediate extraction with no lifecycle event at all, command-only turns, and a removed referent.

**The terminal flush is awaited when the host supports it.** `session_end` now returns the flush promise instead of detaching it, so a host that awaits its hooks (including one emitting `session_end` on shutdown) gets the completion guarantee; a regression pins the returned promise. A concurrency regression also pins that a text deferred during an overlapping immediate extraction survives to the flush.

**A completed-but-barren extraction no longer destroys its inputs.** An extraction that completes but persists nothing (zero candidates, or every candidate rejected by admission) is not a failure, so the failure-path restore never ran, yet the pending ingress queue was already consumed and the history slice cursor already advanced. An explicit request whose first pass came back empty therefore ended with zero rows and zero retries, strictly weaker than ordinary chatter. That exit now deposits history texts for the terminal flush and requeues ingress texts for the next turn (the same shapes as a failed extraction) without rewinding the cursor, since a rewind would re-admit the same slice and spend an extraction call on every subsequent turn. Storage-level regressions pin the retry at both venues.

**On the transient terminal-failure dead-end**, we took the retain-state half of your either/or for this PR. Restored texts stay in the deferred map for any subsequent flush, and with the immediate path above the explicit-remember class no longer reaches `session_end` as its only consumer; ingress sessions additionally retry on every later turn. A timer-scheduled retry inside the plugin brings a lifecycle of its own (cancellation on reset and dispose, double-fire against a late `session_end`), so we would prefer to land it as a follow-up with its own regressions unless you consider it blocking here.

Process restart durability (the process-local maps) stays out of scope, per the minimum path you outlined. Full suite, typecheck, and a fresh dist are green on the new head.

rwmjhb (changes requested):
Re-reviewed head `59a5341`. The new barren-result retry and lifecycle coverage improve the earlier failure paths, and the focused/full suites pass. One terminal-flush blocker remains.

`autoCaptureSessionIdToKey` is written only from `agent_end`, and only when that event carries both a differing `sessionId` and `sessionKey`. A supported asymmetric lifecycle shape can therefore miss the alias entirely: `agent_end` carries only `sessionKey`, while `session_end` carries only `sessionId`. The capture, deferred-text, and in-flight maps are keyed by `sessionKey`; the terminal handler falls back to the raw `sessionId`, awaits an empty in-flight set, and flushes an empty bucket without warning. Every deferred text for that session is silently stranded.

Please establish one canonical lifecycle key from the hook(s) where the identifier relationship is actually available, or preserve aliases independently of `agent_end` carrying both fields. Add an integration regression with `agent_end(sessionKey only)` followed by `session_end(sessionId only)`, asserting that in-flight work is awaited and deferred input is flushed exactly once.

Inline command-plus-payload remember requests, failed immediate-history recovery, settled zero-write outcomes being requeued, and shutdown draining are worthwhile follow-ups, but the cross-hook key mismatch is the merge blocker.

gorkem2020:
Fixed in fb9911a: the alias is now learned through a shared helper at every hook where both lifecycle identifiers co-occur (the unconditional per-turn prompt-build hook, the auto-recall hook, agent_end, and session_end itself), so a sessionKey-only agent_end followed by a sessionId-only session_end resolves to the same per-session buckets.

Added the asymmetric-lifecycle regression you specified: agent_end carrying only the sessionKey followed by session_end carrying only the sessionId must await in-flight work and flush the deferred text exactly once, plus an exactly-once re-fire assertion. The cell fails on the previous head and passes on this one. Full suite green.

rwmjhb (changes requested):
Re-reviewed head `9ff850f`. The previous asymmetric session-id alias blocker is addressed and the focused/full suites pass. One retry-state blocker remains.

The lifecycle caller classifies every result with `created=0`, `merged=0`, and `boundarySkipped=0` as barren and calls `deferBarrenExtractionTexts()`. That shape also represents definitive admission rejections, dedup `skip` decisions, and support outcomes. For ingress-backed runs, the same texts and their counted marker are requeued while the cumulative count stays at or above the extraction threshold. A later `agent_end` can therefore extract the identical snapshot again even with no new ingress, repeatedly invoking extraction/admission, duplicating rejection audits or support evidence, and charging the process-wide limiter until unrelated sessions are throttled.

Please return an explicit outcome that distinguishes retryable extraction failure/barren model output from settled reject, dedup, support, and update outcomes. Consume settled outcomes; requeue only the cases intentionally classified as recoverable. Add repeated-`agent_end` regressions for reject, dedup skip, and support, asserting one extraction/admission evaluation and no repeated quota charge or audit update.

Inline command-plus-payload handling, confirmed-write accounting, rate-limited terminal drains, and remaining lifecycle-key normalization are useful follow-ups, but the settled-outcome retry loop is the merge blocker.

gorkem2020:
The retry-state blocker is fixed in [876b843](https://github.com/CortexReach/memory-lancedb-pro/pull/934/commits/876b8431595a77d3dd68fd82f846f11a374881d7).

ExtractionStats now returns an explicit settledOutcomes flag, stamped when at least one candidate reached a definitive pipeline verdict (create, merge, admission reject, dedup skip, support, or supersede). The lifecycle caller consumes settled zero-persisted runs, advancing the slice cursor exactly like a persisting run, and requeues only barren or failed extractions. A sibling redelivery defect in the same loop is closed with it: an equal-length snapshot fell through the slice condition and re-fed the whole history as new texts, so snapshot identity is now decided by comparing the ordered tail against the previous run's recorded texts; identical redeliveries are consumed while fresh same-length windows still extract (the multi-round merge scenario in the existing branch coverage pins that distinction).

Regressions added as requested, one per settled class: repeated agent_end for an admission-rejected run (single extraction and single admission evaluation), a dedup-skipped run (single extraction, no duplicate row on either fire), and a support-only run (single extraction, support evidence on the anchored row unchanged by the second fire). All three are red without the fix (36 of 39) and green with it (39 of 39); the full chain and the core-regression and cli-smoke groups pass.

The inline command-plus-payload handling, confirmed-write accounting, rate-limited terminal drains, and lifecycle-key normalization are tracked as follow-ups on our side.

rwmjhb (approved):
Re-reviewed head `876b843`. The latest change fixes the prior retry-state blocker: definitively settled zero-persisted outcomes are consumed instead of being requeued as barren work. The targeted auto-capture coverage, full suite, and repository CI pass, and I found no remaining HIGH/CRITICAL issue. Approving.

Follow-ups worth addressing: reconcile queued versus confirmed `bulkStore` writes before retiring capture input; canonicalize known session aliases before every `agent_end` map/in-flight access; cover inline command-plus-payload remember requests; and make terminal failure/rate-limit restoration actually drainable after the single `session_end` delivery.

rwmjhb:
Thanks for the update. After the recent merges, current head `876b843` now has merge conflicts with the latest `master`, so the approved revision cannot be merged as-is. Please rebase onto the latest `master`, resolve the conflicts, rerun the relevant targeted tests and full CI suite, and push the resolved head for a quick re-check.

gorkem2020:
Rebased onto current master (post #952) as requested; review-round history squashed into a single commit for a clean re-verification. Full suite green, dist rebuilt.

rwmjhb (approved):
Reviewed head 7dfb4c3db85af7bd4b736d2d87072b8dc5402481 after the rebase. The capture/admission regression suite and full suite pass, CI is green, and the rebase preserves the previously reviewed lifecycle behavior with no merge-blocking regression found. Non-blocking follow-ups remain around canonical session aliases, confirmed versus attempted bulk writes, and retry classification for policy/grounding drops.

PR #935, opened 2026-07-12
chore: single-source the smart-to-storage category map

gorkem2020:
## Problem

The smart-to-storage category mapping is hand-maintained in two places: a switch in the extractor (`mapToStoreCategory`) and the `SMART_TO_STORAGE_CATEGORY` object in `memory-categories`. The two currently agree, but nothing keeps them agreeing.

## Fix

The extractor delegates to the shared map (`getStorageCategoryForMemoryCategory(category) ?? "other"`), preserving the old switch's defensive default for untyped runtime callers. To keep the delegation type-safe without a cast, the shared getter's return type is tightened via a new exported `SmartStorageCategory` type (type-only change, zero emitted-JS difference in `memory-categories`). A comment documents that `reflection` is minted only by the reflection writer and is deliberately absent from this map.

Byte-identical mapping behavior; no schema or store-format change.

## Tests

Existing suites pass unchanged (the mapping itself is pinned by the extractor tests).


app3apps (commented):
Reviewed head `9081df4`. CI and the full suite pass, but this is currently behavior-neutral maintenance: the two category mappings already agree, the PR links no defect or user-visible impact, and the shared lookup does not fix an observed failure.

Given the maintenance churn and stale base, I recommend closing this PR unless it is tied to a maintainer-approved cleanup objective or a concrete drift-prevention issue. If it is retained, please rebase and state the contract/coverage that this centralization is intended to protect.

gorkem2020:
Rebased onto current master (no conflicts, the 3 commits in between never touched the files this PR changes) and added the requested coverage in [2dfe769](https://github.com/gorkem2020/memory-lancedb-pro/commit/2dfe769facb3682b92801ffa34da167a55415b25): a dedicated test pinning the full 6-category mapping, confirming the extractor-only `reflection` storage category is never produced, and confirming the CLI-facing `resolveToolMemoryCategory` stays in lockstep with the extractor-facing mapping.

On keeping vs closing this PR: no tracked issue drove it, it is a self-initiated drift-prevention chore. The change itself is small, behavior-preserving (confirmed: only a type annotation and export were added, the mapping's actual keys/values are untouched), and now has direct coverage for the contract it protects. Given that, I'd like to keep it rather than close it, but happy to close if that is not the bar for this kind of chore PR here.

rwmjhb (dismissed):
Re-reviewed head `2dfe769`. The author response addresses the previous close recommendation: the branch is rebased onto current master, the protected six-category contract is now stated explicitly, and the new mapping tests are wired into both `npm test` and the storage/schema CI group. The full local suite passes, `npm run build` succeeds, and rebuilding leaves the committed `dist` artifacts clean.

This is modest, behavior-preserving maintenance, but the duplicated production mapping was real and the revised PR now has proportionate regression coverage.

Non-blocking follow-ups: harden the shared object lookup for inherited keys such as `constructor`/`toString` so the claimed untyped fallback always returns `other`; use `SmartStorageCategory` for the extractor mapper return type; and collapse the duplicate JSDoc above that method.

Approving.

app3apps (approved):
Re-reviewed head `2dfe769`. The author response addresses the previous close recommendation: the branch is rebased onto current master, the protected six-category contract is now stated explicitly, and the new mapping tests are wired into both `npm test` and the storage/schema CI group. The full local suite passes, `npm run build` succeeds, and rebuilding leaves the committed `dist` artifacts clean.

This is modest, behavior-preserving maintenance, but the duplicated production mapping was real and the revised PR now has proportionate regression coverage.

Non-blocking follow-ups: harden the shared object lookup for inherited keys such as `constructor`/`toString` so the claimed untyped fallback always returns `other`; use `SmartStorageCategory` for the extractor mapper return type; and collapse the duplicate JSDoc above that method.

Approving.

app3apps:
After #930 and #933 merged, this PR now conflicts with the latest `master` (`CONFLICTING` / `DIRTY`). Please rebase the branch onto current `master`, resolve the conflicts, and push the updated head, then rerun CI.

The existing approval applies to reviewed head `2dfe769`; we will re-check the conflict resolution and resulting diff after the rebased commit is available.

PR #937, opened 2026-07-13
fix: coalesce concurrent recall metadata writes

### LINKED ISSUE #936: Concurrent manual recalls can drop access/governance metadata updates
## Summary

PR #926 keeps manual recall results off the access-metadata write path, but concurrent recalls can still overload the detached write path itself.

With four agents recalling at the same time, all user-facing recall results returned successfully. The resulting per-memory metadata patches then contended on the LanceDB write lock for several minutes, and one patch was dropped.

## Reproduction

- Run four concurrent `memory_recall` calls through `tools.invoke`, each returning three memories.
- Two synchronized four-agent bursts returned 8/8 recall responses without timeout.
- In the second burst, 11/12 metadata updates eventually landed.
- One call logged:

```text
[memory-lancedb-pro] background manual recall metadata patch failed for 1/3 memories
```

The gateway remained healthy throughout.

## Root cause

- Manual recall starts one detached `patchMetadata()` promise per result.
- `patchMetadata()` builds the replacement metadata from a read performed before `update()` acquires the write lock.
- Each update then performs its own LanceDB delete/add cycle under the file lock.
- Concurrent callers can therefore wait through long lock retries, undercount `access_count`, overwrite newer timestamps, or fail to clear suppression/governance fields.

## Expected behavior

- Recall responses remain independent of metadata writes.
- Concurrent access increments are coalesced without loss.
- Access timestamps are monotonic.
- Explicit-recall suppression and bad-recall resets are preserved.
- Updates are merged from rows read inside the write lock and written in a bounded batch.
- Failed batches retry with useful per-ID/error logging.

## Proposed direction

Add a shared debounced write-behind queue for manual-recall metadata events, backed by an exact-ID store operation that reads, merges, and writes a batch inside one lock acquisition.

Related: #925, #926

TurboTheTurtle:
## Summary

- coalesce concurrent manual-recall metadata events by memory ID
- merge access counters and monotonic timestamps from rows read inside one write lock
- batch exact-ID replacements and retry only failed entries with per-ID logging
- preserve scope checks plus explicit-recall governance resets

## Tests

- `npm run test:core-regression`
- `npm run test:storage-and-schema`
- `npm run test:cli-smoke`
- focused recall/store concurrency tests (44 passed)

Depends on #926. This is intentionally stacked: the first commit is #926 and the second commit is the focused #936 fix. Once #926 lands, I’ll rebase this branch onto `master` so the PR contains only the follow-up commit.

Closes #936


app3apps (changes requested):
Reviewed head `1294d1a`. Coalescing these writes is a sensible direction and the targeted/full suites pass, but the production queue can starve indefinitely under the load this PR is meant to handle.

Every `enqueue()` clears and re-arms the sole 50 ms debounce timer. There is no maximum-wait deadline or size trigger, so recalls arriving less than 50 ms apart continuously postpone `flushOnce()`. Unique entries accumulate and no access/governance metadata is persisted until traffic pauses. The coalescing test misses this because it calls the test-only forced flush.

Please add a max-wait or bounded-size flush policy and a production-timer regression proving sustained enqueue traffic still flushes.

The retry path also needs correctness fixes before merge:

- If an earlier chunk commits and a later chunk query throws, the queue requeues the entire batch and applies the committed access deltas twice.
- A partially successful `add()` leaves preserved updated rows in place but reports every row as failed, so those successful rows are retried and incremented again.
- Deferred flushes unconditionally zero `bad_recall_count` and suppression fields, which can overwrite a newer governance decision made after the recall event.
- The unref'd timer has no production shutdown drain, so the final pending/retrying window can be silently discarded.

Please add partial-commit, concurrent-governance, retry-backoff, and shutdown-drain regressions using the production queue path. The branch is also behind the current base, so rerun the suite after rebasing the stacked changes.

app3apps (changes requested):
Re-reviewed head `d9bcc5f`. The targeted and full suites pass, and the max-wait, partial-commit, retry, and shutdown-test coverage are substantially improved.

One data-integrity path still blocks merge: `applyManualRecallMetadataBatch` replaces rows by deleting them before `add()`. If `add()` fails and the recovery/rollback reads also fail, the complete memory rows are gone, the items are classified as non-retryable, and the queue payload does not retain the text/vector data needed to restore them. A correlated transient store failure can therefore permanently delete memories while applying recall metadata.

Please make the replacement atomic, or retain durable restoration material and retry uncertain outcomes. Add a fault-injection regression covering the `add()` plus recovery/rollback double-failure path and prove the original rows survive.

The production queue also still needs an explicit exit drain for the unref'd fire-and-forget window, and the governance snapshot equality check can silently skip a suppression reset when an unrelated governance field changes during the flush window. Please close those lifecycle/concurrency gaps as part of the fix.

Requesting changes.

app3apps (commented):
Re-reviewed the current head `ac0cfda`. The new atomic `mergeInsert` path resolves the previous delete-before-add data-loss blocker, the queue now has production drain coverage, the focused/full suites pass, and `npm run build` completes without source/dist drift.

I am not approving this head yet because the PR is still intentionally stacked on open PR #926: its base is the older `7791bd3` line and the diff still includes the #926 predecessor commit. Please rebase onto current `master` after #926 lands (or remove the stacked commit), confirm that this PR contains only the #936 follow-up, rerun verification, and request review again.

For the final pass, please also call out the two deliberate semantics explicitly: metadata queued in the debounce window can still be lost on abrupt exits that bypass `beforeExit`/`destroy`, and a concurrent governance change causes the entire suppression reset to be skipped. Neither negates the atomic-write fix, but both need maintainer sign-off on the isolated diff.

app3apps (changes requested):
Re-reviewed head `c1f010d`. The atomic `mergeInsert` path still resolves the earlier delete-before-add data-loss blocker, and the build, focused checks, full suite, and GitHub CI are green.

One new merge-resolution issue blocks this head: `scripts/ci-test-manifest.mjs` removes four existing regression suites while adding the two queue suites:

- `test/autocapture-internal-session-guard.test.mjs`
- `test/memory-categories-storage-map.test.mjs`
- `test/delete-invalidate-reflection-caches.test.mjs`
- `test/reflection-mapped-rows-admission.test.mjs`

Those test files still exist, so merging this branch would silently stop CI from running coverage for four shipped fixes. Please restore all four manifest entries and remove the unrelated line-ending churn.

The branch also still contains the #926 predecessor commits. Once #926 lands, please rebase so this PR contains only the #936 follow-up, then rerun the checks. Requesting changes.

TurboTheTurtle:
Restored the dropped manifest entries and normalized the manifest line endings.

Focused checks passed: `node --test test/store-write-queue.test.mjs test/manual-recall-metadata-queue.test.mjs`, plus a manifest import check for the restored entries.

rwmjhb:
Checked head `bd45fad`. The restored CI-manifest entries and the targeted/full test runs are green, but GitHub still reports `mergeable=CONFLICTING` / `merge_state_status=DIRTY`, and the branch is behind current `master`. Deep review is therefore deferred until the resulting diff is stable.

Please rebase onto current `master` (or merge it into this branch), resolve the conflicts, keep the #936 follow-up isolated from already-landed predecessor work where possible, rebuild `dist`, and rerun the full suite before pushing the resolved head. We will pick up the new commit for re-review.

TurboTheTurtle:
Merged current master to clear the conflict and kept the #936 queue changes isolated on top. Focused checks passed: `node --test test/store-write-queue.test.mjs test/manual-recall-metadata-queue.test.mjs test/recall-text-cleanup.test.mjs`, `node scripts/verify-ci-test-manifest.mjs`, and `npm run build`.

rwmjhb (changes requested):
Reviewed head `efae8bc`. The conflict is resolved, and coalescing plus a lock-scoped batch write is the right direction for the observed concurrency issue. One cross-instance lost-update path still blocks merge.

`applyManualRecallMetadataBatch` acquires the shared write lock but then reads through the existing LanceDB table handle without refreshing it. With the default undefined `readConsistencyInterval`, a second long-lived `MemoryStore` can remain pinned to an older table version even after it obtains the lock. In a two-store reproduction, both serialized one-delta batches returned success, but a strongly consistent reader observed final `access_count=1` rather than 2. Please refresh the table to its latest version after acquiring the lock and before reading, and add a two-store regression that asserts both deltas survive.

There is also a retry-generation bug: when an older same-ID update fails, `mergePending` retains its exhausted `attempts` count while merging a fresh pending event. The new event can therefore be dropped after only one write attempt. Please keep fresh deltas on an independent retry budget and add an enqueue-during-retry regression.

The verifier initially timed out the two focused files, but an exact-head rerun passed all 7 tests, so I am not treating that artifact as a reproducible product failure. Requesting changes for the two code paths above.

TurboTheTurtle:
Pushed a fix for both remaining paths:

- refresh the LanceDB table to latest after the write lock is held, before reading rows for the metadata merge
- keep fresh same-ID enqueue traffic from inheriting an older failed retry generation

Checks passed: `node --test test/store-write-queue.test.mjs test/manual-recall-metadata-queue.test.mjs test/recall-text-cleanup.test.mjs`, `node scripts/verify-ci-test-manifest.mjs`, and `npm run build`.

rwmjhb (changes requested):
Re-reviewed head `8c96d63`. Refreshing the table after acquiring the shared write lock fixes the previously reported two-store lost update, and `npm run build` completes with no `dist` drift. One ambiguous-write recovery path still blocks merge.

`classifyUncertainMerge()` (`src/store.ts:2358`) queries the existing table handle after `mergeInsert()` reports an uncertain error, but it does not refresh that handle again before classifying the outcome. If another connection committed the delta while this handle remains on the older LanceDB version, recovery sees the original metadata, marks the row retryable, and the queue applies an already committed delta a second time. An exact-head fault injection reproduced `access_count=2` for a single intended increment.

Please refresh to the latest table version inside the uncertain-outcome recovery path before querying. If that refresh fails, treat the result as unverifiable/non-retryable rather than risking a duplicate application. Please also add an independent-handle post-commit-error regression proving the committed row is recognized and not retried.

Verification note: the exact-head build passed and 20/21 focused queue/store tests passed. `process beforeExit catches same-store work across repeated later-listener microtasks` failed both in the focused run and when run alone: its child was terminated by the hard-coded 10 s timeout, and the isolated parent test exited after about 43 s. The broader suite also exceeded the orchestrator's 180 s window without an assertion failure before timeout. Please stabilize or explain the isolated exit-drain failure as part of the update.

Requesting changes for the stale recovery classification, which can double-count recall metadata.

TurboTheTurtle:
Pushed a recovery fix for the ambiguous merge path: it now refreshes the table again before checking whether a post-commit error actually persisted the row, so a stale long-lived handle won't requeue an already-applied delta. I also added the independent-handle post-commit regression; focused queue/store tests, the storage-and-schema group, manifest check, and build all passed.

rwmjhb (approved):
Re-reviewed head `a568b11`. The uncertain-commit recovery path now refreshes the LanceDB table before classifying the outcome, closing the previously reproduced double-application blocker. GitHub CI is green, the orchestrator's focused and full verification passed, and the six-round review found no merge-blocking issue.

Two follow-ups are worth retaining: fresh same-ID recalls can reset the retry count of an older failed generation under sustained failure, and an unexpected per-row vector conversion error can requeue a whole batch after earlier chunks committed. Both are low-likelihood/best-effort metadata paths rather than blockers for this fix, but generation-separated retries and per-row classification would make the queue more robust. The branch history is also stale/stacked, so please preserve the isolated #936 follow-up and rerun checks if the head is rebased before merge.

Approved.

PR #938, opened 2026-07-13
fix: split the memory pipeline's internal LLM prompts into system and user messages

gorkem2020:
The memory pipeline makes four internal LLM calls: extraction, admission scoring, dedup decision, and merge. All four packed instructions, criteria, output-format contract, and the per-call data into a single user-role message, with no system prompt beyond a generic one-liner shared by every call ("You are a memory extraction assistant..."). This meant the admission judge, the dedup decider, and the merge writer all introduced themselves with extraction-assistant wording, and every call mixed static instructions in with the untrusted conversation excerpt in the same message.

## What changed

Each of the four prompt builders (`buildExtractionPrompt`, `buildDedupPrompt`, `buildMergePrompt` in `src/extraction-prompts.ts`, and `buildUtilityPrompt` in `src/admission-control.ts`) now returns `{system, user}`: static instructions, criteria, identity, and the output-format contract live in `system`; the per-call conversation excerpt, candidate rows, or neighbor rows live in `user`.

Each stage gets an honest, distinct identity instead of sharing generic wording:
- extraction: "You are an extraction agent."
- admission scoring: "You are an admission judge."
- dedup: "You are a dedup decider."
- merge: "You are a merge writer."

`LlmClient.completeJson()` gains an optional third argument, `systemPrompt`. The api-key path threads it into the `messages` array's system role. The OAuth path threads it into the Responses API's existing `instructions` field; that field already carried system-prompt semantics (it previously held the same generic string), so no degrade-to-prepend fallback was needed; the request shape stays exactly as it was, just with a per-call value instead of one fixed string. Both paths default to the historical generic text when `systemPrompt` is omitted, so the one caller outside this pipeline (`memory-upgrader.ts`'s one-off migration prompt) is unaffected.

The merge prompt template had a live-observed formatting bug: broken bold markers rendered as `** Category **:`, progressively indented list items, and a tokenization artifact ("up - to - date" from a space-hyphen-space pattern in the source). Rebuilt the template cleanly; added a snapshot-style test asserting the assembled prompt contains none of those artifacts.

Reflection-sourced admission gating previously reused the extraction path's "Conversation excerpt:" framing even when the excerpt was actually a reflection document. `buildUtilityPrompt` and `AdmissionController.evaluate()` now take an optional `sourceKind: "conversation" | "reflection"` (default `"conversation"`), and the admission judge sees "Source document (agent reflection):" for reflection-sourced rows instead. The admission judge only ever scores the excerpt as given; it never rewrites candidates.

No scoring weights, thresholds, or the decision protocol changed. Same decisions in and out; better-formed prompts.

## Test plan

- New test file covering all four builders' system/user split, per-stage identity strings, the merge template's absence of broken-bold/indentation/tokenization artifacts, and the reflection source-kind framing.
- `LlmClient` tests confirming the api-key path's system message and the OAuth path's `instructions` field both honor an explicit `systemPrompt` and fall back to the historical default when omitted.
- Updated three existing test files whose mock LLM servers matched on the old prompt-opener phrases (now identity phrases, since those phrases moved to the system message).
- Full local test chain green; `tsc` build green.


gorkem2020:
Closing as superseded: current master ships this refactor's content. src/prompt-blocks.ts and src/llm-client.ts on master are byte-identical to this branch, buildUtilityPrompt returns the split { system, user } shape (master also added a batch variant via [#941](https://github.com/CortexReach/memory-lancedb-pro/pull/941)), and the extraction, dedup, merge, and reflection call sites all pass split prompts. Where prompt wording still differs from this branch, master's current form is the deliberately chosen one. Nothing left to merge.

PR #939, opened 2026-07-13
feat: add a captureAssistant context mode for non-extractable assistant context

gorkem2020:
`captureAssistant: false` (the default) filters assistant turns out of extraction input entirely, so only user-authored text is ever eligible. That protects against misattributing an assistant's own words as a user fact, but it also starves the extractor of context it needs to disambiguate a reply like "yes exactly, that one"; the extractor has no idea what "that" refers to.

## What changed

Added a middle mode: `captureAssistant: "context"`. Assistant turns are included in the extraction prompt as clearly marked, non-extractable context, while remaining fully capture-ineligible.

Chosen as a union literal on the existing field (`captureAssistant?: boolean | "context"`) rather than a sibling boolean, since it composes directly with the existing normalization (`cfg.captureAssistant === "context" ? "context" : cfg.captureAssistant === true`) and needed no new config surface. `true` and `false` keep byte-identical semantics; verified directly against `parsePluginConfig`.

Assistant-context texts are collected into their own array during the auto-capture message loop, never merged into `eligibleTexts`. They're rolled across turns in a new `autoCaptureRecentAssistantTexts` map (mirroring the existing `autoCaptureRecentTexts` pattern) and cleared on successful extraction. They never touch `autoCaptureSeenTextCount`, so extraction eligibility and the auto-capture watermark are unaffected by this mode; a hook-level regression test proves cumulative counting across two user+assistant turns stays 1-then-2, not 2-then-4, when assistant turns are present.

`SmartExtractor.extractAndPersist()` gains an `assistantContextTexts` option threaded down to prompt assembly, which appends the marked lines under an "Assistant Context" heading. The extraction prompt's criteria gained a matching rule: a candidate may only assert facts sourced from a user-authored line, never from an assistant-marked line alone. This is prompt-level guidance only; there's no deterministic storage-side gate for it (unlike the grounding work's constructed-tag drop), so test coverage here is structural (the instruction text is present) plus the prompt-assembly behavior (marked lines appear only when `assistantContextTexts` is non-empty).

## Test plan

- Config back-compat: `parsePluginConfig` normalizes `true`/`false`/omitted exactly as before, and `"context"` passes through as the literal.
- `SmartExtractor`-level: marked assistant lines appear in the assembled prompt when `assistantContextTexts` is provided, and are absent when omitted or empty.
- Hook-level regression test: two turns each carrying a user+assistant pair, `captureAssistant: "context"`, `extractMinMessages: 2`; cumulative eligibility count goes 1-then-2 (not 2-then-4), while the extraction prompt on the triggering turn does contain the assistant's marked text.
- Full local test chain green; `tsc` build green.

## Update (2026-07-18)

Rebased onto current master. This revision also lands the pair-shaped extraction window: `extractMinMessages` now bounds the transcript in user/assistant PAIRS (newest N user turns with their replies interleaved in true order), `captureAssistant` context rides the same window instead of an independent 6-slot assistant carry, and consumed pairs drop together. The extraction prompt's assistant-lines rule gained a live-caught counterexample (the assistant greeting the user by name is not the user introducing themselves). Verified live on our fleet: each turn's extraction transcript is exactly its own pair.


gorkem2020:
Superseded by the context-window stack. The goal of the "context" mode (assistant turns as non-extractable context) landed as the rolling context window in [#966](https://github.com/CortexReach/memory-lancedb-pro/pull/966) and [#967](https://github.com/CortexReach/memory-lancedb-pro/pull/967), with captureAssistant staying a plain boolean; the turns plumbing this draft introduced is carried by [#964](https://github.com/CortexReach/memory-lancedb-pro/pull/964). Closing in favor of the new stack.

PR #940, opened 2026-07-13
fix: construct admission control independently of smart extraction

gorkem2020:
Admission control could silently do nothing when `smartExtraction: false`, even with `admissionControl.enabled: true`. The `AdmissionController` was only ever constructed inside the `smartExtraction` init block, so any write path other than `SmartExtractor` had no controller to use and stored content ungated.

This decouples admission-controller construction from the extraction engine choice, so its availability depends only on `admissionControl.enabled`.

## Changes

- `src/admission-control.ts`: adds `createAdmissionController(store, llm, config, debugLog)`, a factory that builds a controller purely from config, independent of any extractor.
- `src/smart-extractor.ts`: `SmartExtractor` no longer builds its own `AdmissionController` from config. It now accepts a pre-built controller via a new `admissionController` constructor option and uses whatever it is given (or none).
- `index.ts`:
  - The LLM client (and now the admission controller) is constructed whenever `smartExtraction !== false` **or** `admissionControl.enabled === true`, instead of only the former.
  - The standalone controller is exposed on the plugin's singleton state alongside `smartExtractor`, so other write paths (for example a reflection-mapped-rows gate, or a regex-fallback gate) can reference it directly instead of reaching through `smartExtractor`.
  - Extends the existing `plugin registered` log line with an `admissionControl: ON|OFF` marker, making the fix directly observable.

## Behavior

- `smartExtraction: false`, `admissionControl.enabled: true` → a controller is now constructed and available.
- `smartExtraction: false`, `admissionControl.enabled: false` → still fully off, content still stored ungated. That is the configured meaning of "off," unchanged.
- `smartExtraction: true` → decisions and audits are unchanged; the controller is built from the exact same store/llm/config/debugLog inputs it always was, just at a different call site.

## Tests (all new, TDD red-first)

- `test/admission-controller-standalone.test.mjs`: `createAdmissionController` returns `null` when disabled, and a working controller when enabled.
- `test/smart-extractor-admission-controller-injection.test.mjs`: `SmartExtractor` calls whatever controller it is given, and skips gating entirely when none is configured (today's off-behavior, unchanged).
- `test/admission-without-smart-extraction.test.mjs`: registers the plugin with `smartExtraction: false` + `admissionControl.enabled: true` and asserts the new log marker; also covers the fully-off and `smartExtraction: true` cases as regression guards.

Full local test chain green (`npm test`, minus the pre-existing host-Ollama port conflict on 11434, unrelated to this change) and `npm run build` green.


gorkem2020:
Closing as superseded on the functional side: since [#934](https://github.com/CortexReach/memory-lancedb-pro/pull/934) merged, master builds a fallback AdmissionController when smart extraction is disabled or fails to initialize and routes the regex-fallback and mapped-reflection paths through it, so admission gating already works independently of smart extraction. The remaining value here is the single-construction-site shape (standalone factory plus injection into SmartExtractor), and that refactor will ride the lane-affinity PR [#942](https://github.com/CortexReach/memory-lancedb-pro/pull/942), which needs the same construction seam. The test coverage from this branch moves there with it.

PR #942, opened 2026-07-13
feat: per-lane admission model affinity with an explicit model override

gorkem2020:
## What

Adds a per-lane model affinity option to admission control, plus an explicit absolute override, and normalizes model references on the direct-client path so both knobs resolve provider-prefixed and bare model names identically.

- `admissionControl.modelAffinity: "global" | "lane"` (default `"global"`, no behavior change when unset). With `"lane"`, the reflection-mapped admission gate resolves the `memoryReflection` model instead of the plugin's global LLM model, so the judge auditing reflection-authored rows is never weaker than the author that wrote them. Extraction and fallback admission stay on the global model.
- `admissionControl.model`: absolute override. When set, every admission LLM call uses this model regardless of lane.
- `normalizeDirectModelRef()`: shared normalization for direct-client model references, used by both the admission model resolution and the host-to-direct fallback path, so one definition decides what a direct client accepts.
- The admission controller is constructed once and injected into the smart extractor (single construction site, one shared controller), keeping availability and audit behavior identical to the current fallback path. Proven by the untouched baseline suite passing as-is.
- `llm.thinkLevel` is threaded through the per-model client factory so lane-resolved clients keep the configured reasoning effort.

## Why

Reflection-mapped rows are distilled by the (typically stronger) `memoryReflection` model, but their admission verdicts previously came from the global memory model, which on cost-tiered setups is the weakest model in the plugin. A weak judge auditing a strong author rejects well-formed durable rows on shallow criteria. Lane affinity fixes the mismatch without forcing a global model upgrade.

## Notes

- Rewritten from scratch on current master after the admission/batch groundwork landed via [#941](https://github.com/CortexReach/memory-lancedb-pro/pull/941); the earlier draft head predated those merges. Single commit, based on the current tip.
- Off by default: `modelAffinity` unset or `"global"` resolves exactly as today, pinned by tests (override wins on every lane, one shared controller).
- Tests: lane resolution matrix, model-reference normalization, standalone controller construction, extractor injection seam, admission-without-smart-extraction parity, and thinkLevel threading (27 assertions across six files, all registered in the package.json chain and the CI manifest).


gorkem2020:
Scope restored in db661f7: llm.thinkLevel now ships with this PR. The reasoning-effort knob rides the direct client (OpenRouter-compatible reasoning effort field, sent only when explicitly configured; unset leaves the provider default), and lane affinity carries the full pair: modelAffinity "lane" runs the reflection pipeline on the reflection lane's model AND thinkLevel, "global" keeps every stage on the plugin llm and its thinkLevel. Wire-level regressions cover configured, unconfigured, and blank-value resolution. The host-transport half of the original delta remains in the transport PR ([#947](https://github.com/CortexReach/memory-lancedb-pro/pull/947)), which is where the reasoning field also needs forwarding through the host runtime.

rwmjhb (approved):
Reviewed head `71cce8f`. The opt-in reflection-lane model affinity and absolute override preserve the default global behavior, the primary plugin path injects the intended controller, and the targeted tests, full local suite, and all GitHub CI checks pass.

Non-blocking follow-ups: translate `memoryReflection.thinkLevel: "off"` to the direct API's `"none"` value and validate wire values; document or support the OAuth transport behavior; preserve `llm.thinkLevel` through auth login/logout; normalize the primary global direct-client model reference as well as admission clients; keep direct `SmartExtractor` construction from silently ignoring enabled admission; and move `test/llm-thinklevel.test.mjs` from `CI_TEST_GROUPS` into `CI_TEST_MANIFEST`.

Approved.

PR #943, opened 2026-07-13
feat: mirror reflection slice rows into the daily memory journal

gorkem2020:
The mdMirror daily journal (`memory/<date>.md`) mirrors writer-1 reflection mapped rows, smart-extraction rows, and manual `memory_store` rows, but not the reflection writer-2 slice rows (`storeReflectionToLanceDB`'s event / invariant / derived rows). Those rows entered the database invisibly; the daily file only got the "Reflection generated: `<path>`" pointer line. This PR closes that gap so every row that enters the database is observable in the daily file without CLI tools.

- `storeReflectionToLanceDB` (`src/reflection-store.ts`) gains an optional `onPersisted(entry, kind)` hook, fired after each `event` / `item-invariant` / `item-derived` row is stored. The `combined-legacy` row is intentionally excluded since it is out of scope for this change (it is a distinct legacy row type, not one of the writer-2 slice rows this PR targets).
- Callback failures are swallowed with a try/catch so a mirror bug can never abort reflection persistence, matching the existing `SmartExtractor.notifyPersisted` convention for the same kind of hook.
- `index.ts` wires `mdMirror` through this hook with source labels `reflection-slice:event`, `reflection-slice:invariant`, and `reflection-slice:derived`.
- The event row's text is a multi-line `key=value` block. It renders as one line for free, because `createMdMirrorWriter`'s existing line formatter already replaces `\n` with a space before appending to the daily file - no changes were needed there.

## Test plan

- [x] Red-first: added tests to `test/memory-reflection.test.mjs` asserting `onPersisted` fires for event/invariant/derived rows, is skipped for the legacy row, does not break storage when absent, does not abort remaining stores if it throws, and that the event row's multi-line text collapses to one line under the existing mirror formatting contract. All five failed before the implementation and pass after.

## Notes for reviewers

While working on this I found that `test/memory-reflection.test.mjs` itself (the file with the most direct coverage of `storeReflectionToLanceDB`) is not wired into either `package.json`'s `test` script or `scripts/ci-test-manifest.mjs` on master, and separately has four pre-existing failing tests unrelated to this change. Both are out of scope here and are called out separately rather than folded into this PR.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -4071,6 +4071,14 @@ const memoryLanceDBProPlugin = {
                             embedPassage: (text) => embedForReflectionPersistence(text, "slice-embedding"),
                             vectorSearch: (vector, limit, minScore, scopeFilter) => store.vectorSearch(vector, limit, minScore, scopeFilter),
                             store: (entry) => store.store(entry),
+                            onPersisted: mdMirror
+                                ? async (entry, kind) => {
+                                    const source = kind === "event" ? "reflection-slice:event"
+                                        : kind === "item-invariant" ? "reflection-slice:invariant"
+                                            : "reflection-slice:derived";
+                                    await mdMirror({ text: entry.text, category: entry.category, scope: entry.scope, timestamp: entry.timestamp }, { source, agentId: sourceAgentId });
+                                }
+                                : undefined,
                         });
                         if (sessionKey && stored.slices.derived.length > 0 && !isSessionBoundaryReflectionAction(action)) {
                             reflectionDerivedBySession.set(sessionKey, {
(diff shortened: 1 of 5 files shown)


rwmjhb (approved):
Reviewed head `a32fdd0`. The targeted tests, full suite, and repository CI are green, and the production change correctly mirrors successfully persisted invariant/derived reflection slices without coupling database persistence to mirror failures. I found no HIGH or CRITICAL correctness issue.

Non-blocking follow-ups: align the PR contract with the intentional event-row exclusion; register an integration test that exercises the real `index.ts -> mdMirror` wiring; and consider observability/backpressure for repeated mirror failures or large slice batches. Approving.

PR #945, opened 2026-07-13
fix: materialize typed-array vectors on store fetch paths

gorkem2020:
LanceDB returns vector columns as typed arrays (Float32Array) or Arrow vector objects, not plain JavaScript arrays. Three row-fetch paths in `src/store.ts` assumed otherwise:

- `fetchForCompaction` used `Array.isArray(row.vector) ? row.vector : []` to decide whether to keep the vector. Since `Array.isArray` is always false for a typed array, every fetched vector silently collapsed to an empty array.
- `bm25Search` and `lexicalFallbackSearch` used a bare `row.vector as number[]` cast, which does not actually convert anything: the returned `MemoryEntry.vector` is a typed array pretending to be `number[]` at the type level, but not a real array at runtime.

This went unnoticed because the one existing consumer of `fetchForCompaction`, the background compactor, ignores the returned vectors entirely (its own clustering step never reads them). The bug surfaced through a downstream consolidation command that clusters rows by embedding similarity and got zero clusters back from a live dry-run, because every vector it saw had length zero.

`getById` and the other row-mapping call sites in this file were already unaffected, since they use `Array.from(row.vector as Iterable<number>)`, which correctly converts a typed array via iteration.

## Fix

All three affected sites now use the existing `toNumberVector` helper (already the pattern the primary `vectorSearch` path relies on), which accepts any array-like input (typed arrays, Arrow vectors, plain arrays), converts every element with `Number()`, and validates the result is entirely finite before returning it, falling back to `[]` only for genuinely absent or malformed vectors.

## Test plan

- [x] Red-first: `test/typed-array-vector-fetch.test.mjs` stores a real vector into a temporary LanceDB table and asserts each of the three affected paths returns it intact. All three failed against the current code (`fetchForCompaction` returned a zero-length vector; `bm25Search` and `lexicalFallbackSearch` returned a non-Array value) and passed after the fix.
- [x] `test/typed-array-vector-fetch.test.mjs` registered in both `package.json`'s `test` script and `scripts/ci-test-manifest.mjs` (storage-and-schema group); manifest verification passes.


rwmjhb (approved):
Reviewed head `5bf3f2d`. The three affected fetch paths now use the established `toNumberVector` runtime normalization instead of type-only casts or `Array.isArray` fallback behavior. The real-LanceDB regression tests fail against the pre-fix implementation and pass on this head for compaction, BM25, and lexical fallback; the full suite also passes.

A fresh compile produced `dist/src/store.js` byte-identical to the committed artifact, and direct packaged-runtime probes returned ordinary JavaScript arrays from all three paths. No correctness blockers found. Approving.

PR #948, opened 2026-07-13
fix: reset the auto-capture watermark after valid-empty extractions

gorkem2020:
The agent-end auto-capture watermark (`autoCaptureSeenTextCount`) is reset after a successful extraction, but a valid-empty extraction (the LLM legitimately returns zero candidates) returns early and skips that reset. The counter then carries a stale, too-high value into the next turn.

In ingress-fed sessions (`message_received` feeds the pending-texts queue and each agent end delivers only the newest messages) this makes the very next turn fire extraction prematurely on a single genuinely new message, instead of waiting for a fresh `extractMinMessages`-sized window. In history-carrying sessions the asymmetry happens to be a no-op (the cumulative count is recomputed from the full history each turn), which is why it went unnoticed.

## What changed

The valid-empty path now applies the identical watermark reset as the successful path. Both cases represent a batch the extractor genuinely and conclusively considered; only errored extractions leave the counter untouched.

## Test plan

New regression test reproduces the exact ingress-fed sequence with the real plugin harness: a three-text turn whose extraction returns valid-empty, then one new message (must defer), then a second new message (fires normally). Watched red against the pre-fix code (the one-message turn fired prematurely), green after. The existing watermark and fallback-gating suites pass unchanged.

Found while live-testing the runtime transport work; the defect predates all current feature branches and reproduces on plain master, so it ships on its own branch.


gorkem2020:
Closing as superseded: master addressed the same problem (texts stuck behind the auto-capture watermark after a valid-empty extraction) with a different landed design: barren texts are deferred for a later sweep (deferBarrenExtractionTexts) instead of the watermark resetting past them. The two semantics are incompatible, and this branch's own consume-contract test fails against master's intended behavior, so merging it now would regress the landed design rather than extend it.

PR #949, opened 2026-07-14
fix(llm-client): disable reasoning per-model via baseURL-aware table

### LINKED ISSUE #929: [Bug] shouldDisableReasoningForJson regex doesn’t match deepseek-v4-flash → wastes ~88-94% of output tokens on reasoning trace
## Title
[Bug] `shouldDisableReasoningForJson` regex doesn't match `deepseek-v4-flash` → wastes ~88-94% of output tokens on reasoning trace

### Summary

`shouldDisableReasoningForJson` in `src/llm-client.ts` (line 102-103) currently uses the regex:

```ts
return /qwen3|deepseek.*r1|qwq/i.test(model);
```

This regex does **not** match `deepseek-v4-flash` (the `v4` family ships without an explicit `r1` suffix). When the model is configured for JSON extraction in `smart-extractor`, the LLM client therefore does **not** inject the `chat_template_kwargs: { enable_thinking: false }` flag, and DashScope's `deepseek-v4-flash` returns the answer only in the `reasoning_content` field with `content` empty.

PR #914 added a graceful fallback (`recoverJsonFromReasoning`) that rescues the JSON from the reasoning trace, so extraction still succeeds — but the regex still leaves the door open for downstream inefficiency:

### Impact (measured)

Using `DashScope` OpenAI-compatible endpoint with model `deepseek-v4-flash` and 6-category extraction prompt:

| Metric | Value |
|---|---|
| Tokens in prompt | ~3,200 |
| `reasoning_content` tokens (per call) | **~2,800** |
| `content` tokens (per call) | ~50 |
| **Reasoning share of output** | **~88–94%** |
| Estimated monthly waste on 200 extractions/day | ~150M reasoning tokens |

Since `deepseek-v4-flash` accepts `chat_template_kwargs: { enable_thinking: false }` on DashScope (verified end-to-end), this is pure waste — the model does have a non-thinking mode and the client just isn't asking for it.

### Reproduction

1. Set `llm.model` to `deepseek-v4-flash` (DashScope OpenAI-compatible endpoint).
2. Trigger `autoCapture` / smart-extractor with a non-trivial conversation.
3. Inspect gateway logs / LLM request payload — note that `chat_template_kwargs` is **not** present in the request body.
4. Inspect LLM response — `choices[0].message.content` is empty or near-empty; the JSON lives inside `reasoning_content`. (PR #914's `recoverJsonFromReasoning` rescues it, so extraction still works, but at the cost above.)

### Expected

The regex should treat the `deepseek-v*` family (non-reasoning variants) the same way it treats `deepseek.*r1` — i.e., inject `enable_thinking: false` so the model emits JSON directly into `content`.

### Suggested fix

One-line patch in `src/llm-client.ts`:

```diff
- return /qwen3|deepseek.*r1|qwq/i.test(model);
+ return /qwen3|deepseek.*(r1|v\d)|qwq/i.test(model);
```

This makes the regex match `deepseek-v4-flash`, `deepseek-v4-pro` (the pro variant also accepts `enable_thinking: false` on DashScope), and any future `deepseek-vN-*` releases, while keeping the existing matches for `deepseek-r1`, `qwen3`, and `qwq`.

### Workaround (in use locally)

Currently working around by configuring DashScope's `deepseek-v4-flash` directly and accepting the reasoning-token waste until the regex is fixed upstream. `recoverJsonFromReasoning` from PR #914 prevents the worst-case extraction failure, so this is a performance/cost issue rather than a correctness issue.

### Related

- PR #914 — `recoverJsonFromReasoning` (merged; rescues JSON but doesn't disable reasoning)
- Issue #918 — separate concern (memory injection causing repetition loops on mimo-v2.5); the maintainer comments there list `deepseek-v4-flash` as a "non-reasoning model that works fine", which is exactly why I think this fix is safe — `v4-flash` is intended to be a non-thinking model.

### Environment

- `memory-lancedb-pro` @ `e0cd6e8` (master, latest)
- DashScope OpenAI-compatible endpoint: `https://dashscope.aliyuncs.com/compatible-mode/v1`
- Model: `deepseek-v4-flash`
- Symptom severity: medium (cost/perf, not correctness — extraction still succeeds via PR #914's fallback)

DerekEXS:
Extends the single-line `/qwen3|deepseek.*r1|qwq/` regex (issue #929) into a baseURL-aware rule table that injects the correct disable-thinking request parameter for each known Chinese reasoning-model family on its respective provider endpoint.

## Why a table (not just a regex fix)

The same model can require different parameter shapes on different providers. For example, `deepseek-v4-flash` uses `chat_template_kwargs.enable_thinking=false` on Aliyun DashScope but `thinking.type="disabled"` on DeepSeek's own API. Model name alone is not enough — the table selects the right parameter per `(model family, provider)`.

## Coverage (parameters from each provider's official docs)

| Family | Provider | Parameter |
|---|---|---|
| Qwen3 / QwQ | DashScope / vLLM | `chat_template_kwargs.enable_thinking=false` |
| DeepSeek V4 / R1 | DashScope | `chat_template_kwargs.enable_thinking=false` |
| DeepSeek V4 | DeepSeek official | `thinking.type="disabled"` |
| GLM-4.6 / GLM-Z1 | Zhipu BigModel | `thinking.type="disabled"` |
| Doubao thinking/seed | Volcano Ark (OpenAI-compat) | `thinking.type="disabled"` |
| Kimi K1.5 / K2 | Moonshot | `thinking.type="disabled"` |
| MiniMax M2 / abab | MiniMax | `thinking_budget=0` |
| Hunyuan T1 / Turbo-S | Tencent | `enable_thinking=false` |
| StepFun step-3 / step-2 | StepFun | `thinking=false` |

## API

- New `buildDisableReasoningParams(model, baseURL)` returns the correct params or `null`.
- `shouldDisableReasoningForJson(model)` is kept as a backward-compatible boolean wrapper (used by existing tests).
- `recoverJsonFromReasoning` (PR #914) remains as a safety net for models/providers not yet in the table.

## Verification

Live test against DashScope `deepseek-v4-flash` with the extraction prompt, three variants:

| Variant | Reasoning chars | Completion tokens | Total |
|---|---|---|---|
| (baseline) no param | 225 | 67 (incl. 53 reasoning) | 102 |
| `chat_template_kwargs.enable_thinking=false` | **0** | **12** | **47** |
| `thinking.type="disabled"` | **0** | **12** | **47** |

Completion tokens -82%. DashScope `deepseek-v4-flash` accepts both parameters. Only this one model lane was live-verified; the rest are doc-sourced and rely on `recoverJsonFromReasoning` as a safety net for unverified models/providers.

## Tests

`test/llm-api-key-client.test.mjs` — all 8 pass (6 original + 2 new: per-provider param table, `deepseek-v4-flash` end-to-end injection).

Fixes #929

### DIFF
--- a/dist/src/llm-client.js
+++ b/dist/src/llm-client.js
@@ -71,8 +71,100 @@ function recoverJsonFromReasoning(reasoningText) {
         return null;
     }
 }
+const REASONING_DISABLE_RULES = [
+    {
+        // Qwen3 / QwQ — DashScope chat_template_kwargs (verified). vLLM-based
+        // aggregators (SiliconFlow, etc.) follow the same convention.
+        modelPattern: /qwen3|qwq/i,
+        providers: [
+            { baseURL: "default", params: { chat_template_kwargs: { enable_thinking: false } } },
+        ],
+    },
+    {
+        // DeepSeek V4 / R1 family.
+        //  - DashScope (Aliyun) hosts deepseek-v4-flash and uses chat_template_kwargs
+        //    (verified, issue #929 — reasoning drops to 0, completion tokens -82%).
+        //  - DeepSeek's own API (api.deepseek.com) uses thinking.type=disabled for
+        //    v4-pro/v4-flash. deepseek-reasoner (R1) on the official API is a pure
+        //    reasoning model and cannot disable thinking.
+        modelPattern: /deepseek.*(v\d|r1)/i,
+        providers: [
+            { baseURL: /dashscope|aliyun/i, params: { chat_template_kwargs: { enable_thinking: false } } },
+            { baseURL: /deepseek\.com/i, params: { thinking: { type: "disabled" } } },
+            { baseURL: "default", params: { chat_template_kwargs: { enable_thinking: false } } },
+        ],
+    },
+    {
+        // Zhipu GLM-4.6 / GLM-Z1 — thinking.type=disabled (docs.bigmodel.cn).
+        modelPattern: /glm.*(4\.6|z1)|glm.*thinking/i,
+        providers: [
+            { baseURL: "default", params: { thinking: { type: "disabled" } } },
+        ],
+    },
+    {
+        // Volcano Ark doubao thinking/seed models — thinking.type=disabled on the
+        // OpenAI-compatible /api/v3/chat/completions endpoint (docs.volcengine.com).
+        // Non-thinking variants (doubao-1.5-pro, etc.) are intentionally excluded.
+        modelPattern: /doubao.*(thinking|seed)/i,
+        providers: [
+            { baseURL: "default", params: { thinking: { type: "disabled" } } },
+        ],
+    },
+    {
+        // Moonshot Kimi K1.5 / K2 — thinking.type=disabled (platform.kimi.com).
+        // Note: kimi-k2.7-code cannot disable thinking.
+        modelPattern: /kimi.*k[12]|kimi.*thinking/i,
+        providers: [
+            { baseURL: "default", params: { thinking: { type: "disabled" } } },
+        ],
+    },
+    {
+        // MiniMax M2 / abab — thinking_budget=0 (api.minimax.chat).
+        modelPattern: /minimax.*m2|abab/i,
+        providers: [
+            { baseURL: "default", params: { thinking_budget: 0 } },
+        ],
+    },
+    {
+        // Tencent Hunyuan T1 / Turbo-S — enable_thinking=false (cloud.tencent.com).
+        modelPattern: /hunyuan.*(t1|turbo)/i,
+        providers: [
+            { baseURL: "default", params: { enable_thinking: false } },
+        ],
+    },
+    {
+        // StepFun step-3 / step-2 — thinking=false (developer.stepfun.com).
+        modelPattern: /step-[32]/i,
+        providers: [
+            { baseURL: "default", params: { thinking: false } },
+        ],
+    },
+];
+/**
+ * Build the request-body parameters that disable thinking/reasoning for the
+ * given model on the given provider endpoint, or `null` if the model is not a
+ * known reasoning model. The provider (baseURL) is consulted so that models
+ * hosted on different endpoints receive the correct parameter shape.
+ */
+function buildDisableReasoningParams(model, baseURL) {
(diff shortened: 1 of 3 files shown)


app3apps (changes requested):
The DashScope `deepseek-v4-flash` fix is valuable, and the verified path looks sound. I am requesting changes because the expanded table also ships provider/model mappings that are incorrect for configurations this project supports.

1. **MiniMax M2 cannot be disabled by this rule** (`src/llm-client.ts:188-192`). The current MiniMax OpenAI-compatible documentation explicitly says thinking cannot be disabled for M2.x, but this regex matches every `MiniMax-M2*` model and injects `thinking_budget: 0`. That cannot deliver the advertised behavior and may be ignored or rejected. Please remove/narrow the M2 and broad `abab` mapping unless there is a provider-specific documented endpoint and request-level verification: https://platform.minimaxi.com/docs/api-reference/text-openai-api

2. **Most entries are not actually provider-aware** (`src/llm-client.ts:164-207`). Seven rules use a universal `default`, so private provider fields are sent to every OpenAI-compatible endpoint after a model-name match. There is a concrete mismatch with the built-in SiliconFlow path: SiliconFlow documents top-level `enable_thinking: false` for `zai-org/GLM-4.6`, while this table sends `{ thinking: { type: "disabled" } }`. At best reasoning remains enabled; a strict gateway may reject the request before `recoverJsonFromReasoning` can help. Please positively identify provider hosts and return `null` for unknown endpoints, or add an explicit capability override. Add negative unknown-provider coverage and a SiliconFlow case: https://docs.siliconflow.cn/en/api-reference/chat-completions/chat-completions

3. **The Kimi matcher includes its own unsupported exception** (`src/llm-client.ts:179-185`). The comment says `kimi-k2.7-code` cannot disable thinking, but `/kimi.*k[12]|kimi.*thinking/i` matches it and injects `thinking.type=disabled`. Please use a disable-capable allowlist or add an explicit exclusion plus a negative test.

A clean way forward would be to narrow this PR to the live-verified DashScope DeepSeek V4 fix, then add the other families only with provider-specific evidence and tests.

Verification performed: full `npm test` passed on the PR branch; `npm run build` passed with no generated diff; the merge result with current `master` also built and passed the combined LLM API-key client tests.

PR #956, opened 2026-07-18
test: unpin the Ollama abort regression from port 11434 and make it actually exercise the abort

gorkem2020:
Two defects in the Ollama abort regression (Test 8 of the CJK suite), found while running the full chain on a host with a real Ollama install:

1. The mock bound 127.0.0.1:11434 to satisfy the provider detector, which EADDRINUSEs on any machine running a real Ollama and kills the whole npm test chain at position 3. The mock now listens on an ephemeral port and satisfies the detector through the /ollama path alternative of its regex instead.

2. A pre-existing false pass: the mock served /v1/embeddings, but the Ollama-native route strips the /v1 suffix and posts to /api/embeddings, so the request 404ed instantly (about 3ms) and the elapsed assertion, which only had an upper bound, passed without the abort path ever running, on any machine including CI. The mock now serves the real path and the test requires the request to hang until the external AbortSignal fires (verified about 2000ms).

With this change the full test chain runs green end to end on hosts with and without a local Ollama.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -4066,10 +4066,15 @@ const memoryLanceDBProPlugin = {
                 const now = new Date(params.timestampMs ?? Date.now());
                 const dateStr = now.toISOString().split("T")[0];
                 const timeStr = now.toISOString().split("T")[1].split(".")[0];
+                // Session key/id stay out of `text`: it is the FTS index surface, and
+                // the `simple` tokenizer splits a key like
+                // `agent:main:cron:<uuid>:run:<uuid>` on its punctuation — so every session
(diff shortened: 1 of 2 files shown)


app3apps (approved):
Reviewed head `11a1854`. The actual merge result against current `master` changes only `test/cjk-recursion-regression.test.mjs`; the apparent `dist/index.js` drift is an already-landed stale-base patch and does not survive the merge.

The focused regression suite passes. In particular, the Ollama case now reaches `/ollama/api/embeddings` on an ephemeral port and aborts after 2004 ms, satisfying both the lower and upper bounds. That proves the request stayed pending until the external signal instead of passing through an immediate 404 or port collision. I found no blocking issue.

PR #957, opened 2026-07-20
docs: use a whole-store filesystem backup before upgrade

DanielMao1:
The pre-v1.1.0 upgrade steps back up with `memory-pro export --scope global`. But `export` defaults to `--limit 1000` and a single `--scope` (`cli.ts`), so a store with more than 1000 rows — or any memories outside `global` (e.g. `agent:*` scopes) — gets a **silently incomplete backup**: exactly the safety net an in-place `upgrade` is supposed to provide.

This replaces the backup command with a filesystem copy of the LanceDB directory (`cp -r`), which captures the whole store regardless of row count or scope, across all 11 language READMEs. Docs-only; no code change.

Noticed while addressing review feedback on #953, which fixes the same `export`-as-backup pattern in the Jina task-hint migration section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,7 +195,9 @@ Done! Your agent now has long-term memory.
 
 ```bash
 # 1) Backup
-openclaw memory-pro export --scope global --output memories-backup.json
+cp -r <your dbPath> <your dbPath>.bak
+#    whole-store copy; `export` caps at --limit 1000 and one --scope,
+#    so it silently drops rows past 1000 and every non-global scope
 # 2) Dry run
 openclaw memory-pro upgrade --dry-run
 # 3) Run upgrade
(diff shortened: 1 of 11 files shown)


app3apps (changes requested):
Reviewed head `f5fa9f5`. Replacing the limited single-scope export with a whole-store filesystem copy is the right direction, but the sequence still does not produce a reliable backup.

All 11 READMEs tell users to run `cp -r <dbPath> <dbPath>.bak` while OpenClaw may still be writing or compacting the LanceDB directory. A recursive hot copy cannot guarantee that the copied manifests and referenced data files represent one consistent point in time. The instructions also never restart the gateway after the in-place upgrade. This is especially important because backup safety is the entire purpose of the change.

Please make the workflow explicitly quiesce the gateway and any other writers before the copy, copy to a new/non-existing backup destination, run dry-run/upgrade/verification while writes remain stopped, and restart only after verification. Apply the same operational sequence across all localized READMEs. Requesting changes.

PR #958, opened 2026-07-20
refactor(reflection): extract the distiller prompt into system and user parts

gorkem2020:
## What Problem This Solves

The reflection distiller's prompt mixes identity plus static instruction blocks with per-run dynamic content (tool error signals and the transcript fence) in one string. A deployment whose embedded runner honors a real system slot cannot deliver the static block on that slot without forking the prompt builder, and the seam between static and dynamic content is untested, so prompt edits can silently leak dynamic content into the static block or the other way around.

## Why This Change Was Made

`buildReflectionPromptParts` builds the prompt as `{ system, user }`: system carries the identity opener and every static block, user carries only the dynamic content. `buildReflectionPrompt` stays exported as the combined single-slot wrapper (system + blank line + user), byte-identical to the previous prompt, and both runners keep sending that combined form (the embedded raw run does not honor a separate system slot, and the CLI fallback carries a single user message). Zero behavior change; the seam becomes explicit and tested.

## User Impact

None at runtime: prompt text, prompt hash, and delivery are unchanged. Deployments whose runner honors a system slot can deliver the parts split without changing one byte of prompt text.

## Evidence

- New `test/reflection-prompt-split.test.mjs` (7 tests): slot placement for every static sentinel, dynamic-only user content, verbatim combined form, tail clipping, embedded runner contract (combined prompt, no system-slot param, raw-run semantics preserved), and the CLI fallback payload.
- The existing prompt-content suite passes unchanged (grounding discipline tests, 18/18).
- Full test chain green locally; dist rebuilt.


gorkem2020:
Closing as superseded: master already ships the split distiller prompt. buildReflectionPrompt returns { system, user } and the reflection call site consumes the two halves separately, which is the entire content of this refactor in equivalent form.

PR #959, opened 2026-07-20
fix(extraction): keep validation-emptied batches out of the noise bank

gorkem2020:
## What Problem This Solves

Two side effects of the extraction validation layer:

1. `extractAndPersist` trains the embedding noise bank whenever extraction returns ok with zero candidates. A batch can reach zero candidates through validation drops (constructed grounding, fiction register, the batch contradiction demotion) rather than genuine emptiness. In that shape the model DID emit candidates; feeding the conversation to the noise bank as a noise exemplar teaches "content like this is noise" and risks pre-filtering similar real content away from future extractions. Observed live: a mixed-register conversation containing a real durable fact was fed to the noise bank after the contradiction check demoted the whole batch.

2. The batch contradiction check logs its demotions at debug only, so a fully demoted batch is indistinguishable from "model found nothing" at the standard log level. That ambiguity cost a live investigation.

## Why This Change Was Made

- `extractCandidates` now reports `rawCandidateCount` (items the model emitted before validation) on the ok result. Noise learning fires only when that raw count is zero; validation-emptied batches log a standard-level skip line instead.
- When the contradiction check demotes one or more candidates, one standard-level line reports the count and batch register, mirroring the admission rejection lines.

The demotion policy itself is unchanged; this only fixes its side effects.

## User Impact

Deployments stop committing wrong "this is noise" beliefs to the learning store when validation empties a batch, and operators can see demotions without enabling debug logging.

## Evidence

- Flipped the regression test that pinned the old behavior ("learns noise when every parsed candidate is dropped by local validation") and added demotion-emptied and log-line cases; red first, then green (7/7).
- Grounding/register suite unchanged (23/23).
- Full test chain green locally; dist rebuilt.

Note: this branch is cut from the grounding PR's head (#927) because the demotion machinery lives there; it stays a draft until #927 settles and merges.

### DIFF
--- a/dist/src/smart-extractor.js
+++ b/dist/src/smart-extractor.js
@@ -462,12 +462,17 @@ export class SmartExtractor {
                 // No LLM call was made, so the caller's rate limiter must not be charged.
                 stats.skippedNoInput = true;
             }
-            if (extraction.status === "ok" && !extraction.groundingOrPolicyDropped) {
+            if (extraction.status === "ok" && extraction.rawCandidateCount === 0) {
                 // LLM genuinely returned zero candidates → strongest noise signal → feedback to noise bank
                 this.learnAsNoise(conversationText);
             }
             else if (extraction.status === "ok") {
-                this.debugLog("memory-pro: smart-extractor: skipping noise-bank learning (batch emptied by grounding/register/policy drops, not a genuine zero-extraction)");
+                // The model DID emit candidates; validation, grounding, or the batch
+                // contradiction check dropped them all. That is a policy verdict about
+                // those candidates, not evidence the conversation is noise — learning
+                // it as noise would pre-filter similar real content away from future
+                // extractions.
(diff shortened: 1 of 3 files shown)


gorkem2020:
Recomposed against current master (post-[#988](https://github.com/CortexReach/memory-lancedb-pro/pull/988)) as a single commit carrying only this family's delta; full suite green locally. Ready for review.

rwmjhb (approved):
Reviewed head `5fe4075`. The new `rawCandidateCount` gate correctly distinguishes genuinely empty model output from non-empty batches removed by validation, preventing those conversations from training the noise bank. The focused noise-gating suite passes 7/7, the full suite and required GitHub checks are green, `npm run build` succeeds, and the generated tree remains clean.

Approved. Two non-blocking cleanup items: the new demotion log always says “constructed siblings present” even for failure-closed cells with no constructed sibling, and the nearby return-path comment still describes the superseded quality-drop learning behavior.

PR #961, opened 2026-07-20
fix: wrap extraction transcript turns in speaker tags so assistant context cannot be misattributed

gorkem2020:
## Problem

With the "User:"/"Assistant:" line-prefix transcript, only the first line of a message carries a speaker marker. A multi-paragraph assistant reply sheds its marker after the first paragraph, so every later paragraph reads as unowned text sitting between two user turns. In live use the extraction model attributed assistant-authored plans and preferences to the user and stored them as user memories (for example, an assistant's own "what clicks for me now" mental-model summary stored as a user preference), even with the context-only rule present in the prompt.

## Change

- `formatConversationTranscript` now wraps every message wholly in `<user_message>...</user_message>` / `<assistant_message>...</assistant_message>` blocks, so every line has an unambiguous owner regardless of paragraph count.
- `buildExtractionPrompt` teaches the tag format up front (new "Transcript format" section) and keeps a one-line reminder on the conversation header. The assistant-context rule is rewritten in tag vocabulary: context ONLY, usable solely to resolve an ambiguous or obscure user message, never a grounding source, and assistant statements are never attributable to the user. The `captureAssistant: true` eligible variant keeps its existing semantics in the same vocabulary.
- New `trimTranscriptToTagBoundary` snaps the `extractMaxChars` truncation to the next opening tag, so the prompt never leads with a headless half message (the old tail slice could cut mid-message and strand text with no speaker).

## Tests

- New `test/extraction-transcript-speaker-tags.test.mjs` (9 cases): whole-message wrapping, multi-paragraph containment, chronological ordering, tag-boundary truncation with raw-tail fallback, format-teaching placement, context-only pinning, and eligible-variant preservation.
- `test/assistant-context-capture.test.mjs` assertions updated to the tagged format (13/13 green).
- Full suite green.

Stacked on [#939](https://github.com/CortexReach/memory-lancedb-pro/pull/939) (captureAssistant context mode); the diff includes that branch until it merges.


gorkem2020:
Superseded by [#964](https://github.com/CortexReach/memory-lancedb-pro/pull/964), a fresh cut of the speaker-tag transcript that also carries the extraction prompt's system/user split and the tag-spoof neutralizer. The context-tag evolution of the same mechanism lives in [#967](https://github.com/CortexReach/memory-lancedb-pro/pull/967). Closing this draft in favor of the new stack.

PR #962, opened 2026-07-20
fix: retain the rolling pair window across successful extractions

gorkem2020:
## Problem

The extraction transcript's rolling pair window (`autoCaptureRecentPairTurns`, introduced with the pair-shaped context work) was deleted after every successful extraction. In steady state, history-carrying sessions extract every turn, so the window never survived long enough to provide context: each extraction call saw only the current pair, and the multi-pair "context continuity" the window exists for only ever materialized across deferred or valid-empty turns. Observed live: with `extractMinMessages: 3`, a turn following a successful extraction carried a bare 1-pair transcript; a turn following a valid-empty extraction carried 2 pairs (the asymmetry made the window appear and disappear).

## Change

Keep the buffer on success (remove the delete). The window stays bounded and safe:

- `trimTurnsToUserCap` already trims at set time to `max(extractMinMessages, this call's new user turns)` user turns.
- The watermark still slices extraction SOURCES to the new delta, so retained context turns cannot re-become candidates.
- `dedupePairWindow` already guards double-includes.

Also promote the warm-up skip line ("skipped smart extraction... cumulative=N < minMessages=M") from debug to info: it is the only signal that a turn was deferred by the gate, and at debug a deferring session reads as a stall in default logs.

## Tests

New `test/pair-window-retention.test.mjs` (full-plugin harness, mock embedding/LLM servers): a successful extraction must carry the previous pair into the next extraction's transcript, and the window stays trimmed to the cap across repeated successes (oldest pairs drop, long-dropped pairs never resurface). Red before the change (second extraction carried only the current pair), green after. Full suite green.

Stacked on [#939](https://github.com/CortexReach/memory-lancedb-pro/pull/939) (the pair-window home); the diff includes that branch until it merges.


gorkem2020:
Superseded by [#966](https://github.com/CortexReach/memory-lancedb-pro/pull/966), which recuts the pair-window retention together with the autoCaptureContextTurns knob that sizes it (0 preserves stock behavior). Closing this draft in favor of the new stack.

PR #969, opened 2026-07-22
feat: group-chat capture policy with a fail-closed classifier and opt-out knob

gorkem2020:
## Problem

In multi-party rooms the host delivers every non-self participant as user-role content. Extraction has no way to attribute speakers, so peer-agent statements become "User believes X" rows, and with `captureAssistant=true` the misattribution surface widens further. Full group support needs per-sender speaker awareness; until that lands, group sessions need a conservative capture policy.

## Change

Fail-closed group-chat policy, keyed on session-key grammar:

- `isDirectConversationSessionKey` allowlists direct (1:1) conversations from the host's key builder: the dmScope-collapsed main key, dashboard/webchat sessions, and the explicit `:direct:` peer forms. Everything else (channels, groups, threads, topics, and any key shape never seen before) counts as a group chat, fail-closed.
- Group chats force `contextTurns=0` (no context tags, no retention) and `captureAssistant=false` (user-only grounding) regardless of config. Both are threaded per-call into the extractor, so the prompt teaching always matches the transcript actually built for that extraction.
- New `autoCaptureGroupChats` opt-out knob: unset follows `autoCapture`, preserving existing behavior on install with no surprises; explicit `false` skips group-chat capture entirely. The schema deliberately declares no default so unset stays unset.

## Tests

Classifier truth table (direct forms, group forms, unknown shapes failing closed), forced `contextTurns=0` on group keys, forced user-only grounding under configured `captureAssistant=true`, unaffected direct sessions, knob-false skip, and knob-unset passthrough, all in `test/pair-window-retention.test.mjs`. Manifest checks pin the no-default contract.

Stacked on [#968](https://github.com/CortexReach/memory-lancedb-pro/pull/968); the diff includes its predecessors until they merge.


gorkem2020:
Closing this draft. Our deployment no longer runs group chats, so the group-chat capture policy has no consumer on our side and we will not be taking it through review. Happy to reopen if the maintainers want the classifier and the opt-out knob independently.

PR #970, opened 2026-07-22
fix: retry transient embedding failures once in the reflection persistence path

gorkem2020:
## Problem

A session reset ran the reflection distiller and wrote the reflection md, but a single transient embedding failure in the persistence path ("Failed to generate embedding from Jina: Request was aborted.") failed the whole hook. The cycle's mapped rows (invariants/derived) never reached admission or storage, even though the content was already generated and sitting on disk. The generation path retries transient upstream failures once (`runWithReflectionTransientRetryOnce`); the persistence path had no retry at all, and the request-abort class was not in the transient classifier either, so even the generation path would not have retried it.

## Change

The persistence-path embeds now share the generation path's transient-retry policy:

- Mapped-row embeds retry once; a row whose retry is also exhausted is skipped with a warning instead of failing the remaining rows and slices (the loop already skips rows on pre-check failures, so this matches the local error posture).
- Slice-store and session-summary embeds retry once and propagate on exhaustion exactly as before (fail-soft to md-only, unchanged).
- Each embed call carries its own single-retry budget, so one healed abort does not spend the budget of later rows.
- `request was aborted` / `AbortError` are classified transient (they are the timeout/cancellation class); auth-style failures stay non-retryable. This also benefits the generation path, which shares the classifier.

## Tests

New `test/reflection-embed-transient-retry.test.mjs`: abort-class classification (including the incident's exact message), retry-once-and-heal, give-up-after-one-retry, fail-fast on non-transient errors, and per-call budget independence. Registered in the test chain and CI manifest.


rwmjhb (approved):
Approved. The retry is bounded to one additional attempt, uses a fresh budget per embedding call, preserves non-transient fail-fast behavior in the helper, and the targeted/full suites are green. I also ran a clean TypeScript build; the committed `dist/` remains unchanged.

One non-blocking follow-up: `AbortError` and `request was aborted` are classified as transient by message alone. If those errors can also represent intentional cancellation, shutdown, or a caller deadline, consider checking the originating signal/reason before retrying. Caller-level integration coverage for mapped rows, slices, and session summaries would also protect the intended propagation policies.

gorkem2020:
Thanks for the thorough pass. Agreed on the follow-up: classifying `AbortError` / "request was aborted" as transient by message alone can conflate an intentional cancellation or shutdown with a network blip. The right shape is checking the originating signal/reason (retry only when no caller abort was requested) plus caller-level integration coverage for the mapped-row, slice, and session-summary paths. Tracking it as a follow-up change rather than amending this PR, so the approved diff stays as reviewed.

PR #971, opened 2026-07-27
feat(reflection): add memoryReflection.includeGroupChats toggle for group-chat reflection generation

gorkem2020:
## What

New knob `memoryReflection.includeGroupChats` (boolean, default `true`, so existing configs are a no-op). When set to `false`, the command:new / command:reset reflection hook skips reflection GENERATION for group-chat sessions, meaning session keys carrying a `:group:` or `:channel:` segment, and logs an info line naming the skip. Direct/DM/main sessions are unaffected either way.

## Why

The reflection distiller's transcript input renders every non-self participant as a plain user turn, so on multi-agent group channels the distilled lessons and rules misattribute peer agents' messages (and can ingest a parallel session's status lines that ride the shared channel window). Observed live on a multi-agent fleet: a group-channel reset distilled peer-agent coaching and cross-session chatter into the standing rule set.

A structural fix for the distiller input (speaker-tagged transcript, matching the extraction pipeline's format) is the durable direction, but until that lands operators running multi-agent group channels need a supported way to keep reflections scoped to direct sessions. This knob gives them that choice without losing reflections wholesale.

## Implementation

- Config surface: `memoryReflection.includeGroupChats` added to the config type, to BOTH arms of the `parsePluginConfig` memoryReflection rebuild (normalized to boolean, default `true`), and to the manifest schema (nested property plus the config-docs entry).
- `isGroupChatSessionKey(sessionKey)`: single bounded regex (`/:(group|channel):/i`), no unbounded quantifiers.
- Gate sits in the shared command-hook body directly after the existing `excludeAgents` guard, before any session-file recovery work, and returns early with an info log:
  `memory-reflection: command:<action> skipped (group-chat reflection disabled, sessionKey=...)`.

## Tests

Four cells added to `test/command-reflection-guard.test.mjs` (the existing full-register harness):

- skips reflection generation for a `:group:` session when disabled (red-proofed: fails without the gate)
- skips reflection generation for a `:channel:` session when disabled (red-proofed)
- still reflects non-group sessions when the toggle is disabled (pin)
- keeps group-chat reflection enabled by default (pin)

Full suite green, typecheck clean, dist rebuilt and committed, CI test manifest untouched (no new test files).

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -778,6 +778,29 @@ function asNonEmptyString(value) {
 function isInternalReflectionSessionKey(sessionKey) {
     return typeof sessionKey === "string" && sessionKey.trim().startsWith("temp:memory-reflection");
 }
+// Multi-party vs direct peer kinds at the session key's STRUCTURAL kind
+// position, mirroring core's parseSessionDeliveryRoute shape
+// (src/sessions/session-key-utils.ts): agent:<agentId>:<channel>:<peerKind>:
+// <opaque peer id...>, with an optional ":thread:<id>" suffix. The peer id
+// tail is opaque and may itself contain segments like ":channel:" (e.g.
+// "agent:main:discord:direct:user:channel:1" is a DIRECT route), so only the
+// kind position may decide -- never a full-key search. Unrecognized shapes
+// (main sessions, subagents, account-scoped forms, cron) fail toward
+// generating reflections, matching the toggle's enabled default.
+const GROUP_SESSION_PEER_KINDS = new Set(["group", "channel", "room"]);
+function isGroupChatSessionKey(sessionKey) {
+    if (typeof sessionKey !== "string" || sessionKey.length === 0) {
+        return false;
+    }
+    const lower = sessionKey.toLowerCase();
+    const threadAt = lower.lastIndexOf(":thread:");
+    const base = threadAt === -1 ? lower : lower.slice(0, threadAt);
+    const parts = base.split(":");
+    if (parts[0] !== "agent" || parts.length < 5) {
+        return false;
+    }
+    return GROUP_SESSION_PEER_KINDS.has(parts[3]);
+}
 // Any :subagent:/:active-memory: sub-build (delegated subagents in general, not only
 // memory-internal ones) is treated as "its context comes from the parent" across every
 // memory-adjacent hook in this file: auto-recall injection, reflection injection, and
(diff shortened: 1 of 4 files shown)


gorkem2020:
Rebased onto current master (f6e63af) after this week's merges; the toggle, helper, and command-path gate carried over intact (verified site-by-site against the original commit). Full gates green: build, fresh dist, manifest verifier, full suite. Ready for review.

rwmjhb (changes requested):
Thanks for adding a default-preserving opt-out. The config wiring is consistent, but the session-kind classifier is not structurally safe yet:

1. `isGroupChatSessionKey` searches the entire key for `:(group|channel):`. OpenClaw session keys encode the peer kind at a fixed structural position, while the remaining tail is an opaque peer ID: its canonical `parseSessionDeliveryRoute` explicitly reconstructs that ID from all remaining segments. A valid direct route such as `agent:main:discord:direct:user:channel:1` therefore has peer kind `direct` but this helper returns true. In the other direction, OpenClaw's session documentation lists room/channel keys using `...:room:<id>`, which this helper returns false for. With the toggle disabled, direct reflection can be suppressed unexpectedly while supported multi-party rooms continue generating the contaminated reflections this option is meant to stop. Please classify the structural peer-kind component using host session metadata/a canonical parser (with an appropriate fallback), include supported `room` forms, and add both false-positive direct and `room` regressions.

2. The new early return runs before the normal `finally` cleanup of `reflectionErrorStateBySession`. On `command:new`/`command:reset`, a disabled group session can therefore retain a pre-boundary tool error and inject that stale `<error-detected>` reminder into the next prompt before asynchronous `session_end` cleanup wins the race. Please run boundary-state cleanup before this return, or centralize it so all boundary early-return paths are covered, and pin the sequence with a regression.

The two positive-path tests currently prove only that the new skip log is absent. Please assert an actual downstream generation signal/artifact for allowed direct and default-enabled group sessions, and assert no generation/artifact for disabled sessions.

The focused command-reflection suite (10/10), local full suite, and current GitHub CI all pass; these edge cases are not covered by those runs.

gorkem2020:
All three points were correct, fixed at `7a725ed`:

1. **The classifier is structural now.** `isGroupChatSessionKey` mirrors the shape `parseSessionDeliveryRoute` parses (`agent:<agentId>:<channel>:<peerKind>:<opaque peer id...>`, optional `:thread:<id>` suffix stripped first): only the kind position decides, with multi-party kinds `group`/`channel`/`room`. Your example `agent:main:discord:direct:user:channel:1` now classifies as direct (the `:channel:` segment is opaque peer-id tail), `...:room:<id>` forms (the docs' room/channel listing, Matrix rooms included) now honor the toggle, and unrecognized shapes (main sessions, subagents, account-scoped forms) fail toward generating reflections, matching the enabled default.
2. **Boundary state no longer leaks across a skipped boundary.** The gate deletes the session's tool-error state for boundary actions before returning (the same cleanup the generation path's `finally` performs), so a pre-boundary error cannot inject a stale `<error-detected>` reminder into the next session while `session_end` cleanup is still pending. The pre-existing skip paths either have an owning run whose `finally` covers them (re-entrancy, cooldown) or never run reflection for the session at all (excluded/invalid agents, where `session_end` remains the cleaner), so the cleanup is scoped to this gate's return.
3. **The positive-path cells now assert a downstream signal.** Allowed sessions (direct route, default-enabled group, `agent:main:main`) must log the pipeline's `hook start` line; disabled sessions must not, on top of the skip log.

Regressions added: the opaque direct-route false positive, a `room` key, a thread-under-channel key, and an end-to-end boundary-cleanup pin (seed a tool error via `after_tool_call`, skip the boundary on a disabled group session, assert `before_prompt_build` injects no error reminder). Red-proofed: the direct-route, room, and boundary-cleanup cells fail against the previous head `062b85f`; all fourteen pass at `7a725ed`. Full gates green (build, fresh dist, manifest verifier, full suite).

rwmjhb (approved):
Re-reviewed the new head. The prior blockers are addressed: classification now uses the structural peer-kind position instead of searching the opaque peer-id tail, room/thread forms are covered, and skipped boundary commands clear their reflection error state. The focused command-reflection suite passes 14/14, the full suite passed in the orchestrator run, `npm run build` succeeds with a clean generated tree, and current GitHub CI is green.

Non-blocking follow-ups:
- The local classifier is still a narrower approximation of OpenClaw's canonical parser. In the account-scoped direct/DM form, an account ID literally named `group`, `channel`, or `room` can still be mistaken for the peer kind. Prefer canonical host metadata/parser when available, or add a regression for that edge.
- The allowed-route tests currently assert the generic hook-start log, which occurs before session recovery and actual reflection generation. A usable transcript fixture plus a generation-start/artifact assertion would make the positive path stronger.
- Update the config wording to describe structural `group`/`channel`/`room` peer kinds; the current text omits rooms and says any matching segment.

PR #973, opened 2026-07-28
fix(reflection): keep caller-requested aborts out of the transient retry class

gorkem2020:
## Problem

The transient-retry layer added in [#970](https://github.com/CortexReach/memory-lancedb-pro/pull/970) classifies request aborts by error message alone. Two consequences, one promised as a follow-up on that PR's approval and one found by the new coverage:

1. A cancellation the caller itself requested (an aborted signal) matches the transient patterns and earns a pointless retry after the caller has already given up. Intentional cancellation is not a transient failure, whatever the surfaced error text says.
2. The embedding client is constructed with `maxRetries: 0`, and it surfaces a killed connection as "Connection error.", which the transient patterns did not match. Persistence-path retries therefore never fired for the most common transient shape on the api-key path. The new caller-level integration tests exposed this immediately.

## Change

- `src/reflection-retry.ts`: the retry runner and `embedWithReflectionTransientRetry` accept an optional caller signal (an `AbortSignal` works structurally). An aborted signal classifies as the new `caller_aborted` reason and is never retried, and the signal is checked again after the backoff sleep so an abort arriving during the wait also cancels the pending second attempt. Existing call sites are unchanged and keep their behavior; the parameter gives deadline-owning callers a correct path. Also added `connection error` to the transient patterns.
- `test/reflection-embed-transient-retry.test.mjs`: four caller-abort unit cells, plus the caller-level integration coverage promised on [#970](https://github.com/CortexReach/memory-lancedb-pro/pull/970): the mapped-row, slice, and session-summary persistence paths each run against a real embedder client and a local mock server that kills the socket on the first request carrying a poisoned text, asserting the wrapper heals the failure once and the rows still land.

## Validation

- Red before, green after: with the source change reverted, 5 of 13 cells fail (the caller-abort cells and both integration cells); with it in place, 13 of 13 pass.
- `npm run build` clean, `tsc --noEmit` clean, full `npm test` chain green, `core-regression` CI group green, dist rebuilt in the same commit. No registration changes needed (the test file was already in the chain and manifest).
- Deployed to a live gateway: a real `/new` reflection cycle ran generation and every persistence embed through the changed module end to end with zero warnings or errors.

### DIFF
--- a/dist/src/reflection-retry.js
+++ b/dist/src/reflection-retry.js
@@ -8,6 +8,7 @@ const REFLECTION_TRANSIENT_PATTERNS = [
     /socket hang up/i,
     /socket (?:closed|disconnected)/i,
     /connection (?:closed|aborted|dropped)/i,
+    /\bconnection error\b/i,
     /early close/i,
     /stream (?:ended|closed) unexpectedly/i,
     /temporar(?:y|ily).*unavailable/i,
@@ -86,6 +87,9 @@ export function classifyReflectionRetry(input) {
     if (!input.inReflectionScope) {
         return { retryable: false, reason: "not_reflection_scope", normalizedError };
     }
+    if (input.callerAborted) {
+        return { retryable: false, reason: "caller_aborted", normalizedError };
+    }
     if (input.retryCount > 0) {
         return { retryable: false, reason: "retry_already_used", normalizedError };
     }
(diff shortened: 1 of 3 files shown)


rwmjhb (approved):
Reviewed head 76a3febb5494f677750bec46f23b22aa145e4ee2. The connection-error transient classification is active on the production reflection persistence paths, the socket-failure integration coverage and full suite pass, CI is green, and no merge-blocking regression was found. Follow-up: production reflection callers currently do not propagate the optional cancellation signal, so the caller-abort distinction remains unused until that plumbing is added.

PR #974, opened 2026-07-28
fix(repair): confirm rollback absence before reporting unrecovered rows, and move the full-row dump off stderr

gorkem2020:
## Problem

Two follow-ups the reviewer named explicitly on [#923](https://github.com/CortexReach/memory-lancedb-pro/pull/923) round 4 ("the rollback commit-then-reject false `unrecovered` report and full-row stderr output are useful follow-ups, but they are not the blocker"):

1. The repair rollback write shares the replacement add's commit-then-reject hazard, which the replacement path already documents and handles: the write can persist and still surface an error. `repairLegacyScopes` treated any rollback error as data loss and pushed the row into `unrecovered`, falsely reporting a row that was actually restored to the table.
2. The `repair-scopes` CLI dumped every unrecovered row as full JSON (memory text, metadata, embedding vectors) to stderr, flooding terminals and shared logs with row content.

## Change

- `src/store.ts`: the rollback catch now re-reads the row by id and reports `unrecovered` only when the table confirms the row is truly absent. A failed confirmation read stays fail-closed and still reports, so genuine loss is never hidden.
- `cli.ts`: full unrecovered rows are written to a recovery file next to the db path (`<dbPath>-unrecovered-<timestamp>.json`), and stderr carries a bounded id/scope/preview summary plus the file path. When the recovery file cannot be written (or no db path is available), the original full stderr dump remains as the fallback, so the data-preservation guarantee is unchanged.
- `test/scope-owner-leak-hardening.test.mjs`, three new cells in the file that owns this subsystem: a rollback commit-then-reject regression (rollback persists then errors, must not be reported unrecovered, row must remain discoverable as legacy), the recovery-file path (full row with vector preserved in the file, no raw JSON on stderr), and the fallback path (unwritable recovery file keeps the complete stderr dump).

## Validation

- Red before, green after: with the source changes reverted the three new cells fail (49 pass, 3 fail); with them in place, 52 of 52 pass.
- `npm run build` clean, `tsc --noEmit` clean, full `npm test` chain green, `cli-smoke` CI group green (the CLI output change gate), `core-regression` group green, dist rebuilt in the same commit. No registration changes needed (the test file was already registered).
- Deployed to a live gateway: plugin initialized clean, and a real `memory-pro repair-scopes` report-only run executed the changed command against the live store without issues.

### DIFF
--- a/cli.ts
+++ b/cli.ts
@@ -2285,12 +2285,42 @@ export function registerMemoryCLI(program: Command, context: CLIContext): void {
         const { repaired, failed, skipped, unrecovered } = await context.store.repairLegacyScopes(targetScope);
         console.log(`\nRepair complete: ${repaired} reassigned to "${targetScope}", ${skipped} skipped (changed or removed since discovery), ${failed} failed.`);
         if (unrecovered.length > 0) {
-          console.error(
-            `\n${unrecovered.length} row(s) could not be restored after a failed replacement write. ` +
-            "Full row content follows (JSON, one per line) so the data is not lost:",
-          );
-          for (const entry of unrecovered) {
-            console.error(JSON.stringify(entry));
+          // Full rows (text, metadata, vectors) go to a recovery file next to
+          // the db, not to stderr; stderr gets a bounded summary. The full
+          // stderr dump remains only as the fallback when the file cannot be
+          // written, so the data is never lost either way.
+          const dbPathForRecovery = typeof context.store.dbPath === "string" && context.store.dbPath.trim() !== ""
+            ? context.store.dbPath
+            : null;
+          let savedTo: string | null = null;
+          if (dbPathForRecovery) {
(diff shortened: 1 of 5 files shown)


rwmjhb (approved):
Reviewed head `fe3f225`. The targeted tests, full suite, and repository CI pass, and the change improves both ambiguous rollback reporting and the prior unbounded full-row stderr dump. I found no demonstrated HIGH/CRITICAL defect on the production same-handle rollback path. Approving.

Please consider follow-up hardening for the recovery artifact: create it with `0600` and exclusive/collision-resistant semantics, avoid including raw memory text even in the bounded stderr summary, and verify the confirmed row is the restored legacy state rather than merely any row sharing the ID. A two-handle/latest-version regression case would also strengthen the rollback guarantee.

PR #975, opened 2026-07-29
fix(reflection): recognize MiniMax regional endpoint and align provider id

octo-patch:
Reason: MiniMax reflection provider inference ignored the regional api.minimaxi.com endpoint and returned a provider id inconsistent with the explicit model-reference path, making bare-name configuration region-dependent.

## What changed
- `inferProviderFromBaseURL` now recognizes both MiniMax base URLs — `api.minimax.io` (global) and `api.minimaxi.com` (regional) — using the existing `"." + suffix` `endsWith` guard that blocks host spoofing (e.g. `fake-minimax.io`, `fake-minimaxi.com`).
- It now returns the `minimax` provider id, matching the id parsed from an explicit `minimax/<model>` reference in `splitProviderModel`, so a bare model name such as `MiniMax-M3` or `MiniMax-M2.7` resolves to the same provider regardless of the configured region.

## Why
The global host previously mapped to `minimax-portal` while the explicit model-reference path used `minimax`, and the regional host was not matched at all. Bare-name reflection configuration was therefore region-dependent and inconsistent between the two resolution branches (`split.provider ?? inferProviderFromBaseURL(...)`).

## Tests
- Extended `test/infer-provider-from-baseurl.test.mjs` with the updated provider id, both regional hosts (`/v1` and `/anthropic` base URLs), and `fake-minimaxi.com` spoofing protection.
- `node --test test/infer-provider-from-baseurl.test.mjs` — 13/13 pass.
- `npx tsc -p tsconfig.json --noEmit` — passes.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1178,14 +1178,17 @@ function splitProviderModel(modelRef: string): { provider?: string; model?: stri
 
 /**
  * When modelRef is a bare name (no / prefix), infer provider from baseURL.
- * Use "." + suffix to prevent fake-minimax.io subdomain spoofing.
+ * Use "." + suffix to prevent host spoofing (e.g. fake-minimax.io). Both
+ * MiniMax regional endpoints (minimax.io / minimaxi.com) map to the same
+ * "minimax" provider id used by the explicit modelRef path (e.g.
+ * "minimax/MiniMax-M3"), so a bare model name resolves consistently.
  */
 export function inferProviderFromBaseURL(baseURL: string | undefined): string | undefined {
   if (!baseURL) return undefined;
   try {
     const url = new URL(baseURL);
     const hostname = url.hostname.toLowerCase();
-    if (hostname.endsWith(".minimax.io")) return "minimax-portal";
+    if (hostname.endsWith(".minimax.io") || hostname.endsWith(".minimaxi.com")) return "minimax";
     if (hostname.endsWith(".openai.com")) return "openai";
     if (hostname.endsWith(".anthropic.com")) return "anthropic";
     return undefined;
--- a/test/infer-provider-from-baseurl.test.mjs
+++ b/test/infer-provider-from-baseurl.test.mjs
@@ -19,9 +19,24 @@ const { inferProviderFromBaseURL } = indexModule;
 describe("inferProviderFromBaseURL - PR #713 regression", () => {
 
   describe("URL hostname inference", () => {
-    it("baseURL with minimax.io returns minimax-portal", () => {
+    it("baseURL with minimax.io returns minimax", () => {
       const result = inferProviderFromBaseURL("https://api.minimax.io/v1");
-      assert.strictEqual(result, "minimax-portal");
+      assert.strictEqual(result, "minimax");
+    });
+
+    it("baseURL with minimax.io anthropic endpoint returns minimax", () => {
+      const result = inferProviderFromBaseURL("https://api.minimax.io/anthropic");
+      assert.strictEqual(result, "minimax");
+    });
+
+    it("baseURL with minimaxi.com (regional endpoint) returns minimax", () => {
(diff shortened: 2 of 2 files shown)


rwmjhb (changes requested):
Reviewed head `43a9711`. The source-level targeted tests and full suite pass, but the published runtime does not contain this fix.

`index.ts` now recognizes `.minimaxi.com` and returns `minimax`, while the tracked `dist/index.js` at this same head still recognizes only `.minimax.io` and returns `minimax-portal`. `package.json.main` points installed plugins at `dist/index.js`, and this PR changes only `index.ts` plus the source-oriented test. As submitted, repository tests pass but installed/published users continue executing the old behavior.

Please run the project build, commit the resulting `dist/index.js`, and run `npm run verify-package-runtime` (or the equivalent package dry-run) so the generated delivery artifact is verified. A focused packaged-runtime assertion for both MiniMax hostnames would prevent this source/dist split from recurring.

Non-blocking compatibility follow-up: the existing `.minimax.io` return value changes from `minimax-portal` to `minimax`; an end-to-end provider consumer test would make that intentional contract change safer than helper-only expectations.

octo-patch:
Thanks for the review. I rebuilt and committed the distribution bundle, then verified the package runtime against fresh build output. The 13 focused tests and package-runtime verification passed.

rwmjhb (commented):
Thanks for rebuilding the distribution bundle. On the current head, the source and published `dist/index.js` both contain the MiniMax regional-host/provider fix, and the focused plus full local suites pass, so the previous source/dist blocker is addressed.

GitHub now reports this PR as `CONFLICTING` / `DIRTY`, with no current checks available. Please rebase onto the latest `master` (or merge it into the branch), resolve the conflicts, rebuild and commit a clean `dist`, and rerun the package-runtime verification. I will complete the provider compatibility review against the rebased head; the current branch cannot be approved or merged as-is.

octo-patch:
Thanks for the review. The PR branch already includes the latest master commit (5035cda), so no local merge conflicts were present. I rebuilt the distribution and ran npm run verify-package-runtime plus the 13 focused provider-inference tests; all passed. I pushed a branch refresh commit for GitHub to recalculate merge state.

rwmjhb:
Thanks for refreshing the branch and reporting the local verification. GitHub still reports the current head `a0a62d4` as `CONFLICTING / DIRTY` against the latest `master`, so the PR remains unmergeable and I cannot complete the final provider-compatibility review on this head.

Please rebase onto the current `master` (or merge it into the branch), resolve the conflicts, rebuild the tracked `dist`, and rerun `npm run verify-package-runtime` plus the focused provider-inference tests. Once GitHub reports a clean merge state, I will re-review the rebased diff.

octo-patch:
Thanks for the review. I merged the latest master into the PR branch, resolved the index.ts conflict, and rebuilt the tracked dist. I ran `node --test test/infer-provider-from-baseurl.test.mjs` (13 passing), `npm run build`, and `node scripts/verify-package-runtime.mjs`; all passed.

rwmjhb (approved):
Re-reviewed head `4114229`. The branch is now mergeable, the conflict resolution preserves the MiniMax regional-host change, and source plus tracked `dist` are synchronized.

Independent verification on this exact head passed:

- `npm run build` (with a clean worktree afterwards)
- `node --test test/infer-provider-from-baseurl.test.mjs` (13/13)
- `node scripts/verify-package-runtime.mjs`
- the orchestrator targeted and full suites

Returning `minimax` for both `.minimax.io` and `.minimaxi.com` now matches the explicit `minimax/<model>` provider path while retaining the hostname-spoofing guard. Approving this head.

Non-blocking compatibility note: this deliberately changes the previous `.minimax.io` inference result from `minimax-portal` to `minimax`; any host catalog that only understands the legacy literal would need an alias, although the existing explicit model-reference path already uses `minimax`.

Issue #981, opened 2026-08-03
Canonical corpus re-embeds and re-upserts every chunk on every sync (no change detection) — ~1GB store growth per restart


benjaml4:
**Plugin version:** 1.1.0-beta.11 (master)
**OpenClaw:** 2026.7.2-beta.6
**Corpus size:** 796 documents / 2505 chunks / 7253 total rows

## Symptom

The LanceDB store grew by roughly **1 GB per gateway restart** — from 76 MB to 9.2 GB across a handful of restarts — for a corpus whose actual vectors are only ~23 MB (2505 × 768-dim float32).

## Root cause

`runSync()` in `src/corpus-indexer.js` has **no change detection**. It unconditionally re-embeds and re-upserts every chunk on every sync:

```js
for (const chunk of chunks) {
    try {
        const vector = await this.params.embedder.embedPassage(chunk.text);
        await this.params.store.upsert(toMemoryEntry(chunk, vector));
        stats.indexed++;
    }
    ...
}
```

`stats.skipped` only ever counts *errors* — there is no "unchanged" path. Since `lastSyncAt` is instance state and resets with each process, every restart performs a full re-embed + re-upsert of the entire corpus, and each upsert pass creates a new LanceDB version set.

Two knock-on effects:

- **Unbounded storage growth**, because…
- **`store.runStorageMaintenance()` cannot reclaim it.** It clamps retention with `clampInt(retentionDays, 1, 3650)`, so `cleanupOlderThan` is always at least 24h in the past and same-day versions are never pruned. `storageMaintenance.autoCleanup` is also disabled by default, so out of the box nothing prunes at all.
- Wasted embedding cost — 2505 embedding calls per restart that produce identical vectors.

## The data to fix it is already written but never read

`toMemoryEntry()` already stores per-chunk and per-document hashes:

```js
corpus_content_sha256: sha256(chunk.text),
corpus_document_sha256: sha256(doc.content),
corpus_mtime_ms: doc.mtimeMs,
```

and `store.listCorpusEntryRefs()` already bulk-returns `{id, scope, metadata}` for all corpus rows. Comparing the stored `corpus_content_sha256` against `sha256(chunk.text)` before embedding is sufficient. `buildCorpusId()` is deterministic, so ids line up.

[...]

PR #982, opened 2026-08-03
fix: activate the canonical corpus indexer (agents.entries + startup sync)

### LINKED ISSUE #980: canonicalCorpus never indexes: sync() is never invoked, and workspace resolution reads agents.list instead of agents.entries
**Plugin version:** 1.1.0-beta.11 (master)
**OpenClaw:** 2026.7.2-beta.6
**Config:** `canonicalCorpus.enabled: true`, `syncOnSearch: true`

With `canonicalCorpus.enabled: true` the in-place index silently stayed empty forever. Two independent causes.

## 1. `canonicalCorpusIndexer.sync()` is never called

`dist/index.js` constructs the indexer but never calls `.sync()`. The only call sites are in `src/openclaw-memory-capability.js`:

- the capability's `search()` (gated on `canonicalCorpus.syncOnSearch`)
- the capability's `sync()`

Both are only reachable when the **host** drives the memory capability via `getActiveMemorySearchManager`. But the plugin also registers its own `memory_search` / `memory_get` tools, so agent searches resolve to those tools and never touch the capability. Net result: the canonical corpus never indexes, with no error and no log.

Reproduce: enable `canonicalCorpus`, own the memory slot, restart, run a `memory_search` from an agent — the store gains no `corpus:` rows.

Suggested fix: kick a fire-and-forget `sync()` when the plugin service starts, and/or drive it on a timer from `syncIntervalMs`.

## 2. `resolveCanonicalCorpusWorkspaces` reads `agents.list`, but OpenClaw uses `agents.entries`

`src/corpus-indexer.js`:

```js
const list = Array.isArray(agents?.list) ? agents.list : [];
for (const entry of list) { ... add(entry.workspace ?? ..., entry.id); }
```

OpenClaw stores agents as an **object keyed by agent id** under `agents.entries`, not an array under `agents.list`. The loop therefore matches nothing and the function falls through to its single-workspace fallback:

```js
byWorkspace.set(join(homeDir, ".openclaw", "workspace"), new Set(["main"]));
```

So only the default workspace is ever indexed. In my case that silently missed 5 of 6 agent workspaces — and because the fallback "works", there is no symptom suggesting anything is wrong.

Supporting both shapes fixes it. After patching, `resolveCanonicalCorpusWorkspaces` returns all 6 workspaces and a sync reports:

```
indexed 2505/2505 canonical corpus chunk(s), 0 stale (startup) — from 796 document(s)
```

Happy to send a PR for either or both if useful.

benjaml4:
Fixes #980.

Two independent faults left the canonical corpus permanently empty when `canonicalCorpus.enabled` was set — details and reproduction in the issue.

1. **`canonicalCorpusIndexer.sync()` was never invoked.** Its only call sites are inside the memory capability, which agent searches bypass because the plugin registers its own `memory_search`/`memory_get` tools. This adds a fire-and-forget sync at service start, mirroring the existing `runStartupChecks` pattern so gateway startup is never blocked. `sync()` already checks `enabled`, rate-limits via `syncIntervalMs`, and de-duplicates concurrent runs.
2. **`resolveCanonicalCorpusWorkspaces` read `agents.list`**, but OpenClaw stores agents keyed by id under `agents.entries` — the loop matched nothing and fell back to the single default workspace. Both shapes are now supported; `agents.list` handling is untouched.

Verified on a 6-agent install (OpenClaw 2026.7.2-beta.6): workspaces resolved went 1 → 6, and the startup sync indexed **2505/2505 chunks from 796 documents**.

Note: the startup log line intentionally omits the `unchanged` counter so this branch stands alone; #981's PR adds that counter. `tsc --noEmit` clean.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -6128,6 +6128,28 @@ const memoryLanceDBProPlugin = {
         // Fire-and-forget: allow gateway to start serving immediately.
         setTimeout(() => void runStartupChecks(), 0);
 
+        // Canonical corpus: nothing else invokes canonicalCorpusIndexer.sync().
+        // Its only call sites are inside the memory capability (search()/sync()),
+        // which the host reaches via getActiveMemorySearchManager — but agent
+        // searches resolve to this plugin's own memory_search/memory_get tools
+        // and bypass the capability, so with canonicalCorpus.enabled the index
+        // silently stayed empty. Kick one sync at service start; sync() checks
+        // canonicalCorpus.enabled itself, rate-limits via syncIntervalMs, and
+        // de-duplicates concurrent runs, so this is safe and cheap when current.
+        setTimeout(() => void (async () => {
+          try {
+            const corpusStats = await canonicalCorpusIndexer.sync({ reason: "startup" });
+            api.logger.info(
+              `memory-lancedb-pro: canonical corpus startup sync — indexed ${corpusStats.indexed}/${corpusStats.chunks} chunk(s) ` +
+              `from ${corpusStats.documents} document(s); ${corpusStats.staleDeleted} stale removed`,
+            );
+          } catch (err) {
+            api.logger.warn(
+              `memory-lancedb-pro: canonical corpus startup sync failed: ${err instanceof Error ? err.message : String(err)}`,
+            );
+          }
+        })(), 0);
+
         // Check for legacy memories that could be upgraded
         setTimeout(async () => {
           try {
--- a/package-lock.json
+++ b/package-lock.json
@@ -86,9 +86,6 @@
         "node": ">= 18"
       }
     },
-    "node_modules/@lancedb/lancedb-darwin-x64": {
-      "optional": true
-    },
     "node_modules/@lancedb/lancedb-linux-arm64-gnu": {
       "version": "0.26.2",
       "resolved": "https://registry.npmjs.org/@lancedb/lancedb-linux-arm64-gnu/-/lancedb-linux-arm64-gnu-0.26.2.tgz",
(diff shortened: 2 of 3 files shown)


rwmjhb (changes requested):
Reviewed head `00550dc`. Activating canonical corpus sync and supporting `agents.entries` addresses issue #980, but startup activation currently introduces a cost/storage blocker.

Every process start calls `sync()`. `runSync()` then unconditionally calls `embedPassage()` and `store.upsert()` for every chunk; `lastSyncAt` is process-local and resets on restart, while `upsert()` deletes and re-adds each row. An unchanged corpus is therefore fully re-embedded and rewritten on every restart. Issue #981 reports roughly 1 GB of LanceDB growth per restart for 2,505 chunks, in addition to repeated embedding cost.

Please land content-hash change detection with startup activation: bulk-load existing corpus refs, compare deterministic IDs and `corpus_content_sha256`, preserve unchanged IDs in `expectedIds`, and embed/upsert only changed chunks. Add regressions for unchanged corpora across restarts, `agents.entries` workspace discovery, enabled/disabled startup sync, and startup failure handling. The one-time startup call also leaves normal `memory_search` stale after later file changes because that path still bypasses `syncOnSearch`; please either wire the rate-limited sync into the primary search path or track that explicitly as a follow-up. Revert or explain the unrelated LanceDB platform lockfile removal.

Requesting changes.

rwmjhb (commented):
Fresh recheck of `00550dc`: there have been no author commits or replies since the August 4 changes-requested review, and the original blocker remains. Enabling startup sync still re-embeds and delete/re-add upserts every unchanged corpus chunk on each process restart because there is no persisted content-hash/ID change detection; this is the repeated cost and LanceDB-growth path described in the prior review.

The current head also cannot be installed from a clean checkout: `npm ci` fails with `EUSAGE` because `package-lock.json` is missing `@lancedb/lancedb-darwin-x64` and is out of sync with `package.json`. Please restore/regenerate the lockfile rather than removing that platform entry.

Please push the content-hash incremental sync, startup/agents.entries/failure regressions, and a clean lockfile before requesting re-review. The existing CHANGES_REQUESTED decision still applies.

Issue #978, opened 2026-08-01
fuseResults: BM25-only candidates bypass weighting and clamp to 1.0, so keyword matches always outrank semantic ones


islandpreneur007:
### Summary

In `hybridRetrieval`, `fuseResults()` scores candidates differently depending on whether they had a
vector hit, and the BM25-only branch bypasses the configured weighting entirely:

```js
const fusedScore = vectorResult
  ? clamp01(Math.max(weightedFusion, bm25Score >= 0.75 ? bm25Score * 0.92 : 0), 0.1)
  : clamp01(bm25Result.score, 0.1);   // <-- raw BM25 score, no weighting
```

BM25 scores are unbounded (Lucene-style), while cosine similarity is bounded to roughly 0–1.
`clamp01` therefore maps essentially **every** BM25-only candidate to **1.0**, while a
semantically-correct candidate is limited to `vectorScore * vectorWeight` — at most `vectorWeight`.

With the default `vectorWeight: 0.7`, a keyword-only match scores a perfect 1.0 and a strong semantic
match caps at 0.7. **Keyword-only candidates systematically outrank semantic ones**, regardless of how
`vectorWeight` / `bm25Weight` are configured, because the weighting is not applied on that branch.

### Measured on a real store (4,316 memories, `jina-embeddings` 1024-dim)

Query: *"what has Mike decided about failed message sends, and why?"*

| side | top score |
|---|---|
| BM25 | **25.86** |
| cosine similarity | **0.4279** |

After fusion: BM25-only candidate → `clamp01(25.86)` = **1.0**; best semantic candidate →
`0.4279 * 0.7` = **0.2995**.

### Observed effect

The correct record ranks **#2 of 4,316** by raw cosine similarity and is returned at **#2** by
`store.vectorSearch` — but never appears in hybrid results. Instead, all ten returned results were
long auto-generated "Reflection" documents, most containing only empty section headers
(`(none captured)`). Those 32 records are **1%** of the store yet occupied **100%** of the top
results for many different queries, because length gives them broad BM25 term coverage while their
semantic similarity is mediocre.

Downstream consequences we hit:
- A cross-encoder rerank cannot help — it only reorders what fusion passes it.
- The tier-1 suppression governance fires constantly (`auto-recall skipped after governance filters`,
  ~822 events), because the candidates genuinely are poor.

### Workaround

Setting `retrieval.mode: "vector"` bypasses fusion and restores correct ordering — verified: the same
[...]

PR #984, opened 2026-08-03
fix: apply fusion weighting uniformly and recalibrate the exact-keyword floor

benjaml4:
Addresses the confirmed remainder of #978 (post-correction).

As the reporter's follow-up established, BM25 scores **are** sigmoid-normalised before fusion — but the structural asymmetry stands: the BM25-only branch used the normalised score directly, bypassing `bm25Weight`, so keyword-only candidates (sigmoid floor 0.5) outranked semantic candidates capped at `vectorWeight` no matter how the weights were configured.

Two changes:

1. **Uniform scoring** — every candidate gets `clamp01(max(v·vectorWeight + b·bm25Weight, keywordFloor), 0.1)`.
2. **Floor recalibration** — the exact-keyword floor's 0.75 threshold predates the sigmoid: `1/(1+e^(-raw/5))` reaches 0.75 at raw ≈ 5.5, which almost every FTS hit clears, so the floor (b·0.92 ≥ 0.69) quietly re-created the bypass on both branches. 0.95 corresponds to raw ≈ 14.7 — genuinely exceptional matches (identifiers, exact strings), preserving the floor's documented purpose of keeping those above `minScore`.

Observed effect on a 7,253-row store: queries with no relevant corpus content used to return unrelated keyword-only hits at floored confidence 0.85; they now score ~b·bm25Weight (≈0.28, below default `minScore`) and are suppressed, while on-topic queries are unaffected (verified before/after on the same store).

If you'd prefer the floor threshold configurable rather than a constant, happy to add a `retrieval.keywordFloorThreshold` option instead. An alternative worth considering longer-term is rank-based fusion (RRF) — which the comment above this code already names — as it sidesteps score-scale calibration entirely. `tsc --noEmit` clean.

### DIFF
--- a/src/retriever.ts
+++ b/src/retriever.ts
@@ -1347,20 +1347,25 @@ export class MemoryRetriever {
       // BM25 hit acts as a bonus (keyword match confirms relevance).
       const vectorScore = vectorResult ? vectorResult.score : 0;
       const bm25Score = bm25Result ? bm25Result.score : 0;
-      // Weighted fusion: vectorWeight/bm25Weight directly control score blending.
-      // BM25 high-score floor (>= 0.75) preserves exact keyword matches
-      // (e.g. API keys, ticket numbers) that may have low vector similarity.
+      // Weighted fusion: vectorWeight/bm25Weight directly control score blending,
+      // applied identically to every candidate. Previously the BM25-only branch
+      // used the normalised BM25 score directly, bypassing bm25Weight — so a
+      // keyword-only candidate kept its full sigmoid-normalised score (floor 0.5
+      // by construction) while a semantic-only candidate was capped at
+      // vectorWeight, and keyword noise systematically outranked strong semantic
+      // matches regardless of configured weights (#978).
+      //
+      // The exact-keyword floor preserves identifier-like matches (API keys,
+      // ticket numbers) above minScore even with weak vector similarity. Its
+      // threshold is calibrated to the sigmoid normalisation in
+      // store.bm25Search (1/(1+e^(-raw/5))): the old 0.75 threshold maps to a
+      // raw BM25 of ~5.5, which almost every FTS hit clears — making the floor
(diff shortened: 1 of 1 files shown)


rwmjhb (changes requested):
Reviewed head `2043eb9`. Applying configured fusion weights uniformly is the right direction, but the new `0.95` keyword-floor policy introduces two retrieval blockers.

1. `test/retriever-rerank-regression.mjs` fails because a strong lexical/semantic candidate (vector `0.5006586`, BM25 `0.78`) now fuses to about `0.5845`, falls below `minScore=0.6`, and is removed before reranking.
2. BM25 graceful degradation is broken. With default `bm25Weight=0.3`, every BM25-only score below `0.95` is weighted below `0.285` and cannot pass the default `minScore=0.3`; the existing graceful-degradation test reproduces this with a `0.9` BM25 hit when vector search fails.

Please preserve configured weighting when both backends participate while renormalizing or using the surviving backend's score when one backend fails, and replace the universal floor with a calibrated/configurable exact-match signal that keeps existing strong lexical matches. Add boundary tests around the floor and keep both existing regressions green.

Requesting changes.

rwmjhb (commented):
Fresh recheck of `2043eb9`: there have been no author commits or replies since the August 4 changes-requested review, and the retrieval regressions remain reproducible.

- `node test/retriever-rerank-regression.mjs` fails at the strong lexical-hit assertion: the expected result count is 1, but the new fusion/floor policy returns 0 before reranking.
- `node --test test/retriever-graceful-degradation.test.mjs` fails 2 of 8 tests, including the BM25-only fallback when vector search fails; the expected result count is 1, but weighting a surviving BM25 score by the default 0.3 removes it.

Please renormalize to the surviving backend when one search backend fails, and replace the universal 0.95 keyword floor with a calibrated/configurable exact-match rule that preserves existing strong lexical hits. Add boundary regressions before requesting re-review. The existing CHANGES_REQUESTED decision still applies.

PR #986, opened 2026-08-03
fix: exempt canonical corpus chunks from length normalization and decay

### LINKED ISSUE #985: Canonical corpus chunks are crushed by length normalization + decay — top-match content returns 0 results
**Plugin version:** 1.1.0-beta.11 (master) + the fixes from #982/#983 applied locally
**OpenClaw:** 2026.7.2-beta.6

## Symptom

With the canonical corpus indexed and healthy (2,505 chunks, correct rows verified present), `memory_search` and `memory_recall` returned **0 results** for a query whose target chunk is the store's **#1 match on both retrieval sides** — FTS raw 27.5 (top hit) and vector rank 6 within the agent's scope.

## Root cause

Post-fusion lifecycle scoring is calibrated for conversation memories and structurally crushes canonical corpus chunks:

1. **`applyLengthNormalization`** anchors at ~500 chars. Corpus chunks are line-span document chunks — ~3,900 chars here — giving `1/(1 + 0.5·log2(7.8)) ≈ 0.40`. Length is a property of the chunker (already bounded by the indexer), not of entry quality.
2. **`applyDecayBoost`/`applyTimeDecay`** score by entry timestamp, and corpus rows carry the **source file's mtime** (`toMemoryEntry: timestamp: doc.mtimeMs`). A six-week-old reference document lands near `boostMin`.

Compound: fused **0.917** (exact-keyword floor) × ~0.40 (length) × ~0.4 (decay) ≈ **0.15** → below `hardMinScore`/`minScore` → dropped. Every sufficiently old or long corpus chunk is unreachable regardless of how well it matches.

This interacts with #981's dedup direction: while `memory-pro import-markdown` copies exist (small, freshly-timestamped), they mask the problem by serving recall; a store relying on the canonical corpus alone loses recall almost entirely.

## Fix (PR attached)

Exempt corpus entries from both stages — corpus ids are always `corpus:`-prefixed (`buildCorpusId`), so the guard is a cheap string check. Verified before/after on a live store: the 0-result query returns its target correctly with the exemption, and conversation-memory scoring is untouched.

An alternative, if you'd rather keep some ageing on corpus content, would be a separate (much gentler) decay profile for `openclaw_corpus` entries — happy to rework the PR that way.

benjaml4:
Fixes #985.

Corpus chunks are line-span document chunks whose length is set by the indexer, and whose timestamps are source-file mtimes — so conversation-memory length normalisation and decay compound to ~0.15 on a chunk that fused at 0.917, dropping the store's #1 match below `minScore`. Details and measurements in the issue.

The guard is `id.startsWith("corpus:")` (always true for `buildCorpusId` output), applied in `applyLengthNormalization`, `applyDecayBoost`, and `applyTimeDecay`. Conversation-memory scoring is untouched. `tsc --noEmit` clean; verified before/after on a live 2,505-chunk store.

Complements #982/#983 — with all three, a canonical-corpus-only store (no import-markdown copies) has working recall.

### DIFF
--- a/src/retriever.ts
+++ b/src/retriever.ts
@@ -1634,7 +1634,11 @@ export class MemoryRetriever {
 
     const reranked = results.map((result, index) => ({
       ...result,
-      score: clamp01(scored[index].score, result.score * 0.3),
+      // Corpus rows carry the source file's mtime as their timestamp;
+      // reference material must not decay like conversation memory.
+      score: (result.entry.id ?? "").startsWith("corpus:")
+        ? result.score
+        : clamp01(scored[index].score, result.score * 0.3),
     }));
 
     return reranked.sort((a, b) => b.score - a.score);
@@ -1654,6 +1658,11 @@ export class MemoryRetriever {
     if (!anchor || anchor <= 0) return results;
 
     const normalized = results.map((r) => {
+      // Canonical corpus chunks are line-span document chunks: their length is
+      // a property of the chunker, not of entry quality. Normalising them by
+      // length double-penalises reference material (chunk size is already
+      // bounded by the indexer).
+      if ((r.entry.id ?? "").startsWith("corpus:")) return r;
       const charLen = r.entry.text.length;
       const ratio = charLen / anchor;
       // No penalty for entries at or below anchor length.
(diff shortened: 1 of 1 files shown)


rwmjhb (approved):
Reviewed head `8ce9371`. Exempting canonical corpus chunks from conversation-oriented length normalization and decay directly addresses the reported retrieval failure. The targeted checks and full local suite pass.

Please follow up with direct retrieval regressions covering all three exemptions and a shared corpus-entry predicate; these are non-blocking for this focused change. GitHub currently reports no check runs for this head.

Approved.

PR #990, opened 2026-08-19
docs: fix broken star history chart across readmes

FaintFlower:
The star history chart is currently broken and renders no image, because the service it points to is limited by GitHub stargazer API restrictions. This fixes it by routing the chart to a working destination that requires no API token, applied consistently to the main readme and every localized version.


rwmjhb (approved):
Reviewed head 79866e1. The change is consistently applied across all 11 README variants, no old api.star-history.com chart references remain, and both light/dark SVG endpoints return HTTP 200 with image/svg+xml while the linked chart page returns HTTP 200. This is a focused documentation-only fix with no code-path risk.

PR #994, opened 2026-09-06
fix: resolve the OpenClaw state dir consistently (md-mirror workspace map, CLI and workspace defaults)

gorkem2020:
## Problem

The plugin assumed a single OpenClaw instance living at `~/.openclaw`:

- `resolveAgentWorkspaceMap` (md-mirror) only read the legacy `agents.list` array. OpenClaw 2026.9 writes the keyed `agents.entries` form, so the mirror logged `mdMirror: no agent workspaces found` and wrote every agent's mirror into the default directory.
- When the runtime config carried no agents, the fallback read `~/.openclaw/openclaw.json`. On a host running a second instance (`OPENCLAW_STATE_DIR` pointing elsewhere) that is another instance's config; the observed effect was the second instance's mirror landing inside the first instance's main-agent workspace.
- The default db, workspace and md-mirror directories, the CLI `--openclaw-home` default and OAuth file path, and the workspace capability default were all rooted at `homedir()/.openclaw` regardless of `OPENCLAW_STATE_DIR`.

## Change

- `src/openclaw-paths.ts`: one resolver. State dir = `OPENCLAW_STATE_DIR`, then `OPENCLAW_HOME`, then `~/.openclaw`; config path = `OPENCLAW_CONFIG_PATH`, then `<state dir>/openclaw.json`. Blank or whitespace-only values count as unset.
- The md-mirror workspace map reads both `agents.entries` (keyed) and `agents.list` (array), from the runtime config first and from the resolved config file second.
- Plugin defaults, the CLI and the workspace capability default all go through the resolver. Single-instance installs resolve to exactly the same paths as before, so there is no behaviour change for the default layout.

## Tests

- `test/openclaw-paths.test.mjs`: precedence and blank-variable handling.
- `test/mdmirror-workspace-map-entries.test.mjs`: entries form, list form, `OPENCLAW_CONFIG_PATH` fallback, state-dir fallback, state-dir-rooted and home-rooted defaults.
- Both are registered in the `npm test` chain and in `scripts/ci-test-manifest.mjs` (core-regression). Full suite, core-regression, cli-smoke, packaging-and-workflow and `verify-package-runtime` are green locally; dist is rebuilt and committed.

## Verification on a live install

Running as a hotfix on a seven-agent OpenClaw 2026.9.2 install: the startup log went from `mdMirror: no agent workspaces found` to `mdMirror: resolved 7 agent workspace(s)`, and mirror files land in each agent's own workspace.


gorkem2020:
Ready for review. Rebased onto the current master ([93899f88](https://github.com/CortexReach/memory-lancedb-pro/commit/93899f88), post [#946](https://github.com/CortexReach/memory-lancedb-pro/pull/946)); only the two test-registration files conflicted (ordered union), the source is unchanged from the draft.

Fleet evidence: the mirror half (per-agent workspace map from `agents.entries`) has run on our gateway since 2026-09-06 and the shared state-dir resolver since the 2026-09-12 cut; every agent's markdown mirror lands in its own workspace and the CLI resolves the same paths as the runtime. `npm run build` and the full `npm test` chain pass on the rebased head.

PR #1000, opened 2026-09-16
fix(memory-runtime): classify workspace memory paths so the host keeps MEMORY.md and USER.md in the bootstrap prompt

gorkem2020:
## Problem

On OpenClaw 2026.9.x the host asks the memory-slot owner to classify the provenance of workspace memory files before it injects them automatically (`MemoryPluginRuntime.classifyWorkspaceMemoryPaths`, see `src/plugins/registry-contribution-types.ts` and `src/agents/bootstrap-files.ts` in openclaw). memory-lancedb-pro registers a runtime without that method, so `classifyActiveMemoryWorkspacePaths` reports `unsupported`, the host logs

```
excluding automatic memory context: selected memory runtime does not support provenance classification (sessionKey=agent:<id>:main)
```

on every session start, and treats the bootstrap candidates (`MEMORY.md` and `USER.md`) as ineligible. The visible effect is the log line; the real effect is that `MEMORY.md` and `USER.md` silently drop out of every agent's bootstrap prompt while this plugin holds the memory slot.

## Fix

The runtime object built by `createOpenClawMemoryCapability` now implements `classifyWorkspaceMemoryPaths`. The classification mirrors the host's own default provider (memory-core `workspace-path-classifier` and `memory-path-provenance`) without its private provenance sidecar, which is not reachable through the plugin SDK:

- `MEMORY.md`, `memory.md`, `USER.md` at the workspace root: `agent`
- `memory/**/*.md` (the plugin's own journal and reflection outputs live there): `agent`
- `DREAMS.md`, `memory/dreaming/**`, `memory/.dreams/**`: `system`
- anything else, and any path that resolves outside the workspace: `untrusted`

`agent` is eligible for automatic injection in the host (`isMemoryOriginEligibleForAutomaticInjection`), so `MEMORY.md` and `USER.md` are injected again and the log line stops. Paths are returned in the requested order, verbatim.

## Tests

`test/memory-runtime-provenance-classification.test.mjs` (registered in the `test` chain and the CI manifest): the runtime exposes the method and keeps both bootstrap candidates eligible; journals and reflections are `agent`; dream artifacts are `system`; other files, non-markdown files under `memory/`, and paths escaping the workspace are `untrusted`; order and paths are preserved.

`npm run build` and the full `npm test` chain pass locally.


gorkem2020:
Ready for review. Deployed on our gateway since 2026-09-16 03:53: the host's `excluding automatic memory context` line no longer appears on session start, and `USER.md` is back in the bootstrap prompt (verified in the recorded generation inputs: absent before, present after). Build and the full test chain pass.

PR #1001, opened 2026-09-17
fix(reflection): hand the distiller instructions to the embedded runner as extraSystemPrompt

gorkem2020:
## Summary

The reflection distiller builds a system/user prompt pair (`buildReflectionPrompt` returns `{ system, user }`), but the embedded runner call concatenated both halves into a single `prompt`. On current hosts the model therefore received the distiller identity and the section rules as user text, while the host's own minimal system prompt stood alone. The completion fallback already kept the split (`completeText(system, user)`), so the two runner paths disagreed on prompt shape.

This change passes the instructions as `extraSystemPrompt` and only the transcript as `prompt` on both runner names (`runEmbeddedAgent`, legacy `runEmbeddedPiAgent`). The host appends `extraSystemPrompt` to its minimal system prompt under a "Subagent Context" header; prompt mode `none` returns before that block, so the explicit `minimal` mode stays. The prompt hash still covers both halves, so dedup keys do not move.

## Tests

- `test/reflection-runner-embedded-agent.test.mjs`: two regression cases (new and legacy runner names) assert that `extraSystemPrompt` opens with the distiller identity, the user prompt carries the transcript, and neither the identity nor the section rules appear in the user prompt. Both fail on the previous code.
- Full suite, typecheck and build green; dist rebuilt in the same commit.

## Compatibility

No config, schema or manifest change. Hosts that ignore `extraSystemPrompt` keep the previous behaviour for the system half only; the transcript still arrives as the prompt.


gorkem2020:
Closing after the live check on our fleet: moving the distiller instructions under the host's system prompt does not give the distiller a system prompt of its own, it gives the host's agent persona the last word.

With `promptMode: "minimal"` the host appends `extraSystemPrompt` as a "Subagent Context" section after its own identity and bootstrap context, and mode `none` drops it entirely. In two /reset runs on a current host the model answered as the agent (a conversational follow-up in the agent's voice, no reflection headings), and one run ended with the provider's `content_filter` finish reason. The previous shape, where the instructions lead the user prompt, keeps producing well-formed reflections.

The completion fallback already receives a real system/user split through the plugin LLM lane, so a true split for the primary path would mean making that lane the primary transport, which is a design change rather than a fix. Leaving this closed; the branch stays on the fork for reference.

PR #1003, opened 2026-09-19
fix(auto-capture): hand the agent_end run back to the host so its async work scope stays open

gorkem2020:
## Summary

On OpenClaw 2026.9.5 every deferred memory extraction fails on the host transport with `Async work scope is closed`, so no new memories are captured after a turn. 2026.9.5 tracks a hook handler's returned promise in the turn's async work scope (`withHookTimeout` registers it through `trackAsyncWork`) and runs `runtime.llm.complete` under that scope. The auto-capture `agent_end` handler returned nothing and ran its extraction on a detached promise, so the host drained the scope right after the handler returned and the later completion was rejected at `AsyncWorkScope.track`. On 2026.9.2 neither mechanism existed, which is why the same code worked there.

This is the same class as [lossless-claw #1183](https://github.com/Martian-Engineering/lossless-claw/issues/1183), fixed there in [lossless-claw #1184](https://github.com/Martian-Engineering/lossless-claw/pull/1184) by letting the host own the background work.

## Changes

- `agent_end` auto-capture returns its tracked run instead of detaching it. The gateway dispatches `agent_end` fire-and-forget, so this adds no user-visible latency; hosts that ignore the return value are unaffected. The `session_end` flush already used this shape.
- The registration carries a 120 s hook timeout so a slow extraction does not trip the 30 s default void-hook timeout.
- A closed-scope failure on the host transport is tagged `[host runtime lifecycle]` in the last-error string (the `request failed for model` phrase the retry classifiers key on is kept) and listed as non-retryable, so no retry budget is spent on it and the log stops blaming the provider.

## Tests

- `test/autocapture-agent-end-owned-run.test.mjs`: the handler returns the tracked run (same object as `__lastRun`), it settles, the registration carries the timeout, and a skipped session still returns nothing. Red on the previous code.
- `test/host-runtime-lifecycle-error.test.mjs`: the tag, the classifier and the unchanged wording for ordinary upstream failures. Red on the previous code.
- Full suite, typecheck and build green locally; dist rebuilt in the commit.

## Validation

Draft until the fix is proven on a live gateway running 2026.9.5 (extraction completes and stores again, no lifecycle warning). Will undraft with that evidence.


gorkem2020:
Live proof on a gateway running OpenClaw 2026.9.5 with this branch folded into the fleet build (restart 2026-09-19 11:29 +03):

- one turn with a durable preference, then in the gateway log: `auto-capture payload shape ... fired=yes` at 11:30:16, `smart-extractor: extracted 1 candidate(s)` at 11:30:22, the admission and decision passes at 11:30:25 and 11:30:39, and `smart extraction settled` at 11:30:39, with no `Async work scope is closed` anywhere;
- the tracing backend shows the three `openai/gpt-oss-120b` extraction generations (extract, admission, decision) behind the agent's reply, which is exactly what failed on the same host before the change.

Undrafting on that evidence.

